PR #1, opened 2019-10-05
update README to define what ash is

andrewcallahan:
Not sure this all needs to go _here_, but figured it was good to get this down in writing anyway. I'm hoping you have some better words to use especially around what exactly a `resource` is. I want to try to go high level enough to articulate value proposition, while also accurately describing just what this does.


zachdaniel:
Beautiful.

PR #4, opened 2019-12-05
update ash resource doc

andrewcallahan:


lib/ash/resource.ex
```diff
@@ -44,22 +44,26 @@ defmodule Ash.Resource do
                         )
 
   @moduledoc """
-  The entry point for creating an `Ash.Resource`.
+  A resource is a public, standardized API on top of your data layer(s). A resource defines which CRUD operations are allowed to be performed, which data attributes are exposed, as well as relationships to other resources.
 
-  This brings in the top level DSL macros, defines module attributes for aggregating state as
-  DSL functions are called, and defines a set of functions internal to the resource that can be
-  used to inspect them.
+  Resource definitions should be placed in the `lib/resources` directory. The file name should be the single underscored name of the data that backs the resource with a `.ex` extension (ie: `lib/resoruces/post.ex`).
```
zachdaniel: I'm not 100% sure we want to be opinionated about where they live/what they are called. Perhaps we can just say "in your typical application using `Ash`, resources would be located ..."

lib/ash/resource.ex
```diff
@@ -44,22 +44,26 @@ defmodule Ash.Resource do
                         )
 
   @moduledoc """
-  The entry point for creating an `Ash.Resource`.
+  A resource is a public, standardized API on top of your data layer(s). A resource defines which CRUD operations are allowed to be performed, which data attributes are exposed, as well as relationships to other resources.
```
zachdaniel: You can technically have a resource w/ no data layer, and just use them to describe resources. A resource is a static definition of an entity in your system. Think "Posts" or "Authors" or "Tweets". They declare their CRUD operations, attributes, relationships, and other behavior, all of which can be customized to map to your underlying business logic. Resources can use "Data Layers", which allow them to be persisted and manipulated via `Ash`. Most importantly, front ends can take your resources and scaffold entire APIs/interaction engines based entirely off of that static declaration."

lib/ash/resource.ex
```diff
@@ -44,22 +44,26 @@ defmodule Ash.Resource do
...
 
-  This brings in the top level DSL macros, defines module attributes for aggregating state as
-  DSL functions are called, and defines a set of functions internal to the resource that can be
-  used to inspect them.
+  Resource definitions should be placed in the `lib/resources` directory. The file name should be the single underscored name of the data that backs the resource with a `.ex` extension (ie: `lib/resoruces/post.ex`).
 
-  Simply add `use Ash.Resource, ...` at the top of your resource module, and refer to the DSL
+  Resources are declaritive and can be configured using the DSL described below. Once declared, resources can be consumed anywhere in an application and are designed to be fully compliant with JSON:API and GraphQL web layers by using the AshJsonApi and AshGraphQL packages respectively.
```
zachdaniel: I would say something like: "They are designed to contain as much of your business logic as possible in a static declaration, so any front end can automatically build a fully compliant API using only your resources"

lib/ash/resource.ex
```diff
@@ -44,22 +44,48 @@ defmodule Ash.Resource do
...
+  In general an entity will refer to a single data concept and use a `Data Layer`, which allow them to be persisted and manipulated via Ash.
 
-  Simply add `use Ash.Resource, ...` at the top of your resource module, and refer to the DSL
+  | Data Layer | Use Case |
+  | :--- | :--- |
+  | Postgres | Database tables ("Posts" or "Authors" or "Tweets") |
+  | Postgres | Database views (SQL statistic such as Tweets Per Hour by Country) |
+  | ETS | _________???________ |
+  | File | Imagine a large json file of thousands of cities with metadata |
+  | External | 3rd Party APIs such as Stripe |
```
zachdaniel: I wouldn’t necessarily list these out like this quite yet. “External” data layers would need to be implemented custom for anyone who wanted to make one. I wouldn’t expect `file` so much as `json_file` or `csv_file`

Entities that aren’t backed by a data layer wouldn’t typically be used for something like registration. Realistically, you might just use resources to *describe* something. For instance, you could use resources to auto document an API that you already made or something. But for now I’d just leave all of that out of the readme. Don’t want to put the cart too far before the horse, the existing ETS And Postgres data layers are pretty far from done as is

lib/ash/resource.ex
```diff
@@ -44,22 +44,40 @@ defmodule Ash.Resource do
...
 
   @moduledoc """
-  The entry point for creating an `Ash.Resource`.
+  A resource is a static definition of an entity in your system.
 
-  This brings in the top level DSL macros, defines module attributes for aggregating state as
-  DSL functions are called, and defines a set of functions internal to the resource that can be
-  used to inspect them.
+  In general an entity will refer to a single data concept and use a `Data Layer`, which allow them to be persisted and manipulated via Ash.
+  Currently Ash supports Postgres and ETS (Erlang Term Storage) as Data Layers, but feel free to make an extention for other data layers such MySQL, JSON files, CSV files, or External APIs.
```
zachdaniel: how about: `Currently Ash provides Postgres and ETS (Erlang Term Storage) as Data Layers, but more are planned. Any one can create a Data Layer, either to provide support for a new database, a new storage format like CSV, or to power a custom use case like a resource that is backed by an external API.`

lib/ash/resource.ex
```diff
@@ -44,22 +44,40 @@ defmodule Ash.Resource do
...
 
-  Simply add `use Ash.Resource, ...` at the top of your resource module, and refer to the DSL
+  It is also possible to create a resource without a Data Layer, which for example could be used to provide autogemerated documentation of an API that you already made.
+
+  Regaurdless of whether or not a resource is backed by data, resources are designed to contain as much of your business logic as possible in a static declaration. Resources provide opportunities to declare CRUD operations, attributes, relationships, and other behavior, all of which can be customized to map to your underlying business logic.
```
zachdaniel: regardless

lib/ash/resource.ex
```diff
@@ -44,22 +44,40 @@ defmodule Ash.Resource do
...
+
+  Once a resource is declared, it will expose a public, standardized API that can be consumed in many different forms.
+
+  | Consumer | Use Case |
+  | :--- | :--- |
+  | Business Logic | Server side code such as from Phoenix Contexts |
+  | Web Layer | Full JSON:API and GraphQL web layer compliance via AshJsonApi and AshGraphQl |
```
zachdaniel: For now, leave out `GraphQL` from the docs, but put a note somewhere that a GraphQL front end is planned. I doubt it will be included in the initial release. Although I guess its very possible that we could...I wouldn't mind taking a bit longer if it means supporting both APIs at release.

lib/ash/resource.ex
```diff
@@ -44,22 +44,40 @@ defmodule Ash.Resource do
...
+  | Front Ends | UIs can use a schema file to know exactly how to interact web layer API |
+
+  In your typical application using Ash, resources would be located in the `lib/resources` directory. The file name should be the single underscored name of the data that backs the resource with a `.ex` extension (ie: `lib/resoruces/post.ex`).
```
zachdaniel: `lib/resources/post.ex`

lib/ash/resource.ex
```diff
@@ -44,22 +44,40 @@ defmodule Ash.Resource do
...
+
+  To create a resource simply add `use Ash.Resource, ...` at the top of your resource module, and refer to the DSL
   documentation for the rest. The options for `use Ash.Resource` are described below.
 
+  For example, here is a resource definition using Postgres:
+  ```elixir
+  defmodule AshKetchum.Pokemon do
```
zachdaniel: I would take out the pokemon references here and use something more generic like articles/posts. It would be cool to mirror those quick start "make a blog" guides but with ash. I wonder if someone will ever write like a "ui" mixin that will render resources to html or something. That could be kind of neat.

PR #5, opened 2019-12-05
update docs for ETS

andrewcallahan:
'

### DIFF
--- a/lib/ash/data_layer/ets.ex
+++ b/lib/ash/data_layer/ets.ex
@@ -1,7 +1,6 @@
 defmodule Ash.DataLayer.Ets do
   @moduledoc """
-  An ETS backed Ash Datalayer. Should only be used for testing, or for
-  unimportant/small datasets.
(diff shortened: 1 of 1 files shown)


lib/ash/data_layer/ets.ex
```diff
@@ -1,7 +1,6 @@
 defmodule Ash.DataLayer.Ets do
   @moduledoc """
-  An ETS backed Ash Datalayer. Should only be used for testing, or for
-  unimportant/small datasets.
+  An ETS (Erlang Term Storage) backed Ash Datalayer. This was initially built for testing purposes, since it comes built into OTP, as opposed to using Postgres, which would have required adding in another dependency into Ash. While this data layer can be used in your application, it should only be used for small/unimportant data sets that do not require long term persistence.
```
zachdaniel: Generally, module docs should always have one short line, and then an empty line, and then the long form documentation. Because documentation hover over features in the editor will show you that first little summary first. I would remove the mention of postgres.

`...since it comes built into OTP. This makes it possible to test resources easily, quickly and in isolation from the data layer.`

PR #11, opened 2020-05-09
refactor(filter): moves defimpl Inspect to corresponding struct files

vherr2:
Context
---
Colocates defimpl Inspect with the corresponding struct. This should help colocate the struct itself with the IO representation of the entity. No functional changes were made -- everything is just a move.


zachdaniel:

PR #12, opened 2020-05-09
refactor(filter, inspect): adds helpers and guards to clarify code

vherr2:
Context
---
* Creates private guard for checking against nil/empty values
* Moves variable assignment outside of function heads when not required for differentiating pattern matching
* Adds helpers for parsing attributes and relationships

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -857,14 +857,10 @@ defimpl Inspect, for: Ash.Filter do
   import Inspect.Algebra
   import Ash.Filter.InspectHelpers
 
-  def inspect(%Ash.Filter{impossible?: impossible, not: not_filter} = filter, opts)
-      when not is_nil(not_filter) do
-    impossible =
-      if impossible do
-        "X"
-      else
-        ""
-      end
+  defguardp is_empty(val) when is_nil(val) or length(val) == 0
+
+  def inspect(%Ash.Filter{not: not_filter} = filter, opts) when not is_nil(not_filter) do
+    impossible = if Map.fetch!(filter, :impossible?), do: "X", else: ""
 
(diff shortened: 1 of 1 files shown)


lib/ash/filter/filter.ex
```diff
@@ -857,14 +857,10 @@ defimpl Inspect, for: Ash.Filter do
...
 
-  def inspect(%Ash.Filter{impossible?: impossible, not: not_filter} = filter, opts)
-      when not is_nil(not_filter) do
-    impossible =
-      if impossible do
-        "X"
-      else
-        ""
-      end
+  defguardp is_empty(val) when is_nil(val) or length(val) == 0
```
zachdaniel, line 860: `val == [] or val == %{}` would be preferable, I think

zachdaniel: for performance.

vherr2: Does it actually matter? `length` should run in constant time right? Unless there's a minor perf difference in how it has to resolve the type first

zachdaniel: I don't see how `length()` would run in constant time. pretty sure its a traversal in Elixir.

zachdaniel: https://github.com/elixir-lang/elixir/issues/6374

zachdaniel: Seems possible they might special case `length(thing) == 0` to `thing == []` though.

vherr2: Ah yeah, I got `*_size` and `length` backwards. Yours is cleaner than what I had with `(is_map(...) and map_size(...) == 0) or length(...) == 0` though, so let's go with your proposal!

zachdaniel:

PR #13, opened 2020-05-10
improvement(filters): simplify/improve inspect logic

zachdaniel:


lib/ash/filter/filter.ex
```diff
@@ -923,33 +920,42 @@ defimpl Inspect, for: Ash.Filter do
...
+              break: :strict,
+              separator: " or "
+            )
+
+          if Enum.empty?(attrs) && Enum.empty?(rels) do
+            or_container
+          else
+            concat(["(", and_container, " or ", or_container, ")"])
+          end
       end
```
vherr2, line 939: Was this just wrong before?

zachdaniel: yep.

PR #94, opened 2020-08-20
Add is_nil support to Ets/Mnesia

TheFirstAvenger:
Initial pass at support for is_nil in Ets/Mnesia. Not positive this is the right approach, let me know if it is close.


lib/ash/filter/filter.ex
```diff
@@ -732,6 +732,13 @@ defmodule Ash.Filter do
     end
   end
 
+  defp add_expression_part({:is_nil, field}, context, expression) do
```
zachdaniel, line 735: Right now you'd do it with `filter(field: [is_nil: true])`. `filer(is_nil: :field)` looks really nice though, good addition.

zachdaniel:
Looks good! Just need to fix the credo error and then its good to go 👍

lib/ash/filter/filter.ex
TheFirstAvenger: Good to know, added support for the `is_nil: true` version.

TheFirstAvenger:
@zachdaniel good to go

zachdaniel:

Issue #50, opened 2020-06-07
Increase the elixir/erlang matrix we use for CI


zachdaniel:
We want to make sure we support more than just the latest elixir/erlang. Additionally, in our extension projects, we want to do the same thing, and also add the ash version to the matrix

TheFirstAvenger:
@zachdaniel Do you have thoughts on how you would test different versions of ash in the extension projects?

zachdaniel:
I haven't really worked it out yet, especially considering that things are currently changing fast enough that it's unlikely for an extension to support more than one version of ash. As it stabilizes though, I'll need a better plan. Additionally, there are parts of ash that can't really be tested without testing against extensions (specifically against ash_postgres). Right now those things are just untested. If you have thoughts/ideas let me know and we can get the ball rolling on better tests in that regard. Things are stabilizing just about enough to warrant it I think.

TheFirstAvenger:
Ecto does something similar to allow you to test ecto_sql against different versions of ecto, I will see if I can do something similar here.

TheFirstAvenger:
Also, as far as the Elixir/Erlang matrix:
Elixir:
v1.9 and below get the following:
```
== Compilation error in file test/actions/create_test.exs ==
** (BadFunctionError) expected a function, got: Ash.Dsl.Transformer
##[error]    (elixir) lib/enum.ex:2338: anonymous fn/3 in Enum.sort_by/3
##[error]    (stdlib) lists.erl:969: :lists.sort/2
##[error]    (elixir) lib/enum.ex:2338: Enum.sort_by/3
##[error]    (ash) lib/ash/dsl/transformer.ex:100: Ash.Dsl.Transformer.sort/1
##[error]    (stdlib) erl_eval.erl:680: :erl_eval.do_apply/6
##[error]    (stdlib) erl_eval.erl:449: :erl_eval.expr/5
##[error]    (stdlib) erl_eval.erl:126: :erl_eval.exprs/5
##[error]    (stdlib) erl_eval.erl:888: :erl_eval.expr_list/6
##[error]    (stdlib) erl_eval.erl:411: :erl_eval.expr/5
##[error]    (stdlib) erl_eval.erl:126: :erl_eval.exprs/5
##[error]    (ash) /home/runner/work/ash/ash/test/actions/create_test.exs:5: Ash.Resource.__before_compile__/1
##[error]    (elixir) lib/code.ex:767: Code.require_file/2
[...]

PR #96, opened 2020-08-21
ci: Improve CI Elixir/Erlang matrix

TheFirstAvenger:
Improve the Elixir/Erlang CI matrix. Copied matrix from Ecto, except bumped OTP 20 to OTP 21. Addresses #50.


TheFirstAvenger:
`ETS` requires at least OTP 21 due to the `:ets.whereis` call, and I dropped the supported version in mix down to `~> 1.8`.

TheFirstAvenger:
Looks like `persistent_term` is not available in 21.3.8.17, even though the OTP docs list it as 21.2/21.3.

zachdaniel:
Nice, I didn't consider using `Map.put` to just ensure the field is manually there. Good catch :)

zachdaniel:
Also, didn't realize that it wasn't the transformer sort that had the problem and that it was the runtime sorting.

zachdaniel:
I specced the mix task as `no_return` in master, so perhaps pulling that into here may resolve the issue.

TheFirstAvenger:
@zachdaniel nice, looks like that was the last issue, seems to be passing now on 1.8+

zachdaniel:
Awesome work as always!

PR #102, opened 2020-08-27
chore: Add validation of sub-projects

TheFirstAvenger:
This is a POC to see if we can validate sub-projects in CI when making changes to the ash project. It runs `mix compile --warnings-as-errors` and `mix test` on each of the subprojects using the current branch of `ash`. This will give advanced notice if a change in ash will require a fix in a subproject.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -98,3 +98,61 @@ jobs:
           consumer-secret: ${{ secrets.TWITTER_CONSUMER_API_SECRET }}
           access-token: ${{ secrets.TWITTER_ACCESS_TOKEN }}
           access-token-secret: ${{ secrets.TWITTER_ACCESS_TOKEN_SECRET }}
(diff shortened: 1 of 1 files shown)


TheFirstAvenger:
@zachdaniel Still learning Github Actions, but take a look at what I added and let me know your thoughts.

zachdaniel:
I think this is a really cool idea! We will most likely just need to accept that these builds will fail every once and a while, so I'll probably want to disable the "require build passing" so that people can merge PRs in the future. But I really like the idea of being able to confirm that master works with all the subprojects.

Very neat. It looks like its in a good spot to me, so let me know if you're ready and I'll merge it :)

TheFirstAvenger:
@zachdaniel yep, good to merge. And yes, the subprojects are going to fail when breaking changes are introduced, so you will need to either mark that status check not required, or use the admin override to merge anyway when it happens.

Issue #58, opened 2020-06-07
Support descriptions at every level of the DSL


zachdaniel:
We want to make sure that everything supports text descriptions, which can be used when scaffolding front end layers and building documentation.

- [x] actions
- [x] relationships
- [x] identities
- [x] validations
- [x] calculations
- [x] aggregates

mario-mazo:
Hello I will start adding `description` to: 

- https://github.com/ash-project/ash/tree/master/lib/ash/resource/actions
- https://github.com/ash-project/ash/tree/master/lib/ash/resource/relationships

since those already have test coverage

zachdaniel:
Remaining things to add descriptions to:

zachdaniel:
Great work!

PR #116, opened 2020-09-22
added description to actions and relationships

mario-mazo:
### Contributor checklist

- [x] My commit messages follow the [Conventional Commit Message Format](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#format-of-the-commit-message)
      For example: `fix: Multiply by appropriate coefficient`, or
      `feat(Calculator): Correctly preserve history`
      Any explanation or long form information in your commit message should be
      in a separate paragraph, separated by a blank line from the primary message

### DIFF
--- a/lib/ash/resource/actions/create.ex
+++ b/lib/ash/resource/actions/create.ex
@@ -1,12 +1,13 @@
 defmodule Ash.Resource.Actions.Create do
   @moduledoc "Represents a create action on a resource."
-  defstruct [:name, :primary?, :accept, :changes, type: :create]
+  defstruct [:name, :primary?, :accept, :changes, :description, type: :create]
 
   @type t :: %__MODULE__{
           type: :create,
           name: atom,
           accept: [atom],
-          primary?: boolean
+          primary?: boolean,
+          description: String.t()
         }
 
   @opt_schema [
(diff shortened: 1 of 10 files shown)


mario-mazo:
closes #58 
Not all resources have description but perhaps we can split in multiple PRs to avoid mega PRs

zachdaniel:
For consistency, I'd say to add the description to the "global_opts", that way it is clear from the documentation that each relationship shares it. We probably also want to add that same global options pattern to actions, if you're up for it.

mario-mazo:
I dont see how adding to `@global_opts` would make it more consistent  if what you mean is adding it `Ash.Resource.Relationships.SharedOptions` then it makes sense.

The only disadvantage I see is that we would lose the resource specific description like
"An optional description for the **many_to_many** relationship" and instead we would get generic ones

- "An optional description for the relationship" 
- "An optional description for the action" <- in case we add a SharedOptions to action resource

zachdaniel:
Yeah, shared options is what I was referring to, sorry :) But I think that is fine, if you're looking at the docs you'll see the description in the context of the action type you are looking at, and using shared opts helps make it really clear when something is unique to a given action type.

zachdaniel (approved):

PR #127, opened 2020-10-05
Module name errors

### LINKED ISSUE #120: Include the module in all DslErrors
Right now it can get confusing in which resource an error is occurring if the error is raised in a transformer. We should make DslError take a resource, and display it always.

WolfDan:
# Contributor checklist

Fixes #120


zachdaniel:

PR #130, opened 2020-10-11
feat: Add `:one_of` constraint to the Atom type

doawoo:
This is something I came across that I wanted during some experimentation with Ash, and I decided to take a crack at drafting a quick PR for it! 

Basically when defining an `:atom` attribute I wanted to ensure that the input value is restricted to a pre-defined list of atoms.

Quick example in context:

```elixir
    attribute :type, :atom, allow_nil?: false, constraints: [
      one_of: [:text_post, :video_post, :image_post]
    ]
```

Feedback encouraged! I love what I see of the Ash framework so far 😄 

# Contributor checklist
- [ ] ~~Bug fixes include regression tests~~

### DIFF
--- a/lib/ash/type/atom.ex
+++ b/lib/ash/type/atom.ex
@@ -1,4 +1,10 @@
 defmodule Ash.Type.Atom do
+  @constraints [
+    one_of: [
+      type: :any,
+      doc: "Allows constraining the value of an atom to a pre-defined list"
+    ]
+  ]
   @moduledoc """
   Stores an atom as a string in the database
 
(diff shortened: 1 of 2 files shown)


zachdaniel:
Hey @doawoo! Thanks for the contribution :) Really glad that you like it so far.

I'm not 100% sure that this is the way to go, specifically because there is already a concept of `validations` - https://hexdocs.pm/ash/1.16.2/Ash.Resource.Validation.Builtins.html#one_of/2

So you can say:
```
validations do
  valdate one_of(:attribute, [:foo, :bar, :baz])
end
```

However, atom fields are a special case that would likely be the underlying implementation for `enum` types in ash_graphql/other interfaces. With that in mind, it may make sense to support this in *both* places. Then, in ash_graphql, when it encounters an `atom` field with a `one_of` constraint, it could make an enum.

With that said, we could also derive that, by checking for a `one_of` validation on that attribute. I'll think on it for a bit. What are your thoughts?

P.S - Thank you again for contributing!

doawoo:
Validations look quite nice!
The idea for this came from wanting an enum-like type-level validation. So the fact that you bring up the GraphQL usage makes me feel like it fits decently. 
Ultimately though I'd be interested in what others think.

zachdaniel:
I think we can go for it. It does fit well with the enum pattern in GraphQL, and I like that it stays generic at the resource level, e.g not `enum :name, ...`, since that doesn't really need to be a resource level concept.

PR #143, opened 2020-11-08
fix: do not require private attributes in create api

kyle5794:
# Contributor checklist


zachdaniel:
Good catch!

PR #150, opened 2020-12-02
feat: complexity metrics

mangeption:
# Contributor checklist


zachdaniel:
Going to close this since we haven't done anything on it for a while. Complexity metrics/some way to limit the complexity of incoming filters/requests is still very important

zachdaniel:
So the primary reason I'm closing this isn't because its stale, its because actions/relationships and everything are still evolving (e.g arguments weren't a thing before this) and the underlying implementation may ultimately just end up as a query preparation/change validation, something along those lines. We will revisit this though.

PR #162, opened 2021-01-08
improvement: Add built in decimal type

zimt28:


zachdaniel:
Interesting that you don't get the dialyzer errors on your machine, but here is what I see:
CleanShot 2021-01-09 at 11 40 48@2x

zachdaniel:
Looks like its an issue with the version of Decimal we're using. Ash is locked on an older version with different behavior for `compare/2`

zachdaniel:
I updated that dependency in dev, so if we merge this now we should be good.

zachdaniel:

PR #164, opened 2021-01-09
Improve attribute defaults

zimt28:
In order to mimic Ecto's behaviour, timestamps aren't allowed to be `nil` by default. Primary keys aren't allowed to be `nil` anyway.

### DIFF
--- a/lib/ash/resource/transformers/belongs_to_attribute.ex
+++ b/lib/ash/resource/transformers/belongs_to_attribute.ex
@@ -23,7 +23,12 @@ defmodule Ash.Resource.Transformers.BelongsToAttribute do
       case Transformer.build_entity(@extension, [:attributes], :attribute,
              name: relationship.source_field,
              type: relationship.field_type,
-             allow_nil?: not relationship.required?,
+             allow_nil?:
+               if relationship.primary_key? do
+                 false
+               else
+                 not relationship.required?
+               end,
              writable?: false,
              private?: true,
              primary_key?: relationship.primary_key?
--- a/lib/ash/resource/transformers/belongs_to_attribute.ex
+++ b/lib/ash/resource/transformers/belongs_to_attribute.ex
@@ -20,34 +20,47 @@ defmodule Ash.Resource.Transformers.BelongsToAttribute do
       |> Enum.find(&(Map.get(&1, :name) == relationship.source_field))
     end)
     |> Enum.reduce_while({:ok, dsl_state}, fn relationship, {:ok, dsl_state} ->
-      case Transformer.build_entity(@extension, [:attributes], :attribute,
-             name: relationship.source_field,
-             type: relationship.field_type,
-             allow_nil?:
-               if relationship.primary_key? do
-                 false
-               else
-                 not relationship.required?
-               end,
-             writable?: false,
-             private?: true,
-             primary_key?: relationship.primary_key?
-           ) do
-        {:ok, attribute} ->
-          {:cont, {:ok, Transformer.add_entity(dsl_state, [:attributes], attribute)}}
+      entity =
+        Transformer.build_entity(@extension, [:attributes], :attribute,
+          name: relationship.source_field,
+          type: relationship.field_type,
+          allow_nil?:
+            if relationship.primary_key? do
+              false
+            else
+              not relationship.required?
+            end,
+          writable?: false,
+          private?: true,
+          primary_key?: relationship.primary_key?
+        )
 
-        {:error, error} ->
-          {:halt,
-           {:error,
-            DslError.exception(
-              module: __MODULE__,
-              message:
-                "Could not create attribute for belongs_to #{relationship.name}: #{inspect(error)}",
-              path: [:relationships, relationship.name]
-            )}}
-      end
+      valid_opts? = not (relationship.primary_key? && !relationship.required?)
+
+      entity_or_error =
+        if valid_opts? do
+          entity
+        else
+          {:error, "Relationship cannot be a primary key unless it is also marked as required"}
+        end
+
+      add_entity(entity_or_error, dsl_state, relationship)
     end)
   end
+
+  defp add_entity({:ok, attribute}, dsl_state, _relationship),
+    do: {:cont, {:ok, Transformer.add_entity(dsl_state, [:attributes], attribute)}}
+
+  defp add_entity({:error, error}, _dsl_state, relationship),
+    do:
+      {:halt,
+       {:error,
+        DslError.exception(
+          module: __MODULE__,
+          message:
+            "Could not create attribute for belongs_to #{relationship.name}: #{inspect(error)}",
+          path: [:relationships, relationship.name]
+        )}}
 
   def after?(Ash.Resource.Transformers.BelongsToSourceField), do: true
   def after?(_), do: false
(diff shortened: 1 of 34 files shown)


zachdaniel:
If primary keys aren't allow to be nil why would we make them `allow_nil? true`?

zimt28:
Where do you see `allow_nil?: false`? My idea was that Postgres doesn't allow `NULL` values for primary keys, but also doesn't add a not null constraint, as it's "included". So we can just omit the `allow_nil?: false` default - a not null constraint doesn't have to be added and it's not writable anyway.

zachdaniel:
Sorry, I meant `allow_nil?: true`. That is the default, and by removing `allow_nil?: false` we are implicitly saying this value could be `nil`.

I think the timestamp change is good, but it is Postgres specific behavior here (that a primary key can't be nil) and Postgres isn't the only Ash data layer. Additionally, `allow_nil` isn't only used for validation, its used as a statement about the attribute, e.g that it will always return a value, so it gets `ID!` in graphql instead of `ID`.

We can update the migration generator to not add a `null: false` flag on attributes that are primary key attributes, or we could make the data layer raise a compile time error on a resource with a nullable primary key.

zachdaniel:
It may also be reasonable to prevent any primary key from being nullable in general, just for consistencies sake. So we'd still want the attribute to say `allow_nil? false` but we wouldn't let it be changed to `allow_nil? true` ever.

zimt28:
I see, I just thought about Postgres. So should I add `allow_nil?: false` back to integer and UUID primary keys? Or does `integer_primary_key` need special treatment as it's generated in the database?

>  migration generator to not add a null: false flag on attributes that are primary key attributes
Yes, that would be nice. It doesn't do anything bad as it's just being ignored, but it would be cleaner without. I think `null`, `default` and `primary_key` can always be omitted, if they are false/ nil.

zachdaniel:
It sounds like it may be a good idea to just prohibit nullable primary keys in general (thus making this no longer a data layer concern, and being a rule in Ash core). To do that, we'd need to 

* update the `transform` function for the `attribute` entity, and have it return an error if `primary_key?: true` and `allow_nil?: true`. 
* For `uuid_primary_key` and `integer_primary_key`, we'd update the DSL entry itself. 
  * First we'd *remove* `allow_nil?` from the schema, e.g `|> Keyword.delete(:allow_nil?)`
  * Then we'd add `allow_nil?: false` to the `auto_set_fields` list.

zachdaniel:
Instead of `|> Keyword.put(:allow_nil?, false)` do `|> Keyword.delete(:allow_nil?)` (because we're saying no `nil` primary keys from now on), and then add the following to the DSL entities found [here](https://github.com/ash-project/ash/blob/master/lib/ash/resource/dsl.ex#L65) and [here](https://github.com/ash-project/ash/blob/master/lib/ash/resource/dsl.ex#L51)

```elixir
%Ash.Dsl.Entity{
  ...,
  auto_set_fields: [allow_nil?: false]
}
```

zimt28:
Doesn't work yet, but is it going in the right direction?

zimt28:
Assuming it's _basically_ right, I don't like this implementation: https://github.com/ash-project/ash/blob/10030017e9f34bc928eb2f77816a495104ca2869/lib/ash/resource/transformers/cache_primary_key.ex#L9-L29

I played around a bit, but this is still the best I could find. Alternatives were:

- `if/else` statement around the case to check for primary keys with `allow_nil?: true`
- `cond` which would check for `allow_nil?: true` first but gets ugly then (using `length` instead of nice pattern matching`
- using `case {pk_allows_nil?, primary_key} do` which leads to ugly pattern matching too

zachdaniel:
LGTM. I'd just suggest using `Enum.any?` instead of `count > 0`

zachdaniel:
Looks like the `destroy_test.exs` just needs to use `uuid_primary_key` instead of `attribute :id, :uuid, ...` since `allow_nil?` defaults to true.

zimt28:
The only thing left is `belongs_to`, I cannot figure out where it sets `allow_nil?` to `true`.

lib/ash/resource/transformers/belongs_to_attribute.ex
```diff
@@ -23,7 +23,12 @@ defmodule Ash.Resource.Transformers.BelongsToAttribute do
       case Transformer.build_entity(@extension, [:attributes], :attribute,
              name: relationship.source_field,
              type: relationship.field_type,
-             allow_nil?: not relationship.required?,
+             allow_nil?:
+               if relationship.primary_key? do
```
zachdaniel, line 27: Here we should return an error, e.g `{:error, "Belongs to relationship cannot be a primary key unless it is also marked as required}`. By setting it to `false` here, we'd override their resource, and they could have `belongs_to :foo, Bar, required?: false` not realizing that it is being ignored.

zimt28: Ok

zimt28:
This is the only failing test now:

```
  1) test upsert (Ash.DataLayer.EtsTest)
     test/ash/data_layer/ets_test.exs:101
     ** (Ash.Error.Invalid) Input Invalid

     * Invalid value provided for id: Attribute is not writable.
     code: create_user(%{name: "Joe", id: id}, upsert?: true)
     stacktrace:
       (ash 1.27.0) lib/ash/api/api.ex:902: Ash.Api.unwrap_or_raise!/2
       test/ash/data_layer/ets_test.exs:106: (test)
```

zachdaniel:
Huh. Going to have to think about that one. I see what the problem, but I'm not really sure what the right answer...

I guess it would have to be a rule that the primary key is writable for an upsert to be acceptable. So you can just update the primary key in that test to have `writeable?: true`.

zimt28:
I've added the suggested error message, and fixed many tests which were then throwing that error. Now some relationship tests are failing because of the added `required?: true` config

zachdaniel:
I think we'll just need to update the failing resources with `required?: true`

zimt28:
Hm, let's look at this one for example:
```
  1) test creating many to many relationships allows creating with a many_to_many relationship (Ash.Test.Actions.CreateTest)
     test/actions/create_test.exs:322
     ** (Ash.Error.Invalid) Input Invalid

     * relationship destination_post is required
     * relationship source_post is required
     code: |> Api.create!()
     stacktrace:
       (ash 1.27.0) lib/ash/api/api.ex:902: Ash.Api.unwrap_or_raise!/2
       test/actions/create_test.exs:338: (test)
```
`replace_relationship` causes the error:
```elixir
post2 = # ...
post3 = # ...

Post
|> new(%{title: "cannot_be_missing"})
|> replace_relationship(:related_posts, [post2, post3])
|> Api.create!()
```
`Post`:
```elixir
relationships do
  # ...
  many_to_many(:related_posts, __MODULE__,
    through: PostLink,
    source_field_on_join_table: :source_post_id,
    destination_field_on_join_table: :destination_post_id
  )
end
```
`PostLink`:
```elixir
relationships do
  belongs_to(:source_post, Ash.Test.Actions.CreateTest.Post,
    primary_key?: true,
    required?: true
  )

  belongs_to(:destination_post, Ash.Test.Actions.CreateTest.Post,
    primary_key?: true,
    required?: true
  )
end
```
I've already added `required?: true` and I think it causes the issue, as the code works fine without:

`PostLink`:
```elixir
attributes do
  # Doesn't work without primary key
  uuid_primary_key :id
end

relationships do
  belongs_to :source_post, Ash.Test.Actions.CreateTest.Post
  belongs_to :destination_post, Ash.Test.Actions.CreateTest.Post
end
```

zachdaniel:
Interesting. I think this may be a separate bug related to how we validate required belongs to relationships. I'll take a look, hopefully tonight but maybe tomorrow.

zachdaniel:
Alright, I've fixed the issue in `master`, so I'm merging this.

Issue #167, opened 2021-01-11
Treat empty values as nil


zimt28:
Currently, this is possible:

```elixir
attributes do
  attribute :title, :string, allow_nil?: false
end

{:ok, _} =
  resource
  |> Changeset.for_create(:default, %{title: ""})
  |> API.create()
```

Values with `String.trim(value) == ""` should be treated as `nil`. 

Edit: Maybe empty characters (https://emptycharacter.com/) should be replaced by space before trimming.

zachdaniel:
We should probably implement it as a `constraint` that defaults to this behavior. We could take the name `empty_values` from ecto. https://hexdocs.pm/ecto/Ecto.Changeset.html#module-empty-values

We could also add options to trim empty characters as another constraint. We could then modify the behavior like so:

```
attribute :foo, :string,
  empty_values: [],
  replace_empty_char?: false
```

zimt28:
I'll send a PR

zimt28:
Gave it a try (#169)

zachdaniel:
Done!

PR #169, opened 2021-01-11
Add constraints for handling empty values (#167)

zimt28:
Doesn't work yet, as `{:ok, value}` isn't a valid return for constraints yet.


zimt28:
I just noticed that `String.trim/1` is actually smart enough to remove **any**  empty characters, not just regular spaces 🙄 So I think I will remove the `empty_values` and `empty_value_replacement` options again, or are they still helpful in any way? It should probably be up to the user how to deal with those details.

zachdaniel:
We can start with just the `trim?` option and add back in customized empty characters/empty values if someone needs it.

zachdaniel:
@zimt28 if you pull master, the `{:ok, new_value}` return type is now supported for `handle_constraints`

zimt28:
@zachdaniel Great, anything left to do here? I could add a test tomorrow

zachdaniel:
Yeah, go ahead and pull master back into your branch and then add a test then we can merge it.

zimt28:
Whoops, that didn't work. Will send new PR!

PR #183, opened 2021-02-03
improvement: Implement string length validation

totaltrash:
Here's a first crack at a string length validator. It can be used with `:min`, `:max`, `:exact` and a combination of `:min` and `:max`. The contents of the default error message is dependent on the other params provided, and the error message can be overridden.

```elixir
    validate string_length(:body, min: 3, message: "must be min %{min} characters")
```

# Contributor checklist

### DIFF
--- a/lib/ash/resource/validation/builtins.ex
+++ b/lib/ash/resource/validation/builtins.ex
@@ -34,6 +34,11 @@ defmodule Ash.Resource.Validation.Builtins do
     {Validation.AttributeEquals, attribute: attribute, value: value}
   end
 
+  @doc "Validates that an attribute on the original record meets the given length criteria"
+  def string_length(attribute, opts \\ []) do
+    {Validation.StringLength, Keyword.merge(opts, attribute: attribute)}
+  end
+
   @doc """
   Validates that an attribute's value matches a given regex or string, using the provided error, message if not.
 
--- a/lib/ash/resource/validation/string_length.ex
+++ b/lib/ash/resource/validation/string_length.ex
@@ -1,0 +1,104 @@
+defmodule Ash.Resource.Validation.StringLength do
+  @moduledoc false
+  use Ash.Resource.Validation
+
+  alias Ash.Error.Changes.InvalidAttribute
+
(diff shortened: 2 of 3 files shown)


zachdaniel:
This looks great! One thing to note: the message overriding feature is already a part of the validation logic (although broken at the moment). I'll wait to merge this until I do my next release which fixes the message overriding, and then you should be able to remove the message logic inside your own validation and have this work:

```elixir
validate string_length(:body, min: 3), message: "must be min %{min} characters"
```

totaltrash:
Ok cool. Will that work with the string_length validator customising the default error message depending on the specific params passed? for instance when `:min` is set the default error is '%{field} must have length of at least %{min}', but when `:max` is set, a different default error message is used, etc.

zachdaniel:
It should be fine, since they can use a different message in that case
```elixir
 validate string_length(:body, min: 3), message: "gotta have %{min} dawg"
 validate string_length(:body, max: 3), message: "can't have %{max} dawg"
```

totaltrash:
Yep, i'm just meaning the defaults so:

```elixir
validate string_length(:body, min: 3)
```
will generate a default message of `'%{field} must have length of at least %{min}'`

whereas...
```elixir
validate string_length(:body, max: 3)
```

will default to `'%{field} must have length of no more than %{min}'`

zachdaniel:
Ah, yeah you still return that message in the validation in the same way you do now, you just don't need `opts[:message] || "message"` because ash will replace the message in your return result with theirs.

totaltrash:
Sweet!

I've got a question around how `nil`s should be handled, when the attribute is `allow_nil?: true`. At the moment it just skips over them... I feel maybe they should be handled by another validation or constraint (`allow_empty?`)?

zachdaniel:
I don't think nil values will be passed into the constraint validation currently.

totaltrash:
Hey @zachdaniel, this should be good to go

zachdaniel (approved):

zachdaniel:

PR #185, opened 2021-02-05
Test Api.create for integer primary key resource

regularfellow:


zachdaniel:
So this test is likely producing a primary key with a `nil` value, since the ETS data layer doesn't have any logic currently to generate an integer primary key. What we could do is get the latest value from a table or create that table if it doesn't exist, then we could support auto-incrementing ids in ETS.

Another action item we may want to add:

`integer_primary_key` sets a `generated?: true` flag on the attribute. We should probably ask the data layer module if it can support each generated attribute, that way the ETS data layer can fail loudly saying that it doesn't support generated integer keys.

zachdaniel:
Going to close this in favor of the test case in ash_postgres

PR #194, opened 2021-03-13
Fix documentation TOC

zimt28:
**I found just one example where this fix is needed, maybe it needs to be the other way around**

Example: https://hexdocs.pm/ash/1.36.3/Ash.Resource.Dsl.html#content

The generated `code_interface` TOC link is `#module-code-interface` while the correct link would be `#module-code_interface`.


zachdaniel:

zachdaniel:
Yep, looks like it wasn't an issue for the others because none of the other section names had multiple words. Good catch 👍

PR #198, opened 2021-03-14
improvement: Add timestamps() attribute

zimt28:
```elixir
attributes do
  timestamps()

  # will create the same attributes as

  create_timestamp :inserted_at
  update_timestamp :updated_at
end
```

I couldn't find a more elegant way to implement this, if there is one I'd be happy to change it :)


zachdaniel:

PR #199, opened 2021-03-14
Add param option

zimt28:
I thought I'd give the `params` option a try. It's still WIP and very ugly, but I wanted to check if this is the right direction. Metaprogramming is quite difficult 😅.


zachdaniel:
Definitely the right direction 👍 Manipulating the list of args and using `unquote_splicing` is definitely the way to go, as the alternative involves defining the same function slightly differently a bunch of times.

zachdaniel:
Going to close this as there hasn't been much movement, but happy to reopen it and get it merged if its completed.

PR #211, opened 2021-03-25
WIP Ash gen api

dkuku:
# Contributor checklist

- [ X] Features include unit/acceptance tests


zachdaniel (dismissed):

zachdaniel:
Hey there @dkuku so I'm planning on just merging this and then making some amendments to it in master before it is released. It just needs to be formatted and has a few compiler warnings. If you wouldn't mind taking care of those, that would be great :) I'll take care of updating the templates/a few small tweaks after we get it in.

PR #219, opened 2021-04-03
Docs: Fix Api.get example

savish:
The `Api.get` example in the docs uses the email instead of the ID to
get the newly created resource. Trying this locally while running
through the getting started tutorial returns a `nil` resource using the
email, but the correct resource when using the ID.

This fix updates the docs to use the ID in the get call.

# Contributor checklist


zachdaniel:

zachdaniel (approved):

PR #220, opened 2021-04-04
allow specifying that calculation can't be nil

michaelst:
I would like a way to specify that a calculation value can't return nil to use in generating the graphql type. Let me know if there is any validation I need to add and where I should add it.

# Contributor checklist


zachdaniel (approved):

zachdaniel:

zachdaniel:
Released in `0.14.0`

PR #226, opened 2021-04-17
Embedded struct casting fix

MrFlorius:
# Contributor checklist

I've discovered a strange behaviour of embedded resources. If they are passed to the `Ash.Changeset.new` as a structs, `Ash` assumes they are already casted, which is not necessarily true. 
Now `cast_input/2` and `cast_stored/2` in `Ash.EmbeddableType.single_embed_implementation` track if the resource was casted adding `%{private: %{casted?: true}}` to the resource metadata


zachdaniel:
Going to close this for now, until we decide whether or not this is how it will be done.

PR #241, opened 2021-05-27
fix: set tenant even on non-tenant resources where relevant

kernel-io:
As discussed on Discord.
This rectifies the edge-case of updating a Resource that doesn't have multitenancy enabled, while modifying one of it's related resources which does. i.e: Tenant belongs_to Address, while Address has a tenant_id key on it for multi-tenancy.


zachdaniel (approved):

zachdaniel:
Merged, will release with the next ash release in the next few days

PR #242, opened 2021-05-30
feat: add compare validator

kernel-io:
feat: added number validator supporting validation of arguments and attributes, all the standard validators are included i.e: greater_than, less_than, less_than_or_equal_to, greater_than_or_equal_to
feat: added two new functions in changeset, `&get_argument_or_attribute/2` and `&fetch_argument_or_change/2`
fix: give `present` validation access to arguments also, not just arguments

added some tests

### DIFF
--- a/lib/ash/api/interface.ex
+++ b/lib/ash/api/interface.ex
@@ -173,7 +173,7 @@ defmodule Ash.Api.Interface do
                     Keyword.take(opts, [:actor, :tenant])
                   )
 
-                unquote(api).create(changeset, opts)
+                unquote(api).create(changeset, Keyword.drop(opts, [:query, :tenant]))
               end
             end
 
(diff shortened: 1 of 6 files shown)


lib/ash/resource/validation/number.ex
```diff
@@ -0,0 +1,118 @@
...
+      _ ->
+        :ok
+    end
+  end
+
+  defp invalid_attribute_error(opts, message \\ nil) do
+    {:error,
+     InvalidAttribute.exception(
+       field: opts[:attribute],
+       message: message || opts[:message],
```
kernel-io, line 109: causing the dialyzer to rightly complain

zachdaniel: Yeah, you want `opts[:message] || message` here right?

zachdaniel:
Can you format the code with `mix format`, take a look at the compiler warning and then if you fix the `message || opts[:message]` thing it should be good to go 👍🏻 great work!

zachdaniel:
🚀 Thank you for your contribution! 🚀w

zachdaniel (approved):

PR #243, opened 2021-05-31
Fix create code interface to correctly accept tenant in the opts list

kernel-io:
fix: ensure create functions generated by code_interface accept tenant in the opts list


zachdaniel:
Thanks for your contribution!

PR #294, opened 2022-03-19
improvement: Improve usability of finding by primary key (ID)

### LINKED ISSUE #293: Implementing `get_by_pkey` action could be more obvious
First thing I tried to do for read actions was implement fetch all and fetch by id, and ended up with this:

```elixir
  actions
    read :all do
      primary? true
    end

    read :get_by_id
  end

  code_interface do
    define_for Api

    define :all, action: :all
    define :get_by_id, action: :get_by_id, args: [:id], get?: true
  end
```

I assumed that the `define :get_by_id, action: :get_by_id, args: [:id], get?: true` line did the filtering. Sometimes I got `multiple results` errors, and even more confusing, if there was only one record I got a valid result. That was super confusing.

It wasn't until I stumbled across the filter example here https://hexdocs.pm/ash/expressions.html#filters that I realised what was missing. This allowed me to fix the problem:

```elixir
  actions do
    read :all do
      primary? true
    end

    read :get_by_id do
      get? true
      argument :id, :uuid, allow_nil?: false

      filter expr(id == ^arg(:id))
    end
  end

  code_interface do
    define_for Api

    define :all, action: :all
    define :get_by_id, action: :get_by_id, args: [:id], get?: true
  end
```

For a standard CRUD resource I'd be repeating this pattern everywhere, but it is explicit.

### Possible Solutions

- A clearer example code in the getting started guide would have helped a bunch here. eg Here's how you do a standard fetch all and fetch by id

- When `get_by_id`  with `get?: true` returned multiple results. This could have produced a more helpful error message eg
  > You've specified "read_one" which expects a single result, but the query has returned more than one result.
  > Please use a filter to ensure a single result is fetched

- An option like `define :get_by_id, ... get_by_pkey?: true` could provide the filter for you against your primary key, removing the need for `args: [:id], get?: true`

joshprice:
Fixes #293

Indicate that MultipleResult error is probably failing due to a missing filter.

The docs for filters live here: https://hexdocs.pm/ash/expressions.html#filters

### DIFF
--- a/lib/ash/error/invalid/multiple_results.ex
+++ b/lib/ash/error/invalid/multiple_results.ex
@@ -10,7 +10,9 @@ defmodule Ash.Error.Invalid.MultipleResults do
     def code(_), do: "multiple_results"
 
     def message(%{count: count, query: nil, at_least?: at_least?}) do
-      "expected at most one result but got #{at_least(at_least?)}#{count}"
+      "expected at most one result but got #{at_least(at_least?)}#{count}.
+      
+      Please ensure your action is configured with an appropriate filter to ensure a single result is returned."
     end
 
     def message(%{count: count, query: query, at_least?: at_least?}) do


zachdaniel (approved):

joshprice:
Think this is good to go as is. I'm no longer sure about implementing a `get_by_pkey` option, it doesn't make sense without the Api interface setup.

Issue #214, opened 2021-03-27
Validate unique names


zimt28:
It's currently possible to give an attribute and relationship the same name, this shouldn't be allowed.

zimt28:
Closed via #300, thanks @tlietz 🎉

PR #300, opened 2022-03-27
improvement: validates attributes and relationships have unique names

tlietz:
Issue: https://github.com/ash-project/ash/issues/214
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
This looks excellent! The only thing I'd say is that we should include calculations and aggregates as well, as they should all be unique.

tlietz:
@zachdaniel Sounds good, should those all be in the same transformer file?

zachdaniel:
Yeah, what you'd want to do is group up all of them with the same name, and then say something simple like: "There are multiple fields(attributes, calculations, aggregates, and relationships) that share the name <foo>".

tlietz:
Since this test case is including all fields that are a part of a resource, should it be moved out of `test/resource/relationships/belongs_to_test.exs` and into something like `test/resource/validations_test.exs` instead?

zachdaniel:
Yeah, good idea! 👍🏻

zachdaniel:
Actually, `validations_test` is for actual validations. I'd make a new file `resource_test.exs`.

tlietz:
Alright, this should be ready for review now.

lib/ash/resource/dsl.ex
```diff
@@ -1110,7 +1110,8 @@ defmodule Ash.Resource.Dsl do
     Ash.Resource.Transformers.DefaultPrimaryKey,
     Ash.Resource.Transformers.DefaultAccept,
     Ash.Resource.Transformers.SetTypes,
-    Ash.Resource.Transformers.ValidateRelationshipAttributes
+    Ash.Resource.Transformers.ValidateRelationshipAttributes,
```
zachdaniel, line 1113: we can remove `ValidateRelationshipAttributes` now right?

zachdaniel:
Had one comment, otherwise this is good to go. Great work!

lib/ash/resource/dsl.ex
tlietz: Yes we can, good catch!

zachdaniel:

Issue #266, opened 2021-09-24
Provide helpful info when using Api.read without a data layer setup


michaelst:
Right now calling Api.read on a resource that does not have a data layer setup will return `{:ok, []}` 

We should improve this to make it more obvious why no data is being returned.

Chat from discord:
<img width="993" alt="image" src="https://user-images.githubusercontent.com/4080508/134741733-990d8e6a-8bb2-4835-9613-905cee85929d.png">

PR #302, opened 2022-03-29
improve: Provide api read error message on no data set

tlietz:
Issue: https://github.com/ash-project/ash/issues/266

I don't know if `class: invalid` fits with the new `no data provided` error. I just thought of it as happening when there is an invalid read from a resource.

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Looks excellent! I think we should use `:framework` as the error type. Using type `:invalid` carries the implication that it may be something you'd like to communicate with the outside world, e.g "your request was invalid because of <X>"

zachdaniel:

PR #310, opened 2022-04-15
improvement: Add time type matching existing date type

joshprice:
# Contributor checklist

- [x] Documentation changes

### DIFF
--- a/lib/ash/type/time.ex
+++ b/lib/ash/type/time.ex
@@ -1,0 +1,31 @@
+defmodule Ash.Type.Time do
+  @moduledoc """
+  Represents a time in the database
+
+  A builtin type that can be referenced via `:time`
+  """
+  use Ash.Type
+
+  @impl true
+  def storage_type, do: :time
+
+  @impl true
+  def cast_input(value, _) do
+    Ecto.Type.cast(:time, value)
+  end
+
+  @impl true
+  def cast_stored(nil, _), do: {:ok, nil}
+
+  def cast_stored(value, _) do
+    Ecto.Type.load(:time, value)
+  end
+
+  @impl true
+
+  def dump_to_native(nil, _), do: {:ok, nil}
+
+  def dump_to_native(value, _) do
+    Ecto.Type.dump(:time, value)
+  end
+end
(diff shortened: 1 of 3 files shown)


joshprice:
This seems to work once I added the time type to the builtin list. However I've run into an issue where I can't seem to set reasonable defaults, on `:time` or `:date` types: 

```elixir
    attribute :day, :date do
      allow_nil? false
      default ~D[2022-04-04]
    end

    attribute :day_end_time, :time do
      allow_nil? false
      default ~T[15:00:00]
    end
```

zachdaniel:
🤔 yeah that is problematic. We should address that default behavior to only support known value types for sure. This is insufficient logic here:

```elixir
  defp default(%{default: value, type: type}, _) do
    case Ash.Type.dump_to_native(type, value) do
      {:ok, value} when is_struct(value) ->
        to_string(value)

      {:ok, value} ->
        inspect(value)

      _ ->
        "nil"
    end
  rescue
    _ ->
      "nil"
  end
```

zachdaniel:
Maybe even a more generic `ToSql`.

lib/ash/type/time.ex
```diff
@@ -0,0 +1,31 @@
...
+
+  A builtin type that can be referenced via `:time`
+  """
+  use Ash.Type
+
+  @impl true
+  def storage_type, do: :time
+
+  @impl true
+  def cast_input(value, _) do
```
zachdaniel, line 13: We should add a `def cast_input(nil, _), do: {:ok, nil}`

joshprice: Fixed

joshprice:
Not sure I quite follow yet what's required for the defaults to work correctly. Can we make that a separate issue?

joshprice:
I wanted to write a test for these changes but noticed that many types weren't tested, I've added a rudimentary time happy case test, but let me know if there's any more cases worth including.

joshprice:
I think all doc changes needed are handled by this line https://github.com/ash-project/ash/blob/main/lib/ash/type/type.ex#L51

zachdaniel:
That looks great :)  A lot of the primitive types just pass through to ecto so they don't all have tests.

zachdaniel:

PR #321, opened 2022-04-27
docs: render contributor images correctly in hexdocs

joshprice:
New contributor list looks great but isn't rendering in hexdocs. Used nested image in anchor tag to render standard markdown for ex_doc.

Problem was that default avatars cannot be resized so have used a random gravatar identicon which can be sized via a URL param.


zachdaniel:
I should probably regenerate this, and/or move it to something we keep up to date in ash-hq.

PR #340, opened 2022-05-30
Add clarifying language around the use of the Enum module

Theosaurus-Rex:
# Contributor checklist

- [x] Documentation changes

In the existing docs for the Enum module, the language around the provided examples wasn't immediately clear that the first example would be your code *before* utilising the module. This adds some small changes that clarify this


zachdaniel:

PR #348, opened 2022-07-07
Use a proper operator in the `attribute_does_not_equal` validation

smt116:
Previously, it was using `if current != expected do raise` while it
needs to do the opposite. Also, simplify the logic by making `update`
work in the same way as `create` (compare upcoming value - not the
original one).

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #357, opened 2022-07-21
Improve module names

zimt28:
It wasn't obvious where to look for the test resources, the module names now match the file paths


zimt28:
Should it be `SimplePolicy` / `RbacPolicy` instead of `PolicySimple` / `PolicyRbac`? I used the latter to "group" the files

zachdaniel:
I think `PolicySimple`/ `PolicyRbac` is good 👍

PR #364, opened 2022-08-22
fix: False default value for argument is nil in changeset

totaltrash:
I have defined an argument for an action:

      argument :some_boolean, :boolean, allow_nil?: false, default: false

When I retrieve that argument in a changeset, the value is `nil`.

If the default is `true`, it correctly comes in as `true`.

This pr contains a failing test for this scenario

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


totaltrash:
now with a potential fix

zachdaniel:

PR #371, opened 2022-08-27
Fix actions documentation

zimt28:


zimt28:
Nevermind, I just assumed the code interface would work differently.

PR #376, opened 2022-09-11
feature: generate mermaid entity relationship diagrams from a given api

joshprice:
Initial cut of generating ERDs in Mermaid from a given API. Other things to look at:

- show attrs
- show private attrs
- show private relships
- show calcs, aggs
- mix task to generate similar to flow chart

# Contributor checklist

### DIFF
--- a/lib/ash/api/diagram/diagram.ex
+++ b/lib/ash/api/diagram/diagram.ex
@@ -12,22 +12,9 @@ defmodule Ash.Api.Info.Diagram do
   def normalise_relationships(api) do
     for resource <- Ash.Api.Info.resources(api) do
       for relationship <- Ash.Resource.Info.public_relationships(resource) do
-        # All relationships will be expressed from both sides and therefore appear twice
-        # Normalise them so we can dedup
-        case relationship.type do
-          :belongs_to ->
-            {relationship.destination, :has_many, relationship.source}
-
-          :many_to_many ->
-            if relationship.destination <= relationship.source do
-              {relationship.destination, :many_to_many, relationship.source}
-            else
-              {relationship.source, :many_to_many, relationship.destination}
-            end
-
-          _ ->
-            {relationship.source, relationship.type, relationship.destination}
-        end
+        [relationship.source, relationship.destination]
+        |> Enum.sort()
+        |> List.to_tuple()
       end
     end
     |> Enum.flat_map(& &1)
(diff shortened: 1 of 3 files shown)


zachdaniel:
A couple interesting things here. I'd generally go with `string[]` as opposed to `[string]`. Aside from that, it looks like mermaid doesn't support special chracters in types, so we're going to have to (at least for now) do something like `stringArray` or `ArrayOfString`.

The last thing to talk about is the relationships. I definitely see what you're saying about not deduplicating the relationships being problematic, but its also typically very important information. Resource A may have multiple relationships to Resource B. Perhaps there is a creative way to show that, if we can label the edges in that way. We could group them up, and show one line both ways, that shows all of the relationship names, potentially?

zachdaniel:
Most importantly: This is *awesome* 🥳

zachdaniel:
Okay, so if we switch to class diagrams, I think we get better features in general: https://mermaid-js.github.io/mermaid/#/classDiagram?id=syntax

Including special characters

zachdaniel:
Here is an example of what that looks like, with array types and labeled relationships: https://mermaid.live/edit#pako:eNptkMFugzAMhl8l8pmWARttc97uk6pdNqbKJQGiQYwSU6mqePcF6NZtmi-Jf3_5HfsCJSkNEsoWvX80WDvsCitCzIp48dqJy6JMsWdnbC0sdnoRx5_0M3n-hz6SOv8V394FY-1_mczNaLWixUiKPhw-EhWeyBnWhylf0BmY0fmRFDhwQ-7GqsMQCld_iKDTrkOjwqjz_wrgRocZQIarQvdRQGHHwA29QtZPyjA5kBW2XkcQzGl_tiVIdoP-gq7r-qZ6tK9Et7wlVDqYXIDP_bTk2ngOLUqylaknfXBtkBvm3ss4nsrr2nAzHNcldbE3qkHHzWmXx3mabzHNdL7J8CHLVHlMdtsqvU8qtblLUoRxHD8B1x2VJw

zachdaniel:
This looks great! Is this read to merge @joshprice ?

joshprice:
I think so, it's missing the mix tasks. Your call if you want to merge this and I can start a new PR.

zachdaniel:
Lets merge it :). As for the mix task, I'd take a look at the `generate_flow_charts` task for inspiration.

zachdaniel (approved):

zachdaniel:

Issue #319, opened 2022-04-27
Forbid special characters in field names


zachdaniel:
Too many extensions currently break if you add anything other than alphanumeric and underscores to an attribute/calculation/aggregate/relationship name. For now, we should simply prevent this at the resource level.

alimnastaev:
@zachdaniel what would be the best/quickest way to reproduce that behavior above?
Thanks a lot!

zachdaniel:
👋🏻  Hello!

The simplest thing to do would be to add an attribute with a question mark at the end, e.g `attribute :foo?, :boolean`. This should fail. Then we'd add a transformer just like this one: https://github.com/ash-project/ash/blob/main/lib/ash/resource/transformers/require_unique_field_names.ex

except instead of validating that they are all unique, it would simply validate that none of them have special characters :)

alimnastaev:
Wow! Awesome! Thanks a lot for the super quick response!
@zachdaniel 🥇

juhalehtonen:
I started looking into this one as my first foray into Ash. 

I followed your suggestion @zachdaniel, and I managed to at least get the system to raise if I give an attribute a `:name` that has unacceptable characters. I tested this out by altering an existing test script:

```elixir
== Compilation error in file test/actions/create_test.exs ==
** (Spark.Error.DslError) [Ash.Test.Actions.CreateTest.Profile]
 The field name biöÖo contains special characters

    (ash 2.0.0-rc.9) lib/ash/resource/transformers/require_no_special_chars_in_field_names.ex:39: anonymous fn/1 in Ash.Resource.Transformers.RequireAlphanumericFieldNames.transform/1
....
```

Before I look into opening a PR, I'd like to hear where tests for transformers should be placed. I couldn't quite find a suitable location for testing for my new `Ash.Resource.Transformers.RequireAlphanumericFieldNames` module (naming TBD). Any tips on organization? 🙂

zachdaniel:
There currently aren't any transformer specific tests, moreso integration tests on them running whenc compiling a resource. For example: https://github.com/ash-project/ash/blob/main/test/resource/attributes_test.exs

[...]

PR #388, opened 2022-09-27
Forbic special characters in field names

juhalehtonen:
This PR aims to address issue #319. 

Per the Contributor checklist below, this PR should probably also update the documentation somewhere to state that such names are no longer accepted?

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/transformers/require_alphanumeric_field_names.ex
+++ b/lib/ash/resource/transformers/require_alphanumeric_field_names.ex
@@ -1,0 +1,49 @@
+defmodule Ash.Resource.Transformers.RequireAlphanumericFieldNames do
+  @moduledoc """
+  Confirms that a resource does not have unaccepted characters (characters other than alphanumeric and underscores) in field names.
+  """
+  use Spark.Dsl.Transformer
+
+  alias Spark.Dsl.Transformer
+  alias Spark.Error.DslError
+
+  def transform(dsl_state) do
+    attributes =
+      dsl_state
+      |> Transformer.get_entities([:attributes])
+
+    relationships =
+      dsl_state
+      |> Transformer.get_entities([:relationships])
+
+    calculations =
+      dsl_state
+      |> Transformer.get_entities([:calculations])
+
+    aggregates =
+      dsl_state
+      |> Transformer.get_entities([:aggregates])
+
+    attributes
+    |> Enum.concat(relationships)
+    |> Enum.concat(calculations)
+    |> Enum.concat(aggregates)
+    |> Enum.group_by(& &1.name)
+    |> Enum.each(fn {name, _groups} ->
+      unless name_acceptable?(name) do
+        raise DslError.exception(message: "The field name #{name} contains special characters")
+      end
+    end)
+
+    {:ok, dsl_state}
+  end
+
+  def after?(Ash.Resource.Transformers.BelongsToAttribute), do: true
+  def after?(_), do: false
+
+  defp name_acceptable?(name) do
+    name
+    |> Atom.to_string()
+    |> String.match?(~r/^[[:alnum:]_]+$/)
+  end
+end
--- a/test/resource/attributes_test.exs
+++ b/test/resource/attributes_test.exs
@@ -91,5 +91,27 @@ defmodule Ash.Test.Resource.AttributesTest do
         end
       )
     end
+
+    test "does not raise if you pass underscore in `name`" do
+      defposts do
+        attributes do
+          attribute :foo_bar, :string
+        end
+      end
+    end
+
+    test "raises if you pass an invalid value for `name`" do
+      assert_raise(
+        Spark.Error.DslError,
+        "[Ash.Test.Resource.AttributesTest.Post]\n The field name foo? contains special characters",
+        fn ->
+          defposts do
+            attributes do
+              attribute :foo?, :string
+            end
+          end
+        end
+      )
+    end
   end
 end
--- a/lib/ash/resource/dsl.ex
+++ b/lib/ash/resource/dsl.ex
@@ -1210,6 +1210,7 @@ defmodule Ash.Resource.Dsl do
     Ash.Resource.Transformers.DefaultAccept,
     Ash.Resource.Transformers.SetTypes,
     Ash.Resource.Transformers.RequireUniqueFieldNames,
+    Ash.Resource.Transformers.RequireAlphanumericFieldNames,
     Ash.Resource.Transformers.ValidateRelationshipAttributes,
     Ash.Resource.Transformers.ValidateEagerIdentities,
     Ash.Resource.Transformers.ValidateAggregatesSupported


test/resource/attributes_test.exs
```diff
@@ -91,5 +91,27 @@ defmodule Ash.Test.Resource.AttributesTest do
         end
       )
     end
+
+    test "does not raise if you pass underscore in `name`" do
```
juhalehtonen, line 95: I am not sure if this test is necessary or wanted in the final PR. I wasn't too sure what kind of a convention Ash has for tests, so I wanted to raise the discussion here.

lib/ash/resource/transformers/require_alphanumeric_field_names.ex
```diff
@@ -0,0 +1,49 @@
...
+      |> Transformer.get_entities([:aggregates])
+
+    attributes
+    |> Enum.concat(relationships)
+    |> Enum.concat(calculations)
+    |> Enum.concat(aggregates)
+    |> Enum.group_by(& &1.name)
+    |> Enum.each(fn {name, _groups} ->
+      unless name_acceptable?(name) do
+        raise DslError.exception(message: "The field name #{name} contains special characters")
```
juhalehtonen, line 34: The message is a bit placeholdery and not too informative (it doesn't say what "special characters" are). What would you hope to see in it?

test/resource/attributes_test.exs
zachdaniel: I think its great 👍🏻 makes it much more explanatory if this one doesn't fail and the other does, and vice versa.

lib/ash/resource/transformers/require_alphanumeric_field_names.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Ash.Resource.Transformers.RequireAlphanumericFieldNames do
```
juhalehtonen, line 1: The module name does not represent the truth of accepted values (underscore is not alphanumeric). A better name suggestion is very welcome :)

lib/ash/resource/transformers/require_alphanumeric_field_names.ex
zachdaniel: A couple things here: 

So calculations/aggregates/relationships/attributes all have to have unique names anyway, so I wouldn't worry about that.

What you can do is something like this:
```elixir
   resource = Transformer.get_persisted(dsl_state, :module)

    attributes
    |> Enum.concat(relationships)
    |> Enum.concat(calculations)
    |> Enum.concat(aggregates)
    |> Enum.each(fn  %{name: name} = field -> 
      path = 
        case field do
          %Ash.Resource.Aggregate{type: type} -> [:aggregates, type, name]
          %Ash.Resource.Calculation{} -> [:calculations, name]
          %Ash.Resource.Attribute{} -> [:attributes, name]
          %{type: type} -> [:relationships, type, name] 
        end
      
      raise DslError, message: "your message is good, but add the regex and/or just say the rules", path: path, module: resource
    end)
```

This sets a path & module to help point people where the error is.

Additionally, instead of `def after?(...)` in your case this is just a validation that can be run in the after compile step, so remove that and add `def after_compile?, do: true`, and then you're good 🔥

lib/ash/resource/transformers/require_alphanumeric_field_names.ex
zachdaniel: Perhaps `ValidateFieldNames`?

juhalehtonen:
I resolved the outdated conversations, as none of them were exactly accurate anymore with the new goal of this PR (forbidding reserved names instead of special characters).

lib/ash/resource/transformers/no_reserved_field_names.ex
```diff
@@ -0,0 +1,69 @@
+defmodule Ash.Resource.Transformers.NoReservedFieldNames do
+  @reserved_names [
```
juhalehtonen, line 2: This might be a bit of an unorthodox approach, but I wanted to have a source of truth that can be referenced in both the moduledocs and the error.

test/resource/attributes_test.exs
```diff
@@ -91,5 +91,19 @@ defmodule Ash.Test.Resource.AttributesTest do
         end
       )
     end
+
+    test "raises if you pass a reserved name for `name`" do
```
juhalehtonen, line 95: This should be considered unsatisfactory testing (we of course need to test for all of the reserved names). 

I wanted to iterate over all reserved names in a for-comprehension, but I had trouble getting my tests to work as I wanted to.

It seems like I need to create a new module for each iteration, as otherwise I'll get
```elixir
     test/resource/attributes_test.exs:23
     ** (ArgumentError) field/association :__metadata__ is already set on schema
     stacktrace:
       (ecto 3.7.2) lib/ecto/schema.ex:2146: Ecto.Schema.put_struct_field/3
       (ecto 3.7.2) lib/ecto/schema.ex:1899: Ecto.Schema.define_field/4
       (stdlib 3.17.1) erl_eval.erl:685: :erl_eval.do_apply/6
       (stdlib 3.17.1) erl_eval.erl:123: :erl_eval.exprs/5
       (stdlib 3.17.1) erl_eval.erl:919: :erl_eval.try_clauses/8
       (stdlib 3.17.1) erl_eval.erl:123: :erl_eval.exprs/5
       (elixir 1.13.0) src/elixir.erl:289: :elixir.recur_eval/3
       (spark 0.1.25) /home/juha/projects/elixir/ash/test/resource/attributes_test.exs:24: Spark.Dsl.__before_compile__/1
```

However, despite my attempts, I couldn't get this to work right (yet).

lib/ash/resource/transformers/no_reserved_field_names.ex
zachdaniel: https://i.ytimg.com/vi/LaiN63o_BxA/maxresdefault.jpg

zachdaniel: Its perfect, this is very much aligned with the way we do things with Ash :D

test/resource/attributes_test.exs
zachdaniel: Yeah, I for loop would be ideal, but its not too big of a deal

juhalehtonen: I'd also prefer us to have a neat loop, so I dug a bit more!

I pushed a new commit to demonstrate the problem I'm having, and I figured this might be the best avenue to ask if you have other ideas. 

As you can see in the latest commit, I identified my "problem" to be the return value of the `after_compile?` function. By setting it to `false`, I was able to sidestep my issue where otherwise I'd get the "is already set on schema" ArgumentError. 

As it seems to me that we want the return value to remain `true`, I want to ask: can you come up with a workaround for this case, or explain why specifically I might be running into this?

zachdaniel: Yeah, okay. So the issue is that `after_compile?` transformers happen after we build the ecto schema, which is a pretty interesting problem. I think we want to retain the nice error messages, so what we ought to do is just ignore any overlaps in the schema. What we can do to make that happen is probably here:

https://github.com/ash-project/ash/blob/main/lib/ash/resource/schema.ex#L92

And just say `struct_fields = Module.get_attribute(__MODULE__, :ash_struct_fields) |> Enum.reject(<is_reserved>)`

zachdaniel: That same code exists in the other branch of that `if` so it has to be done there as well.

juhalehtonen: Thanks Zach! I think I figured this one out, and it doesn't seem to have been about that particular line after all (unless I'm mistaken, of course 😄).

If you have a look at my latest commit, you'll see that I now filter the attributes out at the very beginning, before we call to `Ecto.Schema.field/3`. I realized that we were doing that against each user-provided attribute, and then attempting to call `Ecto.Schema.field/3` for some of the reserved field names right below. This obviously didn't work, and makes perfect sense in the light of the error I was seeing.

What I do still want to fix about this PR is the scattering of knowledge: now I'm just hard-coding the list of reserved names in `lib/ash/resource/schema.ex`, but ideally we'd have a single source to reference. However, I don't have quite the mental model of Ash to best judge where to place such information so that both `lib/ash/resource/schema.ex` and `lib/ash/resource/transformers/no_reserved_field_names.ex` could reference the same source, as well as the test at `test/resource/attributes_test.exs`. Any suggestions on this?

zachdaniel: That makes sense, but you will need to do the same for aggregates, relationships, and calculations in both branches as well :)

In terms of where the list lives, I think its fine to make it public and putting it in `Ash.Resource` sounds like a good idea :)

juhalehtonen: Thanks! 

I pushed an update to relocate the list of reserved names, but I wasn't quite sure about your suggestion on "do the same for aggregates, relationships, and calculations in both branches as well". The PR currently does this for both branches (for both `schema` and `embedded_schema`), and the fields for aggregates, relationships, and calculations seem to be within these two schemas. Are you referring to something else that I'm not getting here?

I also saw that Dialyzer is failing, but that seems to be true for the master branch as well.

zachdaniel: https://github.com/ash-project/ash/pull/388/files#diff-0444918c54e2c63a6c07cd12fcd2da2e8e33119c56afccbf57d5a2ce3131d66cL156

When building the schema, we add virtual fields for aggregates and calculations, so those would need to be updated as well not to use reserved names

zachdaniel: Sorry, here is a better link: https://github.com/ash-project/ash/blob/main/lib/ash/resource/schema.ex#L68

zachdaniel:
Awesome! This looks great, will merge shortly.,

PR #395, opened 2022-10-03
improvement: validate accepted and rejected attributes in actions

### LINKED ISSUE #366: Validate accept list only contains attributes
RIght now if the `accept` list on an action contains non-attribute keys they are just ignored, which can lead to confusion looking at an action thinking you can provide that value. A simple transformer should be able to check each action and confirm that there is a public attribute corresponding to each value in `accept`. It can raise one error if there is no attribute, and another if the attribute exists but is private. This transformer should be `def after_compile?(), do: true`

SpaceEEC:
This PR aims to resolve #366 by raising an error if an attribute name specified in an `accept` or `reject` field of an action is either private or not an attribute at all.

I'm not sure if there is anything else that needs to be done here (looking at the contributor checklist).

I'm also not sure if my change in `default_accept` is me not understanding something or me finding a bug.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/transformers/validate_accept.ex
+++ b/lib/ash/resource/transformers/validate_accept.ex
@@ -38,6 +38,7 @@ defmodule Ash.Resource.Transformers.ValidateAccept do
           end
         end
 
+        if accept != :all do
         Enum.each(
           accept,
           &validate_attribute_name.(
@@ -45,6 +46,9 @@ defmodule Ash.Resource.Transformers.ValidateAccept do
             :accept
           )
         )
+        end
+
+        if reject != :all do
 
         Enum.each(
           reject,
@@ -53,6 +57,7 @@ defmodule Ash.Resource.Transformers.ValidateAccept do
             :reject
           )
         )
+        end
 
         # read types do not have accept / reject fields
       %{type: :read} ->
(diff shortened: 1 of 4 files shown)


lib/ash/resource/transformers/validate_accept.ex
```diff
@@ -0,0 +1,68 @@
...
+              :ok
+
+            true ->
+              raise DslError,
+                path: [:actions, action_name, type, attribute_name],
+                message: "#{attribute_name} is not an attribute"
+          end
+        end
+
+        if accept != :all do
```
zachdaniel, line 41: `accept` should never be `:all` here (same for `reject`), because an earlier transformer should remove it. Are you seeing it as `:all` at this point?

SpaceEEC: I am, this test here is failing without those `if`s:
https://github.com/ash-project/ash/blob/14a1b92963beafe9d91586539b2cf5f3d89d73a1/test/resource/actions/actions_test.exs#L45-L65

SpaceEEC: Looking further into it, it seems like the earlier transformer only updates `accept`, but not `reject`.
I assume `reject` is not needed after updating `accept`, so it isn't updated.

zachdaniel:
@SpaceEEC ideally, if we could actually just replace `reject` in that previous transformer that would be ideal. Would you mind doing that as part of this PR?

SpaceEEC:
I think that should do it.

zachdaniel:

zachdaniel:
If you are in the discord, please let me know who you are so I can give you the contributor role :)

PR #399, opened 2022-10-04
feat: list arguments for resource actions in class diagrams

### LINKED ISSUE #392: Show arguments (maybe only required arguments) in generated class diagrams
At the moment, for something like

```elixir
create :register do
  argument :email, :ci_string do
    allow_nil? false
  end
  
  argument :password, :ci_string do
     allow_nil? false
  end
  
  argument :confirm_password, :ci_string do
    allow_nil? false
  end
end
```

We would only see something like

```
register()
```

in the generated diagram. Would be great to show arguments there.

AliNisarAhmed:
This PR resolves #392 by printing arguments and attributes for resource actions in class diagrams, with an upper limit of 4 (the rest hidden with ellipsis)

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/api/diagram/diagram.ex
+++ b/lib/ash/api/diagram/diagram.ex
@@ -13,6 +13,7 @@ defmodule Ash.Api.Info.Diagram do
 
   @indent "    "
   @show_private? false
+  @argument_print_limit 4
 
   @default_opts indent: @indent, show_private?: @show_private?
 
@@ -123,6 +124,28 @@ defmodule Ash.Api.Info.Diagram do
     Enum.map_join(list, "\n", fn item -> "#{indent}#{indent}#{template_fn.(item)}" end)
   end
 
+  defp list_arguments_and_attributes(resource, action) do
+    arguments = action.arguments
+
+    attributes =
+      case action.type do
+        :read -> []
+        _ -> action.accept |> Enum.map(&Ash.Resource.Info.attribute(resource, &1))
+      end
+
+    list_arguments(arguments ++ attributes)
+  end
+
+  defp list_arguments(arguments) when length(arguments) > @argument_print_limit do
+    {displayed_args, _rest} = Enum.split(arguments, @argument_print_limit)
+    list_arguments(displayed_args) <> ", ..."
+  end
+
+  defp list_arguments(arguments) do
+    arguments
+    |> Enum.map_join(", ", &"#{class_short_type(&1.type)} #{&1.name}")
+  end
+
   @doc """
   Generates a Mermaid Class Diagram for a given API.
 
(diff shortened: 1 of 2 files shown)


lib/ash/api/diagram/diagram.ex
```diff
@@ -123,6 +123,15 @@ defmodule Ash.Api.Info.Diagram do
     Enum.map_join(list, "\n", fn item -> "#{indent}#{indent}#{template_fn.(item)}" end)
   end
 
+  defp list_arguments([]), do: ""
+
+  defp list_arguments(arguments) do
+    arguments
+    |> Enum.map(&"#{class_short_type(&1.type)} #{&1.name}")
```
joshprice, line 130: You'll want to use `Enum.map_join` like above. Credo whinges about this as I discovered.

joshprice (commented):
Looking good so far

lib/ash/api/diagram/diagram.ex
AliNisarAhmed: Fixed, thanks

zachdaniel:
After the arguments, we'll also want to list the accepted attributes (for non-read actions). Something like:

```elixir
action.accept
|> Enum.map(&Ash.Resource.Info.attribute(resource, &1))
# [%Attribute{type: type, name: name, allow_nil?: true/false}, ...]
```

AliNisarAhmed:
@zachdaniel @joshprice Let me know from below diagrams which one you prefer. 

I am limiting the shown arguments to 4, and the rest either turn to ellipsis, or I show their count.

Personally I prefer `+ n more` style, with default values. We could provide options for users to enable/disable printing of types and default args.

Also, I could not find a tool-tip option for Mermaid class diagrams. I think it is only an option for Flowcharts.


- Ellipsis: 

ellipsis

- Ellipsis with default arguments: 

ellipsis_with_default_args

- `+ n` more style 

plus_more

- `+ n` more with default args: 

plus_more_with_default

joshprice:
I like the ellipsis stylistically but also think the extra info is nice too. What about both with `UUID x_id, ...+2 more)`?

AliNisarAhmed:
@zachdaniel @joshprice The diagram below prints both arguments and attributes, and limits arguments to max of 4, and shows the count of rest of the args, omitting the default values to reduce noise. Let me know how it looks now

image

zachdaniel:
I think we should do a *total limit* of 4, for both arguments and attributes, and they should all go before the ellipsis.

zachdaniel:
That looks excellent!

AliNisarAhmed:
@zachdaniel I would have added tests, but I think the current class diagram tests cover all cases, showcasing no arguments, 4 arguments, and 4+ arguments, so I think they suffice, let me know. 

If things look fine I will squash the commits and open the PR for review.

zachdaniel:
I agree :) Lets get it merged 🚀

AliNisarAhmed:
@zachdaniel ready for review

zachdaniel:

PR #405, opened 2022-10-07
Improve error when actions reject and accept keys overlap

### LINKED ISSUE #397: Improve error when actions reject and accept keys overlap
Same key in  a resources actions accept and reject property yields the following runtime `cannot be changed` error:
```
** (Ash.Error.Invalid) Input Invalid

* Invalid value provided for subject: cannot be changed.
```

It may be worthwhile to change the error into a compile time error with a message like so:
```
** (Spark.Error.DslError) <Resource.Module>

* Keys for actions accept and reject property may not overlap.
```

**Example code**
Below some code for reproducing this behaviour. First define the modules likes so:
```elixir
defmodule Resource do
  use Ash.Resource,
    data_layer: Ash.DataLayer.Simple

  attributes do
    uuid_primary_key :id

    attribute :resource_name, :string
  end

  actions do
    create :name do
      accept [:resource_name]
      reject   [:resource_name]
    end
  end
end

defmodule Registry do
  use Ash.Registry,
    extensions: [
      Ash.Registry.ResourceValidation
    ]

   entries do
     entry Resource
   end
end

defmodule Api do
  use Ash.Api

  resources do
    registry Registry
  end
end
```
Then attempt to execute the `:name` create action.
```elixir
Resource
|> Ash.Changeset.for_create(:name, %{resource_name: "Resource #1"})
|> Api.create!()
```

noteven:
fixes #397

Implements a transformer to check if an actions reject and accept keys are not mutually exclusive. No check is done in regards to `default_accept` as `reject` and `accept` will always be a subset of `default_accept`(which defaults to all keys), but must be mutually exclusive in regards to each other.

Also error message could be improved.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ X] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/resource/transformers/validate_accept_reject_overlap.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Ash.Resource.Transformers.ValidateAcceptRejectOverlap do
```
zachdaniel, line 1: I think this logic would be better just as part of `ValidateAccept`.

noteven:
Moved check to `DefaultAccept` transformer, code could probably be made more idiomatic instead of just an isolated if.

zachdaniel (dismissed):

zachdaniel:
LGTM if you run the formatted we can get it merged

PR #408, opened 2022-10-07
Validate attribute names for builtin validations

zimt28:
This currently validates attribute names for builtin validations by checking the `attribute`/`attributes` field. There are also validations with a `field` field, but I've skipped it as the list of options would depend on the validation.

I've marked this as draft as the implementation is a little fragile – it depends on the observation and assumption that all validations with a `attribute`/`attributes` field have more than one option and the attribute field comes first. On the other hand it doesn't worsen anything so it's fine I guess.

### DIFF
--- a/lib/ash/resource/transformers/validate_validation_attributes.ex
+++ b/lib/ash/resource/transformers/validate_validation_attributes.ex
@@ -1,0 +1,52 @@
+defmodule Ash.Resource.Transformers.ValidateValidationAttributes do
+  @moduledoc """
+  Validates that all built in validations point to valid attributes
+  """
+  use Spark.Dsl.Transformer
+  alias Spark.Dsl.Transformer
+
+  @impl true
+  def after_compile?, do: true
+
+  @impl true
+  def transform(dsl) do
+    attribute_names =
+      dsl
+      |> Transformer.get_entities([:attributes])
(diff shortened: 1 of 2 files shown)


lib/ash/resource/transformers/validate_validation_attributes.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Ash.Resource.Transformers.ValidateValidationAttributes do
+  @moduledoc """
```
zachdaniel, line 2: I think this will be too generic, as there are some cases where validation modules may use the key `attributes` but may actually support additional things (i.e arguments). This would also check any non-built in validations and we can't make any assertions about what the `attributes` key means for them.

In the long run, I'd like to solve this via giving validations away to validate their options against a resource. Currently validations have an `init/1` callback, which is meant to validate options, but I'm thinking that we should allow them to have an `init/2` callback that takes the options and the resource and validates them. However, we can't do this right now because that would cause compilations issues, which we need to figure out a way around.

In the short term, I think this validation needs to be run against only specific, known validations, and those validations should be double-checked to ensure that they actually only accept attributes.

Additionally, we should check the validations on the action, which are in the `.changes` key of each action (would need to filter out the ones that are validations.

PR #409, opened 2022-10-09
fix: allow passing query in can/can?/4

frankdugan3:
# Contributor checklist

`Ash.Policy.can/4` and `Ash.Policy.can?/4` need to be able to accept a query in order to support complex reads, such as those w/ required arguments. This PR adds support to pass a query instead of an action/action name.


lib/ash/policy/info.ex
```diff
@@ -146,11 +146,17 @@ defmodule Ash.Policy.Info do
   See the documentation of `can/4` for more.
   """
   @type can_option? :: {:api, module} | {:maybe_is, boolean()}
-  @spec can?(Ash.Resource.t(), atom(), map() | nil, list(can_option?())) :: boolean()
-  def can?(resource, action_or_action_name, actor, opts \\ []) do
+
+  @spec can?(
+          Ash.Resource.t(),
```
zachdaniel, line 151: This should probably also accept a changeset in the same way.

frankdugan3: Makes sense, will add that to the mix later.

Just to make sure I grep, as in `can(resource, changeset, actor, opts)`?

zachdaniel: Yep!

PR #412, opened 2022-10-12
fix: allow overriding validation message on a list of errors

### LINKED ISSUE #411: Combining `present` validation with `message` option no longer stores the validation error
**Describe the bug**

This is a weird one. When I supply the `message` option to a `present` validation, it's like the error message (and the entire error) is thrown away after checking the validation. 

Without the `message` option, it's fine, and the `message` option works for a `compare` validation, so it must be something specific to `present`?

**To Reproduce**

I have this as part of my resource:

```elixir
attributes do
  attribute :duration_in_seconds, :integer
end

validations do
  validate present(:duration_in_seconds)
end
```

And validating a form with the field and invalid data gives a message as expected:

```elixir
Form.errors(form) #=> [duration_in_seconds: "must be present"]
```

If I add the `message` option:

```elixir
validations do
  validate present(:duration_in_seconds), message: "the duration is missing"
end
```

And reload/resubmit the form with invalid data:

```elixir
Form.errors(form) #=> []
```

The form doesn't pass validation, but no message is provided for the invalid field.

**Expected behavior**

If I have:

```elixir
  validate present(:duration_in_seconds), message: "the duration is missing"
```

And the validation fails, I'd expect to get 

```elixir
Form.errors(form) #=> [duration_in_seconds: "the duration is missing"]
```

**Runtime**
 - Elixir version 1.14.0-otp-25
 - Erlang version 25.0.4
 - OS MacOS Monterey 12.6
 - Ash version 2.0.0-rc.10
 - any related extension versions latest published versions of all of them

sevenseacat:
As described in #411, this fixes an issue where applying a custom validation message for a validation that returns a list of errors, would cause weirdnesses to happen.

Because the `present` validation takes a list of fields to validate, it returns a list of errors, possibly including one for each field. 

This worked when no message was provided - the list of `Ash.Error.Changes.InvalidAttribute` errors was passed to `Ash.Changeset.add_error/3` and all was fine - but when a message was provided, the list of errors was converted to a single string (the message) and that was stored on the changeset instead.

Now each element of the list gets the same treatment as a single error, and it seems to work well in my testing! Closes #411 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1156,29 +1156,22 @@ defmodule Ash.Changeset do
 
             {:error, error} when is_exception(error) ->
               if validation.message do
-                error =
-                  case error do
-                    %{field: field} when not is_nil(field) ->
-                      error
-                      |> Map.take([:field, :vars])
-                      |> Map.to_list()
-                      |> Keyword.put(:message, validation.message)
-                      |> InvalidAttribute.exception()
-
-                    %{fields: fields} when fields not in [nil, []] ->
-                      error
-                      |> Map.take([:fields, :vars])
-                      |> Map.to_list()
-                      |> Keyword.put(:message, validation.message)
-                      |> InvalidChanges.exception()
-
-                    _ ->
-                      validation.message
-                  end
-
+                error = override_validation_message(error, validation.message)
                 Ash.Changeset.add_error(changeset, error)
               else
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):
Awesome work!

zachdaniel:

PR #417, opened 2022-10-14
feat: mermaid class diagram mix task

joshprice:
- refactor: remove unused sibling_file fn
- refactor: extract mermaid helpers and simplify
- feat: mix task to generate mermaid class diagrams


zachdaniel:
Perhaps we should rework the diagram type to be an option to a single task? Like `mix ash.generate_resource_charts --type er` or `--type class`?

joshprice:
> Perhaps we should rework the diagram type to be an option to a single task? Like `mix ash.generate_resource_charts --type er` or `--type class`?

Maybe, not sure what other diagrams might be possible yet, so maybe defer that one for now?

zachdaniel:

PR #420, opened 2022-10-16
feat: basic livebook generator and mix task

joshprice:
Basic Livebook generation and mix task.

Some refactoring required around utility functions, as the mermaid tasks are using too.

### DIFF
--- a/lib/ash/api/info/livebook.ex
+++ b/lib/ash/api/info/livebook.ex
@@ -3,6 +3,13 @@ defmodule Ash.Api.Info.Livebook do
   Generate a Livebook from a specified API.
   """
 
+  # Strip Elixir off front of module
+  defp module_name(module) do
+    module
+    |> Module.split()
+    |> Enum.join(".")
+  end
+
   # TODO: move to Ash.Resource.Info as it's also used in diagram
   defp resource_name(resource) do
     resource
@@ -27,7 +34,7 @@ defmodule Ash.Api.Info.Livebook do
 
   def api_section(api) do
     """
-    ## #{api}
+    ## #{module_name(api)}
 
     ### Class Diagram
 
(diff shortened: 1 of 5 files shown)


zachdaniel:
I think that perhaps this should go in https://github.com/ash-project/ash_livebook? Mainly because there is going to be lots of livebook stuff over time. Alternatively we can make all the livebook stuff live in core...

joshprice:
I thought about both ways and then settled on the simplest approach for now. At it's simplest it's really just generate a simple 1 page livebook, but this could easily get more fancy.

Happy to keep this in draft for further discussion.

joshprice:
Update, 😆 I didn't realise you'd already made https://github.com/ash-project/ash_livebook will take a look.

joshprice:
Think an extension is probably the place to keep smartcells, but not planning on using any for the basic generator.

zachdaniel:
Yeah, but splitting up livebook responsibilities seems like a bad idea. We have `ash_livebook` started, so we could just put this in there. If we think `ash_livebook` will exist, we'll have to move this there later.

zachdaniel:
@joshprice ash_livebook is pretty barren, just musings about a smart cell that doesn't actually do anything, but we can fix that :D

joshprice:
If it is an extension, can it work globally? eg not at a resource by resource level, haven't seen an example of a globally applied extension that I can recall

zachdaniel:
Well, the term is a bit nebulous. An actual "extension" only extends one thing, i.e a resource. Probably better to call things like `ash_postgres` and `ash_graphql` a "package" which contains a set of "extensions", and may include code that operates on groups of things.

joshprice:
Even configuring at API level or above seems less desirable than deciding the livebook config at generation time with mix task args. So a package doesn't have to have any code config?

zachdaniel:
Its just an elixir library at the end of the day. So "package" is wrong also lol. Its just a library on hex.

joshprice:
Right, so could ash core depend on ash_livebook and therefore expose the functionality by default?

zachdaniel:
It *could* but if we're going to do that we should just put it in core.

joshprice:
Feel like this is probably the least friction approach from a user and usability perspective

zachdaniel:
🤔 I think you're probably right, I'm leaning towards having it in core as well. Want to think it over a bit more.

PR #423, opened 2022-10-19
feat: Custom short names for types

zimt28:


lib/ash/type/type.ex
```diff
@@ -50,9 +54,13 @@ defmodule Ash.Type do
 
   ## Built in types
 
-  #{Enum.map_join(@short_names, fn {key, module} -> "* `#{inspect(key)}` - `#{inspect(module)}`\n" end)}
+  #{Enum.map_join(@builtin_short_names, fn {key, module} -> "* `#{inspect(key)}` - `#{inspect(module)}`\n" end)}
+
+  #{if @custom_short_names != [], do: "## Custom types"}
```
zachdaniel, line 59: Honestly this is probably unnecessary. When generating hex docs for Ash, it will always be empty, whereas when generating hex docs in user's applications they won't be generating it for dependencies. I think we can remove this.

zachdaniel (dismissed):
one comment on the docs, otherwise good to go.

PR #424, opened 2022-10-19
improvement: Add `quarter` value to duration names

zimt28:


zachdaniel:
I'm far from the expert on time conversions, would be good to check with @mbuhot who has some expertise there :)

mbuhot:
Is `quarter` 3 months? If so we can handle it in `datetime_add` like:

```
  def datetime_add(datetime, amount_to_add, :quarter) do
    datetime_add(datetime, 3 * amount_to_add, :month)
  end
```

zachdaniel:
🤔 its a good question really. If thats how we'd do it, I'd probably rather have people just do 3 months? There are always weird cases with these kinds of things. Like is a quarter one quarter of a day longer on leap years 😆

zachdaniel:
Yeah, I think unless we're absolutely certain how this should work then its best not to add it. I could be convinced, but just want to have a clear specification for it is all.

PR #428, opened 2022-10-21
improvement: SVG, PDF, PNG, Markdown and plain mermaid formats

### LINKED ISSUE #422: Allow defining output format for ash.generate_flow_charts mix task
Currently the `ash.generate_flow_charts` mix task depends on the Mermaid CLI being installed on the users system, and only the PDF output format is supported.

By allowing users to pass in a `--format` option of `.svg`, `.pdf` or `.md` to the mix task, we could both 1) remove the hard dependency on Mermaid CLI when just the text-based output is needed, and 2) enable SVG file format as an alternative output option.

joshprice:
Support a variety of different formats for all mermaid diagrams and flowcharts.

Closes #422

### DIFF
--- a/lib/mix/tasks/generate_flow_charts.ex
+++ b/lib/mix/tasks/generate_flow_charts.ex
@@ -15,6 +15,13 @@ defmodule Mix.Tasks.Ash.GenerateFlowCharts do
 
     * `--only` - only generates the given Flow file
 
+    * `--format` - Can be set to one of either:
+      * `plain` - Prints just the mermaid output as text. This is the default.
+      * `md` - Prints the mermaid diagram in a markdown code block.
+      * `svg` - Generates an SVG
+      * `pdf` - Generates a PDF
+      * `png` - Generates a PNG
+
   """
   use Mix.Task
 
@@ -22,12 +29,15 @@ defmodule Mix.Tasks.Ash.GenerateFlowCharts do
   def run(argv) do
     Mix.Task.run("compile")
 
-    {opts, _} = OptionParser.parse!(argv, strict: [only: :keep], aliases: [o: :only])
+    {opts, _} =
+      OptionParser.parse!(argv, strict: [only: :keep, format: :string], aliases: [o: :only])
 
     only =
       if opts[:only] && opts[:only] != [] do
         Enum.map(List.wrap(opts[:only]), &Path.expand/1)
       end
+
+    format = Keyword.get(opts, :format, "plain")
 
     flows()
     |> Task.async_stream(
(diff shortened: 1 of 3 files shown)


lib/mix/tasks/generate_flow_charts.ex
```diff
@@ -15,47 +15,60 @@ defmodule Mix.Tasks.Ash.GenerateFlowCharts do
...
   """
   use Mix.Task
 
   @shortdoc "Generates Mermaid Flow Charts for each `Ash.Flow`"
   def run(argv) do
     Mix.Task.run("compile")
 
-    {opts, _} = OptionParser.parse!(argv, strict: [only: :keep], aliases: [o: :only])
+    {opts, _} =
+      OptionParser.parse!(argv, strict: [only: :keep, format: :string], aliases: [o: :only])
```
zachdaniel, line 33: Should we add `-f` as an alias for `--format`?

lib/mix/tasks/generate_resource_diagrams.ex
```diff
@@ -31,6 +31,8 @@ defmodule Mix.Tasks.Ash.GenerateResourceDiagrams do
         Enum.map(List.wrap(opts[:only]), &Path.expand/1)
       end
 
+    format = Keyword.get(opts, :format, "plain")
```
zachdaniel, line 34: We aren't parsing the format in this case yet.

joshprice: Good catch. Fixed

lib/mix/tasks/generate_flow_charts.ex
joshprice: :+1: Seems consistent

joshprice:
Leaving this in draft for now until I get a chance to test it fully.

zachdaniel:
Yeah, it will be difficult to test the images we actually produce, so aside from testing "did it make the file", it will be pretty difficult I imagine 😆

lib/mix/tasks/generate_resource_diagrams.ex
```diff
@@ -52,7 +52,7 @@ defmodule Mix.Tasks.Ash.GenerateResourceDiagrams do
                 "mermaid-er-diagram",
                 format,
                 Ash.Api.Info.Diagram.mermaid_er_diagram(api),
-                "Generated ER Diagram for #{inspect(api)}"
+                "Generated ER Diagram for #{inspect(api)} (as #{format})"
```
zachdaniel, line 55: Maybe showing the file path & name here is useful as well? Potentially instead of the format, since the file extension will tell them?

joshprice: good point

joshprice:
Can you see any potential security risk with the File.write traversal as spotted by sobelow?

zachdaniel:
Not realistically, no. Its a security threat if you like...hook this code up to a web interface or something, but that wouldn't make much sense.

joshprice:
OK tested and good to go

zachdaniel (approved):

PR #430, opened 2022-10-26
docs: review guides and docs

joshprice:


documentation/tutorials/get-started.md
```diff
@@ -110,23 +110,26 @@ defp deps do
 end
 ```
 
-### Creating our first resources
+### Building your first Ash API
 
-The basic building blocks of an Ash application are resources. They are tied together by an API module (not to be confused with a web API), which will allow you to interact with those resources.
+The basic building blocks of an Ash application are Ash resources. They are tied together by an API module (not to be confused with a web API), which will allow you to interact with those resources.
+
+It might be helpful to think of an Ash API as a Bounded Context (in the Domain Driven Design sense), or as a Service (in the microservice sense).
```
zachdaniel, line 117: I like the bounded context callout, but equating an Ash Api to a microservice might give people the wrong idea.

documentation/tutorials/get-started.md
```diff
@@ -354,17 +357,19 @@ ticket
 
 ### Querying without persistence
 
-So far, there is no persistence happening. All that this simple resource does is return the record back to us. You can see this lack of persistence by attempting to use a `read` action:
+So far, we haven't stored any data in a database. All that this simple resource does is return the record back to us. You can see this lack of persistence by attempting to use a `read` action:
```
zachdaniel, line 360: We don't want to put across the idea that the only thing you might do with a resource is store it in a database. Might be worth making that distinction here. "Because we haven't used a data layer that does any persistence, for example storing records in a database."

documentation/tutorials/get-started.md
```diff
@@ -354,17 +357,19 @@ ticket
...
 
 ```elixir
 Helpdesk.Support.read!(Helpdesk.Support.Ticket)
 ```
 
 Which will raise an error explaining that there is no data to be read for that resource.
 
-In order to add persistence, we need to add a data layer to our resources. Before we do that, however, lets go over how Ash allows us to work against many different data layers (or even no data layer at all). Resources without a data layer will implicitly be using `Ash.DataLayer.Simple`, which will just return structs and do no persistence. The way that we do this is by leveraging `context`, a free-form map available on queries and changesets. The simple data layer looks for `query.context[:data_layer][:data][resource]`. It provides a utility, `Ash.DataLayer.Simple.set_data/2` to set it.
+In order to save our data somewhere, we need to add a data layer to our resources. Before we do that, however, lets go over how Ash allows us to work against many different data layers (or even no data layer at all).
```
zachdaniel, line 368: Data layers aren't necessarily *only* for saving data. They could be used as the back-end interactor, i.e if you wanted to send requests to an external system in response to create/update/destroy actions. I know that can be a bit of a heady concept for new people, but we want to make sure that people don't get the impression that Ash is for "CRUD persistence"

documentation/tutorials/get-started.md
```diff
@@ -385,28 +390,37 @@ tickets =
...
 Helpdesk.Support.Ticket
 |> Ash.Query.filter(status == :closed and not(contains(subject, "4")))
 |> Ash.DataLayer.Simple.set_data(tickets)
 |> Helpdesk.Support.read!()
 ```
 
-The examples shown here could be implemented easily using things like `Enum.filter`, but the real power here is to allow you to use the same tools when working with any data layer. If you were using AshPostgres, the above code would be exactly the same, except for the call to `set_data/2`.
+The examples above could be easily implemented with `Enum.filter`, but the real power here is to allow you to use the same tools when working with any data layer. If you were using the (link?) AshPostgres data layer, the above code would be exactly the same, except we wouldn't need the call to `set_data/2`.
+
+> TODO talk about any practical use for Simple Data layer, or explain why it is like it is (so you have something presumably until you set a real data layer)
```
zachdaniel, line 415: Many people use the simple data layer to model static data sets in a way that plays nicely with the rest of their app. i.e

```elixir
defmodule MyApp.Role do
 ....
  preparations do
    prepare MyApp.SetRoleData
  end   
end

defmodule MyApp.SetRoleData do
  use Ash.Resource.Preparation
  
    @data [
      %__MODULE__{name: :admin, permissions: [:create_thing, :create_other_thing]}
     ]
  
  def prepare(query, _, _) do
    Ash.DataLayer.Simple.set_data(query, @data)
  end
end
```

documentation/tutorials/get-started.md
```diff
@@ -385,28 +390,37 @@ tickets =
...
+
+`Ash.DataLayer.Simple` isn't actually very useful in practice so
```
zachdaniel, line 417: I'd disagree with this. Its plenty useful :)

documentation/tutorials/get-started.md
```diff
@@ -385,28 +390,37 @@ tickets =
...
 
 ### Adding basic persistence
 
-Before we get into working with relationships, lets add some actual persistence to our resource. This will let us add relationships and try out querying data.
+Before we get into working with relationships, lets add some real persistence to our resource. This will let us add relationships and try out querying data.
 
-There is a built in data layer that is good for testing and prototyping, that uses [ETS](https://elixir-lang.org/getting-started/mix-otp/ets.html).
+There is a built in data layer that is good for testing and prototyping, that uses [ETS](https://elixir-lang.org/getting-started/mix-otp/ets.html). ETS (Erlang Term Storage) is OTP's in-memory database, so the data won't actually stick around beyond the lifespan of your program, but it's simple, and behaves just like a real database.
```
zachdaniel, line 423: I'd hesitate to say "behaves just like a real database" as it has tons of limitations (no transactions, no timeouts, things like that)

documentation/tutorials/get-started.md
```diff
@@ -530,55 +546,76 @@ update :assign do
...
+
+> Long awkward sentence
+
+Where Ash shines is in all of the tools that can operate on your resources, the ability to extend the framework yourself, and the consistent design patterns that enable unparalleled efficiency, power and flexibility as your application and needs grow.
 
 #### Clean up your code that uses Ash?
 
 Creating and using changesets can be verbose. Check out the {{link:ash:guide:Code Interface}} to derive things like `Helpdesk.Support.Ticket.assign!(representative.id)`
 
+> Not sure what this bit is about, needs expanding
```
zachdaniel, line 609: Isn't the link to the other guide the expanding?

documentation/tutorials/get-started.md
```diff
@@ -530,55 +546,76 @@ update :assign do
...
+
 #### Persist your data
 
-See {{link:ash_postgres:guide:Get Started With Postgres:AshPostgres}} to see how to back your resources with postgres. This is highly recommended, as the postgres data layer provides tons of advanced capabilities.
+See {{link:ash_postgres:guide:Get Started With Postgres:AshPostgres}} to see how to back your resources with Postgres. This is highly recommended, as the Postgres data layer provides tons of advanced capabilities.
 
 #### Add an API
 
+> Add links, how should we link to extensions?
```
zachdaniel, line 617: `{{link:ash_postgres:extension:AshPostgres.DataLayer}}` would do it, for example.

zachdaniel: But the focus has to remain on `ash_hq`, I think. And if people aren't reading the docs there we should improve the experience until they do, IMO.

documentation/tutorials/get-started.md
joshprice: Need to describe how though at a high level, as per the comment above

documentation/tutorials/get-started.md
joshprice: Fair point

documentation/tutorials/get-started.md
joshprice: Agreed but the hex experience should be equivalent as that's what Elixir users are used to

documentation/tutorials/get-started.md
joshprice: Sort of, a short high level description would be useful here

documentation/tutorials/get-started.md
zachdaniel: The ship has sailed on that one, we already use the link expansion all over the place.

zachdaniel: So it's gotta be a separate project to add a ex doc pre processor

zachdaniel: Aside from link expansion, though, the experience should be equivalent, so all we need is that one pre processor

documentation/tutorials/get-started.md
joshprice: Agree. Persistence and saving data are basically interchangeable though

zachdaniel:

PR #432, opened 2022-10-27
feat(Ash.PlugHelpers): standardise conn interface for actors/tenants.

jimsynz:
This corresponds with a soon-to-arrive change in `ash_graphql` and `ash_json_api` which will make use of these functions to retrieve the actor and tenant from the `conn` rather than implementing it themselves.  This will make them seemlessly support AshAuthentication when it lands.

@zachdaniel I have added `plug` as an optional dependency, but it could also be just a dev/test dep.  Open to your feedback on that.

# Contributor checklist

- [X] Documentation changes


zachdaniel:
The optional dep vs dev/test only dep is an interesting question. I don't know how they would ultimately differ? I guess what it would do is it would require that they match said version of dep, if they are using that dep. 🤔 we probably don't care what version of plug they are using right? I feel like plug is extremely stable at least as far is this usage goes.

jimsynz:
Yup. I think changing it to a dev/test dep is wiser for that reason.

zachdaniel (approved):

jimsynz:
> This LGTM. Should we wait to merge until we have the PRs for AshGraphql and AshJsonApi teed up?

They're going to need an ash release to depend on.

PR #434, opened 2022-10-28
test: Add test demonstrating minimal example of strange upsert behaviour

oco-adam:
Add test demonstrating minimal example of strange upsert behaviour with lazy load children.

In particular, lazy loading the children of the parent resource in a global change block seems to affect the behaviour when upserting children through the parent.


zachdaniel:
🙇🏻 thanks for the reproduction!

PR #450, opened 2022-11-16
docs: Remove reference to `[:ash, :query]` telemetry

sevenseacat:
This doesn't exist in the codebase anywhere - all other telemetry events are referenced and called.

Also removed a duplicate `[:ash, :after_action]` list item.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Thanks for the report & fix!

It is a problem that we don't publish the query building telemetry event, something we would add here: https://github.com/ash-project/ash/blob/main/lib/ash/query/query.ex#L297

I can make this fix in a separate PR, or if you want to copy what we do here: https://github.com/ash-project/ash/blob/main/lib/ash/changeset/changeset.ex#L491 in this PR that would work too.

sevenseacat:
Fixing the code would probably make more sense! I'll take a look at it.

PR #464, opened 2022-12-19
fix: respect selects in related_query

barnabasJ:
When using ash_graphql it would be nice to only get the selects inside the authorizer which were actually selected. This was the change I had to make to get it to work.  

The related query also contains other fields, I saw `limit: 1` during my debugging, is this something that should also be passed here.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/load.ex
+++ b/lib/ash/actions/load.ex
@@ -1281,6 +1281,7 @@ defmodule Ash.Actions.Load do
     else
       relationship.destination
       |> Ash.Query.new(related_query.api)
+      |> Ash.Query.select(related_query.select)
       |> Ash.Query.filter(^related_query.filter)
     end
   end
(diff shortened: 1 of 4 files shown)


barnabasJ:
I was a bit fast with converting from draft to actual PR, I'm fixing the tests right now.

zachdaniel:
Thanks for catching and fixing this! I'm actually thinking we may want to just use `related_query` there in all cases, as opposed to selectively picking things off of the related_query.

barnabasJ:
> Would you mind trying out removing the calls to `load_query` and deleting that funciton, and just using `related_query` where it is called? 

Sure, I'll take a look at it tomorrow

zachdaniel:

PR #470, opened 2022-12-29
fix authorize? method in actions

kernel-io:
Fixes an issue where Ash was attempting to authorize requests where no actor / explicit call to authorize was set.

am thinking about how I can write a test for this


zachdaniel:
We have to support `actor?: nil` still running authorization, so we can't do this this way.

PR #476, opened 2023-01-11
fix: only pass action name in Ash.Changeset.for_action

barnabasJ:
Because `action` was reassigned, the complete action was passed along instead of just the name.

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -350,13 +350,13 @@ defmodule Ash.Changeset do
 
     case action.type do
       :create ->
-        for_create(initial, action, params, opts)
+        for_create(initial, action.name, params, opts)
 
       :update ->
-        for_update(initial, action, params, opts)
+        for_update(initial, action.name, params, opts)
 
       :destroy ->
-        for_destroy(initial, action, params, opts)
+        for_destroy(initial, action.name, params, opts)
 
       :read ->
         raise ArgumentError,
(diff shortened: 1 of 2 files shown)


zachdaniel (dismissed):

zachdaniel:
Actually...does this raise errors otherwise?

barnabasJ:
> Actually...does this raise errors otherwise?

Yes I always got a `ArgumentError: No such action .. on resource ...` Error

zachdaniel:
I think what we might want to do instead is modify the other calls so that they could accept an action struct, instead of the action name? Since the first thing those functions probably do is lookup the action. Thoughts?

barnabasJ:
> I think what we might want to do instead is modify the other calls so that they could accept an action struct, instead of the action name? Since the first thing those functions probably do is lookup the action. Thoughts?

That would mean updating the [do_for_action](https://github.com/ash-project/ash/blob/2f4d86b1b7d8145019d0f5e4a63ef0131ccb9076/lib/ash/changeset/changeset.ex#L408) as well as the [for_destroy](https://github.com/ash-project/ash/blob/2f4d86b1b7d8145019d0f5e4a63ef0131ccb9076/lib/ash/changeset/changeset.ex#L476) to handle `action / action_name` 

I initially went this way because the parameter was called `action_name` in `do_for_action` and `for_destroy` and I only ever used the `for_<action>` functions with atoms never with complete action structs. 

I think as it is easy to extend the function to work with both action structs as well as names it would be ok to do it, just not sure if anyone would ever really use it that way. 

LMK which way you'd prefer, and I'll make the necessary changes.

zachdaniel:
Yeah, its unlikely that people will use it that way, but the benefit is that we don't have to look up the action twice. I'd say lets make those two functions handle an action name or an action 👍🏻

zachdaniel (approved):

PR #500, opened 2023-02-01
docs: Render `api_short_name` visibly in the Monitoring guide

sevenseacat:
Because it currently looks like a HTML tag, it's not being rendered visibly - 

Screenshot 2023-02-01 at 10 14 34 am

This tweaks the short name to render in parens instead, and adds a note about what the short name actually is. (And the typespec was incorrect)

(I can't really figure out how a short name would actually be defined in an API module, so I guess it's just the default which is an atom!)


zachdaniel:
Yeah, looks like an oversight that you can't set it TBH.

zachdaniel:

PR #502, opened 2023-02-06
docs: Add Glossary guide 

sevenseacat:
An initial pass at setting up a glossary of terms and concepts within Ash, with brief definitions and links to other documentation.

As you can see I'm still fleshing out the definitions, but if there is any feedback on what's written already (or terms that should be added), let me know!

### DIFF
--- a/documentation/topics/glossary.md
+++ b/documentation/topics/glossary.md
@@ -1,0 +1,127 @@
+# Glossary
+
+## Action
+
+An action describes an operation that can be performed for a given resource; it is the _verb_ to a resource's _noun_. Examples of actions:
+
+- User._create_
+- Comment._delete_
+- BlogPost._publish_
+- Article._search_
+
+Ash supports four different types of actions - `create`, `read`, `update` and `destroy` (collectively often abbreviated as CRUD). A resource can define multiple actions per action type, eg. a _publish_ action would be considered an `update` because it is updating an existing instance of a resource. Actions are much more flexible than simple CRUD, but these four action types serve as templates for anything you might want to do.
+
+See the [Actions guide](/documentation/topics/actions.md) for more information.
+
+## Actor
+
+The entity that performs an action. Most actions are run on direct user request, eg. if a user presses a Create button on a page then the actor is the user; but an actor might also be an organization, a group, or the system itself.
+
+Actors are referenced during authorization - ensuring that the actor is allowed to perform an action, before it takes place. The actor can also be used within actions, to record which entity performed the action.
+
+See the [Security guide](/documentation/topics/security.md#actors) for more information.
+
+## Aggregate
+
+An aggregate is a special type of field for a resource, one that summarizes related information of the record. A more specialized type of a [calculation](#calculation).
+
+If a Project resource has_many Ticket resources, an example of an aggregate on the Project might be to count the tickets associated to each project.
+
+See the [Aggregates guide](/documentation/topics/aggregates.md) for more information.
+
+## API
+
+A method of broadly separating resources into different [bounded contexts](https://martinfowler.com/bliki/BoundedContext.html). Small apps might only have one API, in which case you can set-and-forget it, but apps with larger domains can benefit from different contexts having different views of the same resource.
+
+See `Ash.Api.Dsl` for more information.
+
+## Attribute
+
+A piece of data belonging to a resource. The most basic building block; an attribute has a type and a value and is stored within the context of a domain model.
+
+See `d:Ash.Resource.Dsl.attributes` for more information.
+
+## Calculation
+
+A calculation is a special type of field for a resource, one that is not directly stored in the data layer but generated on-demand when specifically requested as part of a query. Typically it will derive from other information on the record, but it may come from some other data source entirely.
+
+See the [Calculations guide](/documentation/topics/calculations.md) for more information.
+
+## Changeset
+
+Changesets encapsulate data changes made while creating or updating an instance of a resource. Similarly to [Ecto changesets](https://hexdocs.pm/ecto/Ecto.Changeset.html), they include data validations but they also have their own callback hook lifecycle.
+
+See `Ash.Changeset` for more information.
+
+## Extension
+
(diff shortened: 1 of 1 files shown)


documentation/topics/glossary.md
```diff
@@ -0,0 +1,95 @@
...
+
+Relationships (also known as _associations_) are named links between resources, that define how they relate to each other. Relationships can be used to signify ownership of a record, membership of a group, or can be used in filtering and querying data.
+
+See the [Relationships guide](/documentation/topics/relationships.md) for more information.
+
+## Registry
+
+## Resource
+
+The central concept in Ash, a resource is a domain model object in your system, the nouns that your app revolves around. Resources contain definitions of the data they hold in the form of attributes, but also define [actions](#action) that can be taken on that data and [actors](#actor) that are allowed to run them.
```
zachdaniel, line 89: I was going to say this should be at the top, but then I realized this was alphabetically sorted which makes sense.

I think it would be good to hint here that resources don't necessarily need to capture state in some way, i.e a database table. There could also be a resource like `ProductEmail` that you use the `send` action on.

documentation/topics/glossary.md
```diff
@@ -0,0 +1,95 @@
...
+
+## API
+
+## Attribute
+
+A piece of data belonging to a resource. The most basic building block; an attribute has a type and a value amd is stored within the context of a domain model.
+
+See `d:Ash.Resource.Dsl.attributes` for more information.
+
+## Calculation
```
zachdaniel, line 42: In Ash terms we'd call a `calculation` a "field", as opposed to a special attribute, generally speaking. There are four types of fields: attributes, calculations, aggregates, and relationships. I also might suggest flipping around the bit about aggregates, and putting it on the aggregate definition (which I see you've already put on there, so maybe just remove it from here? I'm on the fence so its your call).

documentation/topics/glossary.md
```diff
@@ -0,0 +1,95 @@
...
+
+Ash supports four different types of actions - `create`, `read`, `update` and `destroy` (collectively often abbreviated as CRUD). A resource can define multiple actions per action type, eg. a _publish_ action would be considered an `update` because it is updating an existing instance of a resource.
+
+See the [Actions guide](/documentation/topics/actions.md) for more information.
+
+## Actor
+
+The entity that performs an action. Most actions are run on direct user request, eg. if a user presses a Create button on a page then the actor is the user; but an actor might also be an organization, a group, or the system itself.
+
+Actors are referenced during authorization - ensuring that the actor is allowed to perform an action, before it takes place.
```
zachdaniel, line 22: The actor can also be used in the action to do things like relate the current user to the record as an "owner".

sevenseacat:
this is why I thought it would be a good idea to open a draft PR, to get this kind of feedback :D Will tweak what I've done so far and keep going!

documentation/topics/glossary.md
sevenseacat: That's an interesting point re: not needing state. I feel like if there's no state, then things like types of actions, identities, validations, etc. don't make sense (what are you validating? What are you CRUDing?) - which seems like a slippery slope on the way to saying "a resource can be just about anything at all". 🤔

zachdaniel: The action types are really just templates, they aren't meant to restrict you to working only with state that can be read back. The interface to a resource is the important part. A resource *can* be anything. Similarly, you can use an ecto schema to validate anything. For example, we're looking at CQRS/commanded integrations where a resource might have create actions but no read actions as it represents pushing to an event bus (it only tells you "success") but then other resources might have read actions but no create actions.

zachdaniel: So you'd be validating on arguments to actions, for example.

documentation/topics/glossary.md
```diff
@@ -0,0 +1,119 @@
...
+## Action
+
+An action describes an operation that can be performed for a given resource; it is the _verb_ to a resource's _noun_. Examples of actions:
+
+- User._create_
+- Comment._delete_
+- BlogPost._publish_
+- Article._search_
+
+Ash supports four different types of actions - `create`, `read`, `update` and `destroy` (collectively often abbreviated as CRUD). A resource can define multiple actions per action type, eg. a _publish_ action would be considered an `update` because it is updating an existing instance of a resource.
```
zachdaniel, line 14: I might add something like this: "Actions are much more flexible than simple CRUD, but these four action types serve as templates for anything you might want to do." ?

documentation/topics/glossary.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Calculation
+
+A calculation is a special type of field for a resource, one that is not directly stored in the data layer but generated on-demand when specifically requested as part of a query. Typically it will derive from other information on the record, but it may come from some other data source entirely.
+
+See the [Calculations guide](/documentation/topics/calculations.md) for more information.
+
+## Changeset
+
+Changesets encapsulate data changes made while creating or updating an instance of a resource. A wrapper around Ecto changesets, they include data validations and also have their own callback hook lifecycle.
```
zachdaniel, line 52: They aren't actually a wrapper around ecto changesets. AshPostgres creates an ecto changeset from an ash changeset before inserting but that is just an implementation detail.

documentation/topics/glossary.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Filter
+
+The tools and functions used to reduce the amount of data returned when running queries on the data layer. This may look like:
+
+- Fetching Articles that include a certain search term in the title
+- Fetching Posts created by a specific user
+- Fetching Tickets updated in the last week
+
+See `d:Ash.Filter` for more information.
```
zachdaniel, line 70:
```suggestion
See `Ash.Filter` for more information.
```

zachdaniel (approved):

sevenseacat:
if this has been approved, is there something stopping it from being merged?

zachdaniel:
only my memory :)

zachdaniel:

PR #508, opened 2023-02-09
fix: Type error: Suggest custom short type names as well

zimt28:


zachdaniel:

PR #509, opened 2023-02-09
docs: Update "Get Started" guide with an example

waseigo:
Added example of how to write the Representative ID directly into the Ticket during Ticket creation. Based on https://discord.com/channels/711271361523351632/1073248091324551228/1073249825363075102

# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/tutorials/get-started.md
+++ b/documentation/tutorials/get-started.md
@@ -608,6 +608,31 @@ ticket
 |> Helpdesk.Support.update!()
 ```
 
+### Assigning the Representative to a Ticket during its creation
+
+With the current definition of the Ticket resource the following will execute without error, but the `representative_id` field of the newly generated ticket will still remain empty:
+
+```elixir
+Helpdesk.Support.Ticket
+|> Ash.Changeset.for_create(:open, %{subject: "My spoon is too big!", representative_id: representative.id})
+|> Helpdesk.Support.create!()
+)
+```
+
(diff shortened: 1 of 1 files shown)


zachdaniel:
Hey there! I think perhaps this would fit a bit better in the relationships guide? I just worry about potentially bloating what is a pretty small guide. We could link to the relationships guide from here though?

waseigo:
The relationships guide already covers the `attribute_writable? true` option. From my point of view as an Ash beginner, I find it would be useful to mention this in the Getting Started guide -- after all, it's a very common thing a user would like to do, and it sensibly extends the Representative/Ticket example with a practical use case.

zachdaniel:
@sevenseacat thoughts on adding this tidbit to the current getting started guide?

sevenseacat:
I think it's a good idea! Do we want to encourage using `attribute_writable?` over `manage_relationship` though?

zachdaniel:
Honestly, I think attribute writable is simpler in general, and if you're using a data layer like postgres then it is a good idea because there is little value in the application side lookup pattern that we do in manage_relationship and ash_postgres will give you a good error. We should probably recommend that in the general case, and mention that "you can still manage a belongs_to relationship with `manage_relationship` but if you only want to relate the record to the other record, `attribute_writable?` is simpler and easier to reason about."

Issue #505, opened 2023-02-07
Add error message for page option in `read/2` with pagination not enabled 


minibikini:
**Is your feature request related to a problem? Please describe.**

When pagination is not explicitly enabled in a resource, calling `MyApp.MyAPI.read!(MyResource, page: [limit: 1])` returns all the records instead of just one. This can lead to confusion and unexpected behavior.

**Describe the solution you'd like**

To address this problem, I propose that when the `page` option is present, but pagination is not enabled in a resource, `read/2` should return an error with clear information about how to set up pagination in the resource.

kevinam99:
I could try working on this. Would appreciate any help with where to get started. I've been perusing the code base for some time but I haven't found that what needs to be updated. Thanks!

zachdaniel:
Hey there! pagination is enabled if `action.pagination` is not `nil` or `false`. A great place to check this would be right here: 
https://github.com/ash-project/ash/blob/main/lib/ash/api/api.ex#L1007

Where we have the options they've provided and the action itself. Something like `Keyword.has_key?(opts, :page) && !action.pagination`

PR #511, opened 2023-02-13
improvement: Add error message when page is added but pagination is not enabled

kevinam99:
Leaving the checklist here. Bit unsure how to test here. Aims to fix https://github.com/ash-project/ash/issues/505
# Contributor checklist.

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/api/api.ex
+++ b/lib/ash/api/api.ex
@@ -1003,7 +1003,8 @@ defmodule Ash.Api do
     query = Ash.Query.set_api(query, api)
 
     with {:ok, opts} <- Spark.OptionsHelpers.validate(opts, @read_opts_schema),
-         {:ok, action} <- get_action(query.resource, opts, :read, query.action) do
+         {:ok, action} <- get_action(query.resource, opts, :read, query.action),
+         {:ok, action} <- pagination_check(action, query.resource, opts) do
       Read.run(query, action, opts)
     else
       {:error, error} ->
@@ -1191,6 +1192,17 @@ defmodule Ash.Api do
               {:ok, action}
           end
         end
+    end
+  end
+
+  defp pagination_check(action, resource, opts) do
+    case Keyword.has_key?(opts, :page) && Map.get(action, :pagination) do
+      true ->
+        {:ok, action}
+
+      false ->
+        {:error,
+         "Pagination is not enabled on resource #{resource}. Check that you've set pagination to `true` in your action and also that you have set a page in your opts like this MyApp.MyAPI.read!(MyResource, page: [limit: 1])"}
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/ash/api/api.ex
```diff
@@ -1194,6 +1195,17 @@ defmodule Ash.Api do
     end
   end
 
+  defp pagination_check(action, resource, opts) do
+    case Keyword.has_key?(opts, :page) && Map.get(action, :pagination) do
+      true ->
+        {:ok, action}
+
+      false ->
+        {:error,
```
zachdaniel, line 1204: This is a great start! 

Two things:
1. If you don't mind, can you create a new `invalid` error class module for it? Then, you'd return `Ash.Error.to_error_class(YourNewModule.exception(resource: resource, action: action))`.

2. include the action in the message. Pagination is enabled on an action-by-action basis.

Great work so far!

kevinam99: Thanks for the feedback, Zach. I was working on exactly this when writing the test but I was stuck. I guess it was because I was adding it to the forbidden class? Will try with the invalid class and push here asap.

Does the error message look good?

zachdaniel: It looks good, but should be updated to something like this:
```elixir
Pagination is not enabled on the #{inspect action} action on #{inspect resource}.
```

zachdaniel: The first sentence, I mean.

lib/ash/error/invalid/page_requires_pagination.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  def_ash_error([:resource, :action], class: :invalid)
+
+  defimpl Ash.ErrorKind do
+    def id(_), do: Ash.UUID.generate()
+
+    def code(_), do: "page_requires_pagination"
+
+    def message(%{resource: resource, action: action}) do
+      "Pagination is not enabled on resource #{inspect(resource)} for the action #{inspect(action)}. Check that you've set pagination to `true` in your action and also that you have set a page in your opts like this MyApp.MyAPI.read!(MyResource, page: [limit: 1])"
```
zachdaniel:
```suggestion
      """
      Pagination is not enabled on resource #{inspect(resource)} for the action #{inspect(action)}. Check that you've
      enabled pagination in your action. For example:
      
      
      read #{action.name} do
        pagination offset?: true, keyset?: true, required?: false
      end
      """
```

zachdaniel: Making the error multiline will make it read a bit better and adding an example of how to fix it will be useful :D

kevinam99: Pushed this. Thank you for the suggestion!

kevinam99: > Making the error multiline will make it read a bit better and adding an example of how to fix it will be useful :D

Agreed!

zachdaniel (approved):

zachdaniel:

kevinam99:
Thank you for the support!

PR #522, opened 2023-03-09
docs: add drop-in configuration code snippet for the Spark formatter plugin

axelbdt:
I was frustrated by the existing code snippet for the configuration of the spark formatter.

Basically the ellipsis means I can't copy paste it to my own config, so I added a complete code snipped from the Ash-HQ config.

I kept the "If you use a different module than Ash.Resource" part in a separate subsection.


zachdaniel:
Nice! Thanks for the contribution!

PR #524, opened 2023-03-10
docs: rewrite Policies guide to be clearer

sevenseacat:
A rewrite of the Policies guide.

I've been bashing my head against policies for a few days now because I missed some of the fundamentals - the difference between simple and filter checks, which type is created when, and how `:unknown` values fit in. The info was kinda in the guide, but hidden away/really hard for me to wrap my head around - so I want to make it a lot clearer for future devs.

Please give it a good read and let me know if there are any technical issues - there might still be some holes in my understanding but I think I finally worked it all out!

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/topics/policies.md
+++ b/documentation/topics/policies.md
@@ -2,36 +2,34 @@
 
 Policies determine what actions on a resource are permitted for a given actor.
 
-## Important!
-
 Read and understand the [Security](/documentation/topics/security.md) guide before proceeding, which explains actors, how to set them, and other relevant configurations.
 
-## Guide
+## Setup
 
 You'll need to add the extension to your resource, like so:
 
 ```elixir
-use Ash.Resource,
-  authorizers: [Ash.Policy.Authorizer]
+use Ash.Resource, authorizers: [Ash.Policy.Authorizer]
 ```
 
-### Policy
+Then you can start defining policies for your resource.
 
-Every policy that applies must pass for a given request.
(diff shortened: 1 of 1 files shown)


documentation/topics/policies.md
```diff
@@ -72,19 +69,42 @@ end
...
+authorize_if RegularUserAuthorized
+```
+
+### Not all policy checks have yes/no answers
+
+This will be covered in greater detail in [Checks](#checks), but will be briefly mentioned here.
+
+Ash provides two basic types of policy checks - _simple_ checks and _filter_ checks. Simple checks are what we commonly think of with authorization, and what the above example would suggest - is an actor allowed to perform a given operation, yes or no? But we can also use filter checks - given a list of resources, which ones is an actor allowed to perform the operation on?
+
+Filter checks are frequently used with read actions, as they can refer to multiple instances (eg. "list all products"), but may also be applied to actions like bulk-deleting records.
```
zachdaniel, line 99:
```suggestion
Filter checks are frequently used with read actions, as they can refer to multiple instances (eg. "list all products"), but may also be applied to actions like bulk-deleting records (which is not currently supported, but will be eventually).
```

zachdaniel (approved):

sevenseacat:
You have no idea how relieved I am to finally "get it" and have that validated by this, lol

PR #526, opened 2023-03-13
improvement: Add `before_action/1` and `after_action/1` to `Ash.Resource.Preparations.Builtins`.

jimsynz:
This allows folks to add lifecycle hooks right in their read actions, eg:

```elixir
prepare before_action(&IO.inspect(&1))
```

rather than:

```elixir
prepare fn query, _context -> Ash.Query.before_action(query, &IO.inspect/1) end
```

* Add (better) typespecs for `Ash.Query`.
* Remove redundant guards as detected by dialyzer.
* Add new `phase` field to `Ash.Query` which can be used to introspec whether a query is currently running lifecycle hooks.

### DIFF
--- a/lib/ash/actions/read.ex
+++ b/lib/ash/actions/read.ex
@@ -1054,7 +1054,7 @@ defmodule Ash.Actions.Read do
                  {:ok, query} <- set_tenant(query, ash_query),
                  {:ok, results} <-
                    run_query(
-                     ash_query,
+                     set_phase(ash_query, :executing),
                      query,
                      %{
                        actor: actor,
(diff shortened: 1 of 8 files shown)


lib/ash/actions/read.ex
```diff
@@ -1303,39 +1303,44 @@ defmodule Ash.Actions.Read do
...
   defp run_after_action(query, results) do
+    query = set_phase(query, :after_action)
+
     query.after_action
     |> Enum.reduce_while({query, {:ok, results, []}}, fn after_action,
                                                          {query, {:ok, results, notifications}} ->
       case after_action.(query, results) do
         {:ok, results} ->
-          {:cont, {query, {:ok, results, notifications}}}
+          {:cont, {set_phase(query), {:ok, results, notifications}}}
```
zachdaniel, line 1337: I don't think we want to reset the phase here, since its still in the `after_action`.

zachdaniel (approved):
nvm ignore me

zachdaniel (approved):

PR #528, opened 2023-03-17
improvement: add option `set_when_nil?` to builtin change `set_attribute`

bcksl:
Adds [requested](https://discord.com/channels/711271361523351632/712035333432147989/1085967846632345682) option `set_when_nil?` to builtin change `set_attribute`, indicating that the change should only set the attribute when its current value is not `nil`.

Includes update of dependency `picosat_elixir` to `~> v0.2.3`.

*Note: Existing test suite passes, however this PR does not yet include additional docs/tests.*

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/change/builtins.ex
+++ b/lib/ash/resource/change/builtins.ex
@@ -18,6 +18,25 @@ defmodule Ash.Resource.Change.Builtins do
     ]
   ]
 
+  @set_attribute_opts [
+    attribute: [
+      doc: "The attribute to change.",
+      required: true,
+      type: :atom
+    ],
+    value: [
+      doc:
+        "The value to set the attribute to; may be a fn/0 which will be called to produce the value.",
+      required: true,
+      type: {:custom, Ash.Resource.Change.SetAttribute, :validate_value, []}
+    ],
+    set_when_nil?: [
+      doc: "When false, decline setting the attribute if it is nil.",
+      type: :boolean,
+      default: true
+    ]
+  ]
+
   @doc """
   Relates the actor to the data being changed, as the provided relationship.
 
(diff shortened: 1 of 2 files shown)


lib/ash/resource/change/builtins.ex
```diff
@@ -50,16 +58,29 @@ defmodule Ash.Resource.Change.Builtins do
...
+  @spec set_attribute(relationship :: atom, (() -> term) | {:_arg, :status} | term(), opts :: Keyword.t()) ::
           Ash.Resource.Change.ref()
-  def set_attribute(attribute, value) do
-    {Ash.Resource.Change.SetAttribute, attribute: attribute, value: value}
+  def set_attribute(attribute, value, opts \\ []) do
+    opts =
+      opts
+      |> Keyword.put_new(:set_when_nil?, true)
+
+    {Ash.Resource.Change.SetAttribute, attribute: attribute, value: value, extra_opts: opts}
```
zachdaniel, line 83: I think it would probably be simpler to do`set_when_nil?: Keyword.get(opts, :set_when_nil?, true)` instead of using an `extra_opts` list.

mix.exs
```diff
@@ -276,7 +276,7 @@ defmodule Ash.MixProject do
       {:ecto, "~> 3.7"},
       {:ets, "~> 0.8.0"},
       {:decimal, "~> 2.0"},
-      {:picosat_elixir, "~> 0.2"},
+      {:picosat_elixir, "~> 0.2.3"},
```
zachdaniel, line 279: This probably isn't necessary for this PR, right?

lib/ash/resource/change/builtins.ex
zachdaniel: I like that plan 👍

mix.exs
zachdaniel: I think we should stay without the patch dependency. Users will get the latest fixes by updating the dep, individual patches don't need to update their dependency on other packages when they add fixes.

lib/ash/resource/change/builtins.ex
```diff
@@ -50,16 +68,30 @@ defmodule Ash.Resource.Change.Builtins do
 
     Use `arg(:argument_name)` to use the value of the given argument. If the argument is not supplied then nothing happens.
 
+  ## Options
+
+  #{Spark.OptionsHelpers.docs(@set_attribute_opts)}
```
zachdaniel, line 73:
```suggestion
  #{Spark.OptionsHelpers.docs(Keyword.drop(@set_attribute_opts, [:attribute, :value]))}
```

Since they are passed as arguments we can hide them from docs.

mix.exs
zachdaniel: huh. That is pretty interesting...I wonder why that is?

zachdaniel (approved):

PR #530, opened 2023-03-21
improvement: Add `get_by` option to read actions.

jimsynz:


lib/ash/resource/transformers/get_by_read_actions.ex
```diff
@@ -0,0 +1,241 @@
...
+              [:actions, :read, action.name, :get_by],
+              "`#{inspect(get_by)}` is not a valid attribute, calculation or aggregate"
+            )}}
+      end
+    end)
+  end
+
+  defp validate_existing_arguments(_dsl_state, action) when action.arguments == [], do: :ok
+
+  defp validate_existing_arguments(dsl_state, action) do
```
zachdaniel, line 133: So, its not *necessarily* the case that the types have to match...this kind of thing may be too restrictive. I guess it can't really hurt to start here though.

zachdaniel (approved):

PR #556, opened 2023-04-05
improvement: add constraints to maps

### LINKED ISSUE #555: Constraints for maps
Maps are nice to get going, but it would be nice to be able to specify constraints on them, like what key/value pairs should be possible to pass or make some key/value pairs required.

I'd like to add constraints to the map type like this:

```elixir
actions do
  create :new do
    argument :params, :map do
       constraints fields: [{:field1, :type, [:type_constrains]}, {:field2, :type, [:type_constrains]}]
    end 
  end
end
```

I tried to use embedded resources as types, but because all keys are always present in the struct, it's impossible to detect if a value was not set or if it was set to nil as the resulting struct will always have nil value

The constraints could be used to create a corresponding Graphql Type in AshGraphql

Discord Thread: https://discord.com/channels/711271361523351632/1092361830166450216/1092361830166450216

barnabasJ:
# Contributor checklist

Add constraints to maps, should close #555

- [x] Documentation changes


lib/ash/type/map.ex
```diff
@@ -16,26 +54,69 @@ defmodule Ash.Type.Map do
...
-  def cast_input(value, _) when is_map(value), do: {:ok, value}
+  def cast_input(value, constraints) when is_map(value), do: check_constraints(value, constraints)
   def cast_input(_, _), do: :error
 
   @impl true
   def cast_stored(nil, _), do: {:ok, nil}
-  def cast_stored(value, _) when is_map(value), do: {:ok, value}
+
+  def cast_stored(value, constraints) when is_map(value),
+    do: check_constraints(value, constraints)
```
barnabasJ, line 74: I don't know if this is needed here. Theoretically, a stored value should already be validated, but if other services use the same DB, for example, it could be that it's not correct.

zachdaniel: Yeah, we should leave it out of here because not all types can necessarily expect to be `cast_input` from the result of `cast_stored` (like it might be a one way translation). We could potentially add that as a constraint/option later down the road if people need it.

lib/ash/type/map.ex
```diff
@@ -1,13 +1,51 @@
 defmodule Ash.Type.Map do
+  @constraints [
+    fields: [
+      type: {:custom, __MODULE__, :field_types, []},
```
zachdaniel, line 4:
```elixir
  [
    fields: [
      type: :keyword_list,
      keys: [
        *: [
          type: :keyword_list,
          schema: [
            type: [
              type: Ash.OptionsHelpers.ash_type(),
              required: true
            ],
            constraints: [
              type: :keyword,
              default: []
            ]
          ]
        ]
      ]
    ]
  ]
```
Should be able to do something like this

lib/ash/type/map.ex
```diff
@@ -16,26 +54,69 @@ defmodule Ash.Type.Map do
 
   def cast_input(nil, _), do: {:ok, nil}
 
-  def cast_input(value, _) when is_binary(value) do
+  def cast_input(value, constraints) when is_binary(value) do
     case Jason.decode(value) do
       {:ok, value} ->
-        {:ok, value}
+        check_constraints(value, constraints)
```
zachdaniel, line 60:
```suggestion
        cast_input(value, constraints)
```

zachdaniel:
This is really great!

lib/ash/type/map.ex
barnabasJ: I tried this, but I'm a bit stuck because it does not compile with this error:

```
== Compilation error in file lib/ash/type/map.ex ==
** (FunctionClauseError) no function clause matching in NimbleOptions.Docs.get_raw_type_str/1    
    
    The following arguments were given to NimbleOptions.Docs.get_raw_type_str/1:
    
        # 1
        {:spark_type, Ash.Type, :builtins, [{"{:array, inner_type}", "list", "{:array, ${1:inner_type}}"}]}
    
    Attempted function clauses (showing 10 out of 27):
    
        defp get_raw_type_str(nil)
        defp get_raw_type_str({:custom, _mod, _fun, _args})
        defp get_raw_type_str(:mfa)
        defp get_raw_type_str(:mod_arg)
        defp get_raw_type_str({:or, _values})
        defp get_raw_type_str({:in, _})
        defp get_raw_type_str({:fun, arity})
        defp get_raw_type_str(:any)
        defp get_raw_type_str(:reference)
        defp get_raw_type_str(:pid)
        ...
        (17 clauses not shown)
    
    (nimble_options 1.0.1) lib/nimble_options/docs.ex:113: NimbleOptions.Docs.get_raw_type_str/1
    (nimble_options 1.0.1) lib/nimble_options/docs.ex:104: NimbleOptions.Docs.get_type_str/1
    (nimble_options 1.0.1) lib/nimble_options/docs.ex:43: NimbleOptions.Docs.option_doc/2
    (elixir 1.14.3) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
    (nimble_options 1.0.1) lib/nimble_options/docs.ex:63: NimbleOptions.Docs.option_doc/2
    (elixir 1.14.3) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
    (nimble_options 1.0.1) lib/nimble_options/docs.ex:6: NimbleOptions.Docs.generate/2
    lib/ash/type/map.ex:57: (module)
```

barnabasJ: This is the result of the Ash.OptionsHelpers.ash_type() call

lib/ash/type/map.ex
```diff
@@ -16,26 +74,77 @@ defmodule Ash.Type.Map do
...
+
+  defp check_fields(value, fields) do
+    Enum.reduce(fields, {:ok, %{}}, fn
+      {field, field_constraints}, {:ok, checked_value} ->
+        case fetch_field(value, field) do
+          nil ->
+            if field_constraints[:allow_nil?] == false do
+              {:error, [[message: "must not be nil", field: field]]}
+            else
+              {:ok, Map.put(checked_value, field, nil)}
```
barnabasJ, line 120: Not completely sure about this, but the change overall would lead to all the keys in the constraints always being present (set to nil) and extra keys being removed. 

I think I prefer it this way.

barnabasJ:
Thanks for the Feedback, I might need a bit more help with the constraint "schema"

barnabasJ:
credo and tests passed on my machine, not sure why they aren't working in ci

lib/ash/type/map.ex
zachdaniel: If you change `fetch_field` to return `:error` when neither key is present then you can detect the two cases. And not add keys to the map that were missing in the input.

zachdaniel:
Looks like I broke credo :)

lib/ash/type/map.ex
barnabasJ: Do you know how Absinthe will react if we create the graphql type from the constraints and the map that gets passed does not contain all the keys of the type?

lib/ash/type/map.ex
zachdaniel: CI should release spark: v1.0.6 shortly.

zachdaniel:
Main has been fixed. The `mix test (false, false)` can't be helped, that will fail for now, but the other checks should be passing if you pull main.

barnabasJ:
I think it should be ok now. While working on the changes, I also saw that I shouldn't just apply the constraints at cast time but that it is a behaviour callback.

lib/ash/type/map.ex
```diff
@@ -31,11 +90,71 @@ defmodule Ash.Type.Map do
...
+        case fetch_field(value, field) do
+          {:ok, nil} ->
+            if field_constraints[:allow_nil?] == false do
+              {:error, [[message: "value must not be nil", field: field]]}
+            else
+              {:ok, Map.put(checked_value, field, nil)}
+            end
+
+          {:ok, field_value} ->
+            case Ash.Type.apply_constraints(
```
zachdaniel, line 126: First we want to cast it.

```elixir
with {:ok, value} <- cast(...),
       {:ok, value} <- apply_constraints() do
end
```

Lastly, we need to move the `allow_nil?` check from above into this branch. A type might choose to cast `nil` -> `""` for example, or like `:string`, it might do `"" -> nil` meaning we need to do nil checks after casting.

zachdaniel:
Alright, one more comment and then I think its good to go 👍

zachdaniel (approved):

zachdaniel:
This is great work, and a great idea!

PR #565, opened 2023-04-16
Fix: Add tenant attribute to identity attributes on upserts

marmor157:
Currently when running upserts with `upsert_identity` option set to some identity, on resource with attribute multitenancy, it sets the `on_conflict` fields to only identity fields and ignores the multitenancy attribute. This PR aims to fix that by adding this attribute if needed.


zachdaniel (approved):

zachdaniel:
🚀Thank you for your contribution!🚀

PR #587, opened 2023-05-26
Add `Negate` validation

marmor157:
This PR adds `negate` validation that negates other validations. This allows to do things like `negate(one_of(:attr, [:val1, :val2])`.

Originally I wanted to name it `not` but when adding a builtin I've found that it might conflict with `Kernel.not` so I've renamed it to `negate`

### DIFF
--- a/lib/ash/resource/validation/negate.ex
+++ b/lib/ash/resource/validation/negate.ex
@@ -18,14 +18,8 @@ defmodule Ash.Resource.Validation.Negate do
     with {:ok, opts} <- Spark.OptionsHelpers.validate(opts, @opt_schema),
          {validation, validation_opts} = opts[:validation],
          {:ok, validation_opts} <- validation.init(validation_opts) do
-       Keyword.put(opts, :validation, {validation, validation_opts})
+      opts = Keyword.put(opts, :validation, {validation, validation_opts})
       {:ok, opts}
-    else
-      {:error, exception} = error when is_binary(exception) ->
-        error
-
-      {:error, error} ->
-        {:error, Exception.message(error)}
     end
   end
 
--- a/lib/ash/resource/validation/negate.ex
+++ b/lib/ash/resource/validation/negate.ex
@@ -1,0 +1,47 @@
+defmodule Ash.Resource.Validation.Negate do
+  @moduledoc false
+
+  alias Ash.Error.Changes.InvalidAttribute
+
+  @opt_schema [
+    validation: [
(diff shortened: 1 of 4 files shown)


lib/ash/resource/validation/negate.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  @impl true
+  def init(opts) do
+    with {:ok, opts} <- Spark.OptionsHelpers.validate(opts, @opt_schema),
+         {validation, validation_opts} = opts[:validation],
+         {:ok, _opts} <- validation.init(validation_opts) do
+      {:ok, opts}
+    else
+      {:error, exception} = error when is_binary(exception) ->
+        error
```
zachdaniel, line 24: It ought to be safe to return whatever error that nested init returns. (i.e removing the `else` case here should be okay).

lib/ash/resource/validation/negate.ex
```diff
@@ -0,0 +1,47 @@
...
+    ]
+  ]
+
+  use Ash.Resource.Validation
+
+  @impl true
+  def init(opts) do
+    with {:ok, opts} <- Spark.OptionsHelpers.validate(opts, @opt_schema),
+         {validation, validation_opts} = opts[:validation],
+         {:ok, _opts} <- validation.init(validation_opts) do
```
zachdaniel, line 20:
```suggestion
         {:ok, validation_opts} <- validation.init(validation_opts) do
       Keyword.put(opts, :validation, {validation, validation_opts})
```

lib/ash/resource/validation/negate.ex
```diff
@@ -0,0 +1,47 @@
...
+    {validation, validation_opts} = opts[:validation]
+
+    case validation.validate(changeset, validation_opts) do
+      {:error, _} ->
+        :ok
+
+      :ok ->
+        {:error,
+         InvalidAttribute.exception(
+           message: "must not pass validation %{validation}",
```
zachdaniel, line 37: This part is a bit interesting...we can't put the validation and opts module into the message like this for security reasons. In the short term, this probably looks like adding a new, optional callback to validations called `describe/1` that takes its opts and returns a description. Then, in here, we'd see if the child validation implements that callback, and if so we'd use `not: <its_description>` or something like that.

zachdaniel:
This is awesome! A bit more might be required to make it work, but I'm really excited about it :D

zachdaniel:
Closing for now as there are changes that need to be made before this could be merged. Feel free to reopen if progress continues.

PR #594, opened 2023-05-31
fix: remove field name from string length error message

vonagam:
Other errors do not contain the field in the message. Removed it from this one to be consistent with them and more flexible (a user can still add the field to the message by themselves as it is present on the error).

About consistency of messages - seems like there is a split when it comes to capitalization of a first letter. Close to 50-50. Is there a rule of thumb when the message should be capitalized and when it should not?

Shouldn't the code for `Error.Forbidden.Policy` be "policy" and not ["forbidden"](https://github.com/ash-project/ash/blob/2384d35eeadeb4cc632711776a2f31597f548b1e/lib/ash/error/forbidden/policy.ex#L410), just like in other specific forbidden errors?

There is `Error.Changes.UnknownError` and `Error.Unknown.UnknownError`, later is used in the codebase while former seems to be not, is first one still exists just for backwards compatibility of sorts?

### DIFF
--- a/lib/ash/error/forbidden/policy.ex
+++ b/lib/ash/error/forbidden/policy.ex
@@ -407,6 +407,6 @@ defmodule Ash.Error.Forbidden.Policy do
       end
     end
 
-    def code(_), do: "forbidden"
+    def code(_), do: "forbidden_by_policy"
   end
 end


zachdaniel:
Great changes :)

For forbidden errors, they should realistically all share the same code. That code is used to potentially present to external interfaces and we don’t want to leak reasons for failed authorization.

the unused error you mentioned should be removed, good catch. And yeah we should probably keep first letter lowercased.

If you’d like to do any additional changes like removing the unused error in this same PR, feel free. Otherwise lmk and I will merge :)

zachdaniel (dismissed):

vonagam:
Ok, added removal of `Error.Changes.UnknownError` as a separate commit into this PR, can squash if this is preferable, but resulting commit message will become less descriptive.

About `code()` for policy, seems like all places that work with errors (phoenix forms, graphql, json api) implement their own protocol for exposing errors and in those implementations forbidden errors are handled specifically not to expose real codes, so I assume it is ok to have forbidden errors to have specific codes.

vonagam:
Added lowercasing of ash error messages. Some tests in other ash repositories might need an update - for example, I see `assert message =~ "Confirmation did not match value"` in graphql tests.

zachdaniel:
> Ok, added removal of `Error.Changes.UnknownError` as a separate commit into this PR, can squash if this is preferable, but resulting commit message will become less descriptive.
> 
> About `code()` for policy, seems like all places that work with errors (phoenix forms, graphql, json api) implement their own protocol for exposing errors and in those implementations forbidden errors are handled specifically not to expose real codes, so I assume it is ok to have forbidden errors to have specific codes.

Yeah, you’re right. Go for it :) but make it `policy_forbidden` or something like that I think. Or even `forbidden_by_policy`

zachdaniel:
ready to go merge? LGTM :D

vonagam:
Yeah, nothing holding that I know of.

zachdaniel (approved):

zachdaniel:

PR #595, opened 2023-06-01
Only create values for attributes that did not get a generator

barnabasJ:
This would eliminate this error ` ** (RuntimeError) generator/1 unimplemented for Ash.Type.Map` if the Resource has an attribute of type map that is not nullable and a generator is passed for that attribute.


zachdaniel (dismissed):

zachdaniel:
Looks like a credo issue to fix but otherwise looks good to me

PR #596, opened 2023-06-02
fix: remove `class` and `stacktrace` from `ErrorKind` implementations

vonagam:
Still looking into errors and noticed that some implementations of `ErrorKind` have `class` and `stacktrace` while `ErrorKind` protocol does not have those methods. I assume that those are outdated placeholder things and can be removed, right?


zachdaniel (approved):

zachdaniel:
Really appreciate the help in making things more consistent. Awesome work!

PR #597, opened 2023-06-02
fix: rename few error files to match module names

vonagam:
Just when last PR was merged I noticed some wrong file names for query errors (nothing mismatching for other errors):
```
calculations_not_supported copy.ex → calculations_not_supported.ex
offset.ex → invalid_offset.ex
unsortable_field.ex → unsortable_attribute.ex
```

If you want you can keep this PR open for some time if you do prefer to keep such small changes in a single PR and then squash it into one commit and so if I find something else I can add it here. (Just saying because before you did prefer that.)


vonagam:
Okay, do see another thing that a little bit inconsistent. Many `NoSuch` errors use field `name` to reference the thing that they are about. But this is not how other errors do that, like `Query.InvalidArgument` or `Query.InvalidQuery` use `field` but `Query.NoSuchAttribute` uses `name`. And even not all `NoSuch` errors use `name` - `Invalid.NoSuchAction` `action`, `Invalid.NoSuchResource` has `resource` and `Load.NoSuchRelationship` has `relationship`.

I think it is reasonable to rename these fields to align with other errors: 

```
Changes.NoSuchAttribute name -> field
Changes.NoSuchRelationship name -> relationship
Query.NoSuchAttributeOrRelationship attribute_or_relationship -> field_or_relationship
Query.NoSuchAttribute name -> field
Query.NoSuchFunction name -> function
Query.NoSuchOperator name -> operator
Query.Query.NoSuchRelationship name -> relationship
```

Does it makes sense?

zachdaniel:
I think to start we would need to just add the new fields you’re talking about and maintain both, because this could be a breaking change to error handling. Then on 3.0 we could remove the old fields. If we add those we can’t forget to make an issue to track removing the old ones for 3.0

vonagam:
Hm, maintaining both simultaneously can be troublesome. I guess the change can be done specifically for 3.0 when it happens. It is not crucial anyway, just a naming thing.

zachdaniel:
Sounds good, could you create an issue for it? Much appreciated :)

vonagam:
Done. (Though I think labels on the issue need changing.)

zachdaniel:
Awesome, thanks!

PR #608, opened 2023-06-09
fix: drop proper opts keys when calling api in code interface

vonagam:
Looking at the code of code_interface saw some inconsistencies when it comes to `Keyword.drop` calls in api calls.

`action` was dropping `:changeset` when it should have dropped `:input`.
`action!` did the same plus `tenant` was missing in drops.
`read` and `read!` were missing `tracer` drop.
`create!` was missing `tenant` drop.

Reordered drop keys to be consistently - `[...specific_keys, :actor, :tenant, :authorize?, :tracer]`. 

And in two places replaced simple `||` with `> Kernel.||`, just because majority of relevant places used later form.

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -449,7 +449,7 @@ defmodule Ash.CodeInterface do
 
                 unquote(api).run_action(
                   action_input,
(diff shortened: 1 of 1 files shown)


zachdaniel (approved):
beautiful, as always!

PR #609, opened 2023-06-09
fix: refactor out some repetition in code interface

vonagam:
Part addition of a feature and part refactor/fix.

Main refactor part: currently in code interface file there is a lot of repetition. So I moved repeated parts of each action into quoted expressions and reuse those.

Also replaced such construction:
```elixir
if opts == [] && Keyword.keyword?(params_or_opts) do
  apply(__MODULE__, elem(__ENV__.function, 0), [
    unquote_splicing(arg_vars),
    %{},
    params_or_opts
  ])
else
```
With:
```elixir
{params, opts} =
  if opts == [] && Keyword.keyword?(params_or_opts),
    do: {%{}, params_or_opts},
    else: {params_or_opts, opts}
```

And renamed `input` variables into `params`, since this is what arguments they are passed into are called.

Fix comes from the fact that repetition of the code leads to incomplete fixes, changes and inconsistencies over time. Just right now there was a fix commit f3703f9ebebf72fa7c41686075a6e0883a706e35 that touched the code of generated `read` func, but not `read!`.

Addition comes in the form of new automatically generated functions `input_for_$action`, `query_for_$read` and `changeset_for_$create/$update/$destroy`. Desire for those came when I was implementing a simple rest controller that deals with json. Code interface is suggested as preferable way to work with actions in code and it covers most resource generated routes: `index` and `show` -> `read!`, `create` -> `create!`, `update` -> `update!`, `delete` -> `destroy!`. But `new` and `edit` were hanging. In html/live view there are helpers that create form objects and those fill that need in that context, but when it comes to a generic solution there was none other than to go and write code that documentation calls verbose.

So I added those additional methods. Since they do share almost all code with their respective counter-parts and that shared code was moved into a variable in total this PR reduces number of lines in the file, not increases it.

I guess, I'll need to add at least few tests/assertions for those new generated methods, but before that want to confirm that the addition can be accepted and if there are any alternative naming suggestions.

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -417,6 +417,30 @@ defmodule Ash.CodeInterface do
 
         case action.type do
           :action ->
+            resolve_opts_params_input =
+              quote do
+                {params, opts} =
+                  if opts == [] && Keyword.keyword?(params_or_opts),
+                    do: {%{}, params_or_opts},
+                    else: {params_or_opts, opts}
+
+                params =
+                  unquote(args)
+                  |> Enum.zip([unquote_splicing(arg_vars)])
+                  |> Enum.reduce(params, fn {key, value}, params ->
+                    Map.put(params, key, value)
+                  end)
+
+                input =
+                  opts[:input]
+                  |> Kernel.||(unquote(resource))
+                  |> Ash.ActionInput.for_action(
+                    unquote(action.name),
+                    params,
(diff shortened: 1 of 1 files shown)


vonagam:
Removed those additional functions, so it is a simply refactor/fix now. Adding those additional methods can be done easily in a separate PR, but now this one is not holded back from being merged by potential discussion on naming or validity of feature reasoning.

zachdaniel (approved):

zachdaniel:

PR #611, opened 2023-06-11
fix: continue refactor in code interface

vonagam:
Continuation from #609.

Move common code that swaps opts/params and merges named args into params as `resolve_opts_params`.

Instead of doing `Keyword.take` and `Keyword.drop` to separate between subject (input/query/changeset) options and api options, use `Keyword.split`.

Good changes on their own, but now it also will be quite easy for actions `case` to return tuple `{subject_name, resolve_subject, subject}` and then define three additional functions outside of case (so no code repetition) for `subject_to_act`, `can_act` and `can_act?`. So likes of `query_to_find_by_slug`/`changeset_to_archive`/`can_publish?` can be defined automatically with code interface and the code to add such functionality is quite small - just return the subject or pass it into `api.can`/`api.can?`. Are there any reservations about adding those methods or is it acceptable addition (in separate PR)?

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -415,30 +415,36 @@ defmodule Ash.CodeInterface do
         #{Spark.OptionsHelpers.docs(Ash.Resource.Interface.interface_options(action.type))}
         """
 
+        resolve_opts_params =
+          quote do
+            {params, opts} =
+              if opts == [] && Keyword.keyword?(params_or_opts),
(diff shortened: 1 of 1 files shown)


zachdaniel (approved):

zachdaniel:
Those functions sound really useful :). Id like to take some time to consider their names, but the ones you’ve proposed look pretty good. Keep up the amazing work!

PR #614, opened 2023-06-12
improvement: generate some action helpers with code interface

vonagam:
Adds three new auto generated helper methods for actions in code interface.

`$input_to_$act` - `input_to_execute`, `query_to_find_by_id`, `changeset_to_create` and so on. Creates and returns corresponding input based on action type. Generic actions produce `Ash.ActionInput`, reads produce `Ash.Query` and create/update/destroy - `Ash.Changeset`.

`can_$act`/`can_$act?` - passes produced input right into `api.can`/`api.can?`. (Is not generated for generic actions as `api.can` works only with queries/changesets at the moment.) First required argument is an actor after that the same arguments as in rest of generated methods.

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -430,250 +430,217 @@ defmodule Ash.CodeInterface do
               end)
           end
 
-        case action.type do
-          :action ->
-            resolve_opts_params_input =
-              quote do
-                unquote(resolve_opts_params)
+        {subject, subject_args, resolve_subject, act, act!} =
+          case action.type do
+            :action ->
+              subject = quote do: input
 
-                {input_opts, opts} =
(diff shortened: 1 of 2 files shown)


zachdaniel:
🔥 🔥 🔥 🔥 🔥

zachdaniel (approved):

zachdaniel:
This is great! If you're interested you could make an announcement in the #showcase channel in the discord if you'd like? I'm sure people would like to know about this :)

vonagam:
Thanks. After looking into discord, seems like #announcements would be better fit (as the channel regularly contains updates on released features or info on in-production ones) than #showcase (seems like it is about video/talks/projects).

I am still really new to ash (so need to battle test the feature before I am confident that everything works as it should, not confident enough in current test coverage) and not a press-release guy in general so will pass on this opportunity, but thanks for suggestion and the project itself.

PR #617, opened 2023-06-13
improvement: add support for generic actions in `api.can`

vonagam:
Now `api.can`/`api.can?` can accept `ActionInput` or a name of a generic action.

Now code interface generates `can_$act`/`can_$act?` helpers for generic actions.

In code interface made sure that the subject (query/input/changeset) get `api` property set to current api, before it was done only for query (with recent change), but based on types and code in `api.can` I assume it should be applicable to changesets/inputs too.


zachdaniel (approved):

PR #624, opened 2023-06-14
docs: `:maybe` means true in `can?` by default, not false

vonagam:


zachdaniel (approved):

zachdaniel:

PR #626, opened 2023-06-16
adds verifier for identity fields.

### LINKED ISSUE #576: Validation: Ash.Resource.Verifiers.VerifyIdentityFields

Given this schema:
```elixir
  attributes do
    uuid_primary_key :id

    attribute :title, :string do
      allow_nil? false
    end

    attribute :slug, :string do
      allow_nil? false
    end

    attribute :content, :string do
      allow_nil? false
    end

    create_timestamp(:inserted_at)
    update_timestamp(:updated_at)
  end

 relationships do
    belongs_to :site, MyApp.Sites.Site
  end

  identities do
    identity :slug_site, [:slug, :site] # <-- Should be :site_id
  end
```

Since `:site` is a relationship, this would silently fail to enforce identity constraints.

At least the case of relationship identities could be validated at compile time in a new `Ash.Resource.Verifiers.VerifyIdentityFields` validator module and emit a warning (or perhaps error?) at compile time. There may be other scenarios that could be checked, but this stands out an footgun that should be caught automatically.

TwistingTwists:
* adds verifier for identity fields.
* raises error if verifier fails
* tests added

Closes #576.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/verifiers/verfiy_identities.ex
+++ b/lib/ash/resource/verifiers/verfiy_identities.ex
@@ -7,21 +7,17 @@ defmodule Ash.Resource.Verifiers.VerifyIdentityFields do
   def verify(dsl) do
     identities = Ash.Resource.Info.identities(dsl)
 
-    identity_names = Enum.flat_map(identities, fn identity -> identity.keys end)
+    for identity <- identities do
+      for key <- identity.keys do
+        if !Ash.Resource.Info.attribute(dsl, key) do
+          raise Spark.Error.DslError,
+            module: Spark.Dsl.Verifier.get_persisted(dsl, :module),
+            message: "All identity keys must be attributes. Got: #{inspect(key)}",
+            path: [:identities, identity.name]
+        end
+      end
+    end
 
-    relations =
-      dsl
-      |> Ash.Resource.Info.relationships()
-      |> Enum.filter(fn relation ->
-        relation.type == :belongs_to and relation.name in identity_names
-      end)
-
-    Enum.each(relations, fn rel ->
-      raise Spark.Error.DslError,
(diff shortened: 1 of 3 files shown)


zachdaniel:
I think a simpler check here would be a bit better. Specifically “is this field an attribute, calculation, or aggregate”. That would cover the case of using a relationship name in addition to cases like typos or things like that.

lib/ash/resource/verifiers/verfiy_identities.ex
```diff
@@ -0,0 +1,22 @@
...
+
+  def verify(dsl) do
+    identities = Ash.Resource.Info.identities(dsl)
+
+    for identity <- identities do
+      for key <- identity.keys do
+        if !Ash.Resource.Info.attribute(dsl, key) do
+          raise Spark.Error.DslError,
+            module: Spark.Dsl.Verifier.get_persisted(dsl, :module),
+            message: "no such attribute #{inspect(key)}"
```
zachdaniel, line 15: lets add a `path` to the error, `path: [:identities, identity.name]`.

Additionally, let's make the error message a bit clearer: "All identity keys must be attributes. Got: #{inspect(key)}"

zachdaniel (approved):

lib/ash/resource/verifiers/verfiy_identities.ex
TwistingTwists: done.

zachdaniel (approved):

zachdaniel (approved):

zachdaniel:

PR #632, opened 2023-06-25
feat: Adds Ash.Changeset around_transaction

sezaru:
This doesn't unit tests since I couldn't find unit tests for `around_action`, `before_transaction` or `after_transaction` to use as reference.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -26,6 +26,7 @@ defmodule Ash.Changeset do
     after_transaction: [],
     arguments: %{},
     around_action: [],
+    around_transaction: [],
     attributes: %{},
     before_action: [],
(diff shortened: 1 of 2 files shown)


zachdaniel:
@sezaru we should add a test at least confirming this hook is properly run. We also need to raise an error in bulk actions if they have around hooks configured

sezaru:
@zachdaniel I added a test for it.

Not sure where exactly I should raise an error when running bulk actions, any suggestion or some code I can use as reference?

zachdaniel:
I’ll handle that part :) thanks for adding the test!

zachdaniel (approved):

PR #636, opened 2023-06-29
Add `Negate` validation

marmor157:
Continue of work done in https://github.com/ash-project/ash/pull/587

> This PR adds negate validation that negates other validations. This allows to do things like negate(one_of(:attr, [:val1, :val2]).
> 
> Originally I wanted to name it not but when adding a builtin I've found that it might conflict with Kernel.not so I've renamed it to negate

### DIFF
--- a/lib/ash/resource/validation.ex
+++ b/lib/ash/resource/validation.ex
@@ -138,4 +138,6 @@ defmodule Ash.Resource.Validation do
         {:error, "Expected items of [:create, :update, :destroy], got: #{inspect(list)}"}
     end
   end
+
+  def validation_type, do: @validation_type
 end
--- a/lib/ash/resource/validation.ex
+++ b/lib/ash/resource/validation.ex
@@ -42,6 +42,7 @@ defmodule Ash.Resource.Validation do
 
   @callback init(Keyword.t()) :: {:ok, Keyword.t()} | {:error, String.t()}
   @callback validate(Ash.Changeset.t(), Keyword.t()) :: :ok | {:error, term}
+  @callback describe(Keyword.t()) :: [message: String.t(), vars: Keyword.t()]
 
   @validation_type {:spark_function_behaviour, Ash.Resource.Validation,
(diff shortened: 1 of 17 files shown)


zachdaniel:
Looks great! One last note: let's make `describe` an optional callback with `@optional_callbacks describe: 1`, and then in the `negate` validation we can do something like this. Or perhaps even just raise an error if the target module does not implement  description.
```elixir
def init(opts) do
   Code.ensure_compiled!(module)
   
   if function_exported?(module, :describe, 1) do
      {:error, ...}
   else
      {:error, "unknown validation"}
   end
end
```

marmor157:
`Code.ensure_compiled!(module)` is a great suggestion, thanks!

For the `@optional_callbacks` and erroring when there's do `describe/1` tho, what would be the advantage of that, over overridable default `describe/1` of `inspect(__MODULE__)`? IMO, when somebody creates his own validation and does not want to specify the description for it, the `inspect(__MODULE__)` is good fallback as it at least specifies which validation failed, rather than forcing developer to write the description.

zachdaniel:
What we want to avoid is leaking module names in error messages. Perhaps we could have a middle ground and make a special type of `:unknown` error with a special message, and return that. We don’t ever show unknown errors to users.

zachdaniel:
Although what we should actually do, come to think of it, is write a verifier on the resource DSL that checks all usage and raises a *compile time warning or error* if the describe callback isn’t defined.

marmor157:
> Although what we should actually do, come to think of it, is write a verifier on the resource DSL that checks all usage and raises a compile time warning or error if the describe callback isn’t defined.

Isn't it what `init/1` is doing? I've added this validation inside it and tested it, seems to work ok. 
Although I'm not entirely happy about `with_description` implementation now :/

zachdaniel:
Nice work! Like the consistency changes on returning errors. Ready to merge?

marmor157:
Ready 🎉

zachdaniel:

PR #642, opened 2023-07-09
fix: add `handle_change/3` and `prepare_change/3` to `Ash.Type.Union`

jimsynz:
This is a WIP change because I don't really know what I'm doing - I just did what Zach said.

### DIFF
--- a/lib/ash/type/union.ex
+++ b/lib/ash/type/union.ex
@@ -377,4 +377,106 @@ defmodule Ash.Type.Union do
       :error
     end
   end
+
+  @impl true
+  def handle_change(nil, %Ash.Union{type: type_name, value: new_value}, constraints) do
+    with {:ok, type_configs} <- Keyword.fetch(constraints, :types),
+         {:ok, type_config} <- Keyword.fetch(type_configs, type_name),
+         {:ok, type} <- Keyword.fetch(type_config, :type),
+         type_constraints <- Keyword.get(type_config, :constraints, []),
+         {:ok, new_value} <- type.handle_change(nil, new_value, type_constraints) do
+      {:ok, %Ash.Union{type: type_name, value: new_value}}
+    end
+  end
+
+  def handle_change(%Ash.Union{type: type_name, value: old_value}, nil, constraints) do
+    with {:ok, type_configs} <- Keyword.fetch(constraints, :types),
+         {:ok, type_config} <- Keyword.fetch(type_configs, type_name),
+         {:ok, type} <- Keyword.fetch(type_config, :type),
+         type_constraints <- Keyword.get(type_config, :constraints, []),
+         {:ok, new_value} <- type.handle_change(old_value, nil, type_constraints) do
+      {:ok, %Ash.Union{type: type_name, value: new_value}}
+    end
+  end
+
+  def handle_change(
+        %Ash.Union{type: type_name, value: old_value},
+        %Ash.Union{type: type_name, value: new_value},
+        constraints
+      ) do
+    with {:ok, type_configs} <- Keyword.fetch(constraints, :types),
+         {:ok, type_config} <- Keyword.fetch(type_configs, type_name),
+         {:ok, type} <- Keyword.fetch(type_config, :type),
+         type_constraints <- Keyword.get(type_config, :constraints, []),
+         {:ok, new_value} <- type.handle_change(old_value, new_value, type_constraints) do
+      {:ok, %Ash.Union{type: type_name, value: new_value}}
+    end
+  end
+
+  def handle_change(
+        %Ash.Union{},
+        %Ash.Union{type: type_name, value: new_value},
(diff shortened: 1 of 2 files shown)


lib/ash/type/union.ex
```diff
@@ -377,4 +377,53 @@ defmodule Ash.Type.Union do
...
+        constraints
+      ) do
+    with {:ok, type_configs} <- Keyword.fetch(constraints, :types),
+         {:ok, type_config} <- Keyword.fetch(type_configs, type_name),
+         {:ok, type} <- Keyword.fetch(type_config, :type),
+         {:ok, type_constraints} <- Keyword.fetch(type_config, :constraints),
+         {:ok, new_value} <- type.handle_change(old_value, new_value, type_constraints) do
+      {:ok, %Ash.Union{type: type_name, value: new_value}}
+    end
+  end
```
zachdaniel, line 414: We want to call `handle_change` on the new type as if the old value was `nil`.

zachdaniel: like when the types don't match I mean.

lib/ash/type/union.ex
```diff
@@ -377,4 +377,67 @@ defmodule Ash.Type.Union do
...
+
+  @impl true
+  # FIXME I don't really understand what to do with other ways this function is
```
zachdaniel, line 431: This one is essentially exactly the same as the `handle_change`, except we don't know the type of the new value here. It could be an `%Ash.Union{}` *or* it could be any value that will (or won't) cast to this type. So what we need to do is something like this:

```elixir
def prepare_change(_old_value, nil, _) do
  {:ok, nil}
end

def prepare_change(%Ash.Union{type: type_name}, new_value, constraints) do
  case cast_input(new_value, constraints) do
     {:ok, %Ash.Union{type: ^type_name, value: value}} ->
       do_prepare_change(other_type, nil, value, constraints)
     {:ok, %Ash.Union{type: other_type, value: value}} ->
        do_prepare_change(other_type, nil, value, constraints)
     {:error, _} ->
        {:ok, new_value}
   end
end

defp do_prepare_change(type_name, old_value, new_value, constraints) do
  # call `prepare_change` on the right type, and return `{:ok, %Ash.Union{type_name: type_name, value: prepared_value}}` on success
end
```

lib/ash/type/union.ex
```diff
@@ -377,4 +377,67 @@ defmodule Ash.Type.Union do
...
+         {:ok, type} <- Keyword.fetch(type_config, :type),
+         {:ok, type_constraints} <- Keyword.fetch(type_config, :constraints),
+         {:ok, new_value} <- type.handle_change(old_value, new_value, type_constraints) do
+      {:ok, %Ash.Union{type: type_name, value: new_value}}
+    end
+  end
+
+  def handle_change(
+        %Ash.Union{value: old_value},
+        %Ash.Union{type: type_name, value: new_value},
```
zachdaniel, line 418: we don't want to use the old value here, because they aren't of the same type. We want to do a `handle_change` with `nil` instead.

zachdaniel: this implementation could call something like `Integer.handle_change("foobar", 10)` which wouldn't make much sense.

lib/ash/type/union.ex
```diff
@@ -377,4 +377,79 @@ defmodule Ash.Type.Union do
...
+         {:ok, new_value} <- type.handle_change(old_value, new_value, type_constraints) do
+      {:ok, %Ash.Union{type: type_name, value: new_value}}
+    end
+  end
+
+  @impl true
+  # FIXME I don't really understand what to do with other ways this function is
+  # called.
+
+  def prepare_change(_old_value, "", _constraints), do: {:ok, nil}
```
zachdaniel, line 434: We shouldn't need the `""` case here. The type casting will handle if that resolves to a `nil` value according to the type.

zachdaniel: We need the `nil` case because thats a valid value of all types.

lib/ash/type/union.ex
```diff
@@ -377,4 +377,96 @@ defmodule Ash.Type.Union do
...
+        if new_value[to_string(field)] == tag,
+          do: do_prepare_change(type_name, old_value, new_value, constraints),
+          else: {:ok, new_value}
+
+      _ ->
+        {:ok, new_value}
+    end
+  end
+
+  def prepare_change(%Ash.Union{type: type_name, value: old_value}, new_value, constraints) do
```
zachdaniel, line 454: we also need:
```elixir
def prepare_change(nil, new_value, constraints) do
  with {:ok, %Ash.Union{type: type_name, value: value}} <- cast_input(new_value, constraints) do
    do_prepare_change(type_name, nil, value, constraints)
  end
end
```

zachdaniel (approved):

Issue #652, opened 2023-07-11
Some `:read` flow steps are erroring after update to Ash 2.11.3


sswrk:
**Describe the bug**

One of my flows errors after changing versions from 2.11.2 to 2.11.3. The latest version (2.11.4) doesn't work either

**To Reproduce**
1. I have a resource (let's call it `MyApp.Resource1`) that has a `:create` action. This action is manual (`MyApp.Actions.CreateResource1`).
2. The manual action `MyApp.Actions.CreateResource1` runs a flow `MyApp.Flows.CreateResource1` and passes all changeset attributes and arguments to it.
3. The `MyApp.Flows.CreateResource1` flow's steps look like that:

```
steps do
  transaction :maybe_create_resource_1, MyApp.Resource1 do
    branch :maybe_create_resource_2, expr(<some expression>) do
      run_flow :resource_2, MyApp.Flows.MaybeCreateResource2ForResource1 do
        input <some input>
      end
    end
    
    ...
  end
end
```
The `run_flow :resource_2` step is being run
4. The `MyApp.Flows.MaybeCreateResource2ForResource1` flow's steps look like that:

```
steps do
  branch :resource_2_by_id, expr(<some expression>) do
    read :existing_resource_2_by_id, MyApp.Resource2, :read do
      not_found_error? false
      get? true
      
      input <some input>
      
      tenant arg(:tenant)
    end
  end
  ...
end
```

Once it gets to the `:existing_resource_2_by_id` step, it errors:

```
** (Ash.Error.Unknown) Context: resolving data on perform MyApp.Resource1.create > resolving data on Transaction :maybe_create_resource_1
     
     Unknown Error
     
     Context: resolving intermediate_data on fetch MyApp.Resource2.read
     * Context: resolving intermediate_data on fetch MyApp.Resource2.read
     
     ** (ArgumentError) `nil` is not a Spark DSL module.
     
       nil.spark_dsl_config()
       (spark 1.1.18) lib/spark/dsl/extension.ex:134: Spark.Dsl.Extension.dsl!/1
       (spark 1.1.18) lib/spark/dsl/extension.ex:206: Spark.Dsl.Extension.fetch_opt/4
       (spark 1.1.18) lib/spark/dsl/extension.ex:172: Spark.Dsl.Extension.get_opt/5
       (ash 2.11.4) lib/ash/query/query.ex:302: Ash.Query.new/3
[...]

PR #657, opened 2023-07-18
Add test that recreates problem from issue 652

marmor157:
This PR adds test that recreates problem from https://github.com/ash-project/ash/issues/652


zachdaniel:

PR #664, opened 2023-07-20
fix: read argument value in `ArgumentEquals` validation

sevenseacat:
We had an almost-identical `ArgumentEquals` validation in our codebase, and when I removed it in favour of the new built-in one, some of our tests started failing. 

Upon comparing them, I noticed this issue - it wasn't reading the actual argument, instead looking for an attribute with the same name!


zachdaniel (approved):

zachdaniel:
🚀Thank you for your contribution!🚀

PR #666, opened 2023-07-26
add Ash.Policy.Check.Expr to builtin checks

barnabasJ:
I needed to add an expr to a policy condition, and the helper would have been nice to have.


zachdaniel:
This shouldn’t be necessary, as it should support expressions automatically. Does it not do that?

barnabasJ:
> This shouldn’t be necessary, as it should support expressions automatically. Does it not do that?

You are right. I have again made it more complicated than it is. :facepalm:

Issue #580, opened 2023-05-17
Defining an aggregate that returns a resource generates an Ecto error


sevenseacat:
**Describe the bug**

I had the case where I wanted to define an aggregate that returned a full resource instance, not a field on a resource, eg.

* Instance has_one :most_recent_instance_log
* Project has_many :instances
* Project has an aggregate for the most recent instance log across all of the instances

I imagined this aggregate definition to look like:

```elixir
aggregates do
  first :most_recent_instance_log, :instances, :most_recent_instance_log
end
```

But this gave me a big Ecto error:

```
iex(5)>  MyApp.Project.by_id("43d74c6a-23a1-4959-8070-a5591613f88b", load: [:most_recent_instance_log])
** (Ecto.SubQueryError) the following exception happened when compiling a subquery.

    ** (UndefinedFunctionError) function :has_one.type/0 is undefined (module :has_one is not available)

The subquery originated from the following query:

from p0 in MyApp.Project,
  as: 0,
  left_lateral_join: i1 in subquery(from i0 in MyApp.Instance,
  as: 0,
  where: is_nil(type(as(0).archived_at, {:parameterized, Ash.Type.UtcDatetimeUsec.EctoType, []})) == ^true,
  where: is_nil(type(as(0).archived_at, {:parameterized, Ash.Type.UtcDatetimeUsec.EctoType, []})) == ^true,
  where: parent_as(0).id == as(0).project_id,
  group_by: [i0.project_id],
  select: merge(
  %{
    most_recent_instance_log:
      type(
        coalesce(
          fragment("array_agg(?)", ^most_recent_instance_log),
          type(^[], {:array, :has_one})
        ),
        {:array, :has_one}
      )
  },
  map(i0, [:project_id])
)),
  as: 1,
  on: true,
  where: is_nil(type(as(0).archived_at, {:parameterized, Ash.Type.UtcDatetimeUsec.EctoType, []})) == ^true,
  where: type(as(0).id, {:parameterized, Ash.Type.UUID.EctoType, []}) ==
  type(^"43d74c6a-23a1-4959-8070-a5591613f88b", {:parameterized, Ash.Type.UUID.EctoType, []}),
  select: merge(
  struct(p0, [
    :archived_at,
    :id,
    :project_id,
    :api_enabled,
    :display_name,
    :inserted_at,
    :updated_at,
    :vendor_id
  ]),
  %{
    most_recent_instance_log:
      type(
        coalesce(
[...]

PR #671, opened 2023-08-12
Feature: Ensure aggregate field is attribute or calculation. 

franckstifler:
This PR solves this bug: https://github.com/ash-project/ash/issues/580

- [ ] Documentation changes

### DIFF
--- a/lib/ash/resource/verifiers/ensure_aggregate_field_is_attribute_or_calculation.ex
+++ b/lib/ash/resource/verifiers/ensure_aggregate_field_is_attribute_or_calculation.ex
@@ -5,19 +5,18 @@ defmodule Ash.Resource.Verifiers.EnsureAggregateFieldIsAttributeOrCalculation do
   use Spark.Dsl.Verifier
 
   def verify(dsl) do
-    relationships = Ash.Resource.Info.relationships(dsl)
     aggregates = Ash.Resource.Info.aggregates(dsl)
 
     for %{name: name, relationship_path: paths} <- aggregates do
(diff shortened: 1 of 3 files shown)


lib/ash/resource/verifiers/ensure_aggregate_field_is_attribute_or_calculation.ex
```diff
@@ -0,0 +1,31 @@
...
+  def verify(dsl) do
+    aggregates = Ash.Resource.Info.aggregates(dsl)
+
+    for %{name: name, relationship_path: paths} <- aggregates do
+      case Ash.Resource.Info.related(dsl, paths) do
+        nil ->
+          nil
+
+        destination ->
+          attribute = Ash.Resource.Info.attribute(destination, name)
```
zachdaniel, line 16: What we need to use is the aggregate's `field` here (not the name). If the field is `nil` then we shouldn't check it.

zachdaniel (changes requested):

lib/ash/resource/verifiers/ensure_aggregate_field_is_attribute_or_calculation.ex
franckstifler: Fixed:
- if field is nil, we don't check if it's in aggregates or calculations.
- Used `field` instead of `name`.
- Added a test that raises Spark.Error.DslError when an invalid field is specified.

zachdaniel (approved):

zachdaniel:

PR #672, opened 2023-08-14
fix: ensure union tag_values are compared as string

rgraff:
When casting input for a union type, ensure the tag value is a string.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Thanks for the fix! I had to make some changes to it, so I pushed those changes to main. 🙇

PR #680, opened 2023-08-24
fix(Ash.Type.Binary): Base64 encode binaries when embedding.

jimsynz:
So when we serialise an embedded type into JSON we can't just throw a blob in, or it fails.

Example from trying to store a small PNG:

```
** (Jason.EncodeError) invalid byte 0xFF in <<255, 216, 255, 224, 0, 16, 74, 70, 73, 70, 0, 1, 2, 0, 0, 1, 0, 1, 0, 0, 255, 192, 0, 17, 8, 1, 0, 1, 0, 3, 1, 17, 0, 2, 17, 1, 3, 17, 1, 255, 219, 0, 67, 0, 8, 6, 6, 7, 6, 5, ...>>
  (jason 1.4.1) lib/jason.ex:213: Jason.encode_to_iodata!/2
```

Now, I am not sure if this is a good solution (actually I'm pretty sure it's not). It'd be cool if there was a `cast_embedded/2` callback specifically.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/binary.ex
+++ b/lib/ash/type/binary.ex
@@ -24,7 +24,13 @@ defmodule Ash.Type.Binary do
   def cast_stored(nil, _), do: {:ok, nil}
 
   def cast_stored(value, _) do
-    Ecto.Type.load(:binary, value)
+    case Ecto.Type.load(:binary, value) do
+      {:ok, value} -> case Base.decode64(value) do
+        {:ok, value} -> {:ok, value}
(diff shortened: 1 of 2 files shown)


zachdaniel:
Yeah i think the overhead of this can be of this will be pretty significant, as type conversion functions are called often. We should likely just implement the cast_stored_embedded.

PR #691, opened 2023-09-07
improvement(code interface): warn when `define_for` is missing

joshprice:
When `define_for` is missing no `code_interface` functions are generated and no explanation is provided why.

# Contributor checklist

- [x] Documentation changes

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -186,40 +186,36 @@ defmodule Ash.CodeInterface do
   @doc """
   Defines the code interface for a given resource + api combination in the current module. For example:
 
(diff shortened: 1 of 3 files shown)


zachdaniel:
So `define_for` is only one way of using the code interface. Check out the code interface guide for more.

joshprice:
Do you mean this?

https://github.com/ash-project/ash/blob/1ddba9499c75fcf92669a578f473f4eb71c266b8/documentation/topics/code-interface.md#L15

Could we generate the functions that throw errors showing the warning?

zachdaniel:
Yes, we could do that :) I think its a great idea.

PR #700, opened 2023-09-21
fix: more usages of `:wrap_list` type

vonagam:
Saw that type for `keys` in `identity` was document as `any`. Thought why wouldn't it show `atom | list(atom)`? Found out that there are custom type validations that just do what `:wrap_list` handles. Replaced them.

In one case of `Ash.Notifier.PubSub.Publication.topic` there was even a small bug - it would allow a single string or a list of string/atoms, but not a single atom.

Note: `{:wrap_list, x}` produces type documentation `list(x) | x`. Personally I would find it more conventional for it to be `x | list(x)` instead - from simpler to more complex.


zachdaniel (approved):

zachdaniel:
I agree about the type docs. That would involve a PR to spark. Would you like to take a stab at that?

zachdaniel:

PR #701, opened 2023-09-22
feat: expose query or changeset to field policies

barnabasJ:
This would make checks like `accessing_from` work with field policies.

I couldn't think of a reason not to add this. Please let me know if you see it the same way or if there are possible security implications. 

If it is ok to do, I will add tests. 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -30,6 +30,8 @@ if Mix.env() == :test do
   config :ash, :validate_api_resource_inclusion?, false
   config :ash, :validate_api_config_inclusion?, false
 
+  config :logger, level: :debug
+
   config :ash, :policies,
     show_policy_breakdowns?: true,
     log_policy_breakdowns: :info,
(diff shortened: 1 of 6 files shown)


zachdaniel:
Hmm....so there are issues with `accessing_from` in field policies because field policies are run in contexts where you do a filter i.e `Ash.Query.filter(relationship.field == "foo")`...if we can add some tests that confirm this works for `accessing_from` in those contexts, I'd be cool adding it.

zachdaniel:
Yeah, I'm realizing now that we actually already essentially had to solve that problem for just regular access policies in filters

zachdaniel:
I think it should work! 👍

zachdaniel:

zachdaniel (approved):

PR #702, opened 2023-09-22
fix: revert `:wrap_list` usage for `topic` in PubSub, needs recursion

vonagam:
Revert one of the changes in #700. I've missed that the custom function supports recursion.

Also in that PR I've made remark that with custom function `topic` supports `"string"` or `["string"]` or `[:atom]` but not `:atom`. I've kept it like it was before, I assume this is because atoms represent changing parts and it is not advisable for topic to consist only of an attribute value. Though there is a potential `prefix`, so maybe single atom is ok.

Also on topic of `:wrap_list` removed `List.wrap` call `Ash.Resource.Validation` that is not needed anymore because `:wrap_list` does that now.


zachdaniel:

PR #705, opened 2023-09-25
improvement: default `keys` in identity to its `name`

vonagam:
Allows to write `identity :email` instead of `identity :email, :email` or `identity :email, [:email]`.


zachdaniel:
🤔 interesting. It would potentially be strange because you couldn't do `identity [:foo, :bar]`, so you'd have to know both formats. Not necessarily sold that this is a good idea.

vonagam:
No problem with the closing. But for record: 
> It would potentially be strange because you couldn't do `identity [:foo, :bar]`

From my perspective this one is already solved by provided error message if you write it like that:
> invalid value for :name option: expected atom, got: [:foo, :bar]

PR #709, opened 2023-09-27
fix: move `Resource.Transformers.BelongsToSourceField` into entity

vonagam:
Unless I am missing something there is no reason for `Resource.Transformers.BelongsToSourceField` to be a whole resource transform and not `Relationships.BelongsTo` entity transform.

Also:

- There was missing `@doc false` for `transform` in `Relationships.ManyToMany`.

- `DefaultPrimaryKey` transformer had wrong `after?` case: it should go after `BelongsToAttribute` (which adds an attribute, possible primary key one) and not after `BelongsToSourceField` (which simply sets a default name for the attribute).


zachdaniel (approved):

zachdaniel:

PR #710, opened 2023-09-27
add test showing the problem with expressions in field_policy conditions

barnabasJ:
I was working with field policies, and some things didn't work as I would have expected them to. 

I tried to recreate these situations in a test and added them here. 

I tried to debug it but was unable to find a solution so far.

### DIFF
--- a/test/policy/field_policies/expr_conditions_test.exs
+++ b/test/policy/field_policies/expr_conditions_test.exs
@@ -1,0 +1,178 @@
+defmodule Ash.Test.Policy.FieldPolicy.ExpressionConditionTest do
+  use ExUnit.Case, async: true
+
+  defmodule Api do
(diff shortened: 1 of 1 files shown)


zachdaniel:
@barnabasJ I've just pushed up to make it properly set those options. Does that resolve the issue? If so, we should still commit the tests confirming the fixed behavior.

barnabasJ:
I have to check but probably not. I added the options myself in ash_rbac but there were still problems where the expression didn't apply and I would still get a forbidden field even though a different policy would allow me to read the field. And I think because the expression is in the condition and not a `authorize_if` inside the block, it should work if the other policy would allow it

barnabasJ:
I will rebate this branch asap. That should lead to the tests failing in a different way

PR #712, opened 2023-09-28
fix: usage of `get_by` with `belongs_to` attribute

### LINKED ISSUE #637: bug: read action get_by don't recognize relationship attributes
**Describe the bug**

If you have a relationship in your resource, for example with `belongs_to`, `get_by` shortcut for a `read` action will not find the correspondent relationship `id` attribute.

**To Reproduce**

Add a relationship to your resource like this:

```elixir
  belongs_to :customer, Customer
```

Now create a read action like this:

```elixir
    read :get_by_customer_id do
      get_by :customer_id
    end
```

**Expected behavior**

`get_by` should find the relationship `id` attribute and use it

**Runtime**
 - Elixir version 1.15.0
 - Erlang version 26.0.1
 - OS Fedora 38
 - Ash version 2.11.0-rc.3

vonagam:
`get_by` transform should happen after `belongs_to` attributes are generated.

Closes #637.

### DIFF
--- a/lib/ash/resource/transformers/get_by_read_actions.ex
+++ b/lib/ash/resource/transformers/get_by_read_actions.ex
@@ -14,6 +14,7 @@ defmodule Ash.Resource.Transformers.GetByReadActions do
 
   @doc false
   @spec after?(module) :: boolean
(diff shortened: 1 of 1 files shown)


zachdaniel:

PR #713, opened 2023-09-28
update test to load forbidden field through union

barnabasJ:
I found another case where field_policies were not respected correctly


zachdaniel:

PR #714, opened 2023-09-29
fix: return unions as is if load is empty

barnabasJ:
After the change you made to fix the permissions when loading through unions, I got another error.

This might not be the right place to fix this issue, but it got rid of the error.

What happened was that I tried to execute an update on a resource, and the resource I put in had a %Ash.ForbiddenField{} struct in place for the union. Ash then tried to go through the load function and failed
because the Enum function couldn't handle the %Ash.ForbiddenField{} value. 

This is not actually fixing this, but because I didn't need to load anything from the union in this action, the load was empty, and I could circumvent the problem like this.

### DIFF
--- a/lib/ash/type/type.ex
+++ b/lib/ash/type/type.ex
@@ -834,6 +834,7 @@ defmodule Ash.Type do
           {:ok, list(term)} | {:error, Ash.Error.t()}
   def load(_, [], _, _, _), do: {:ok, []}
   def load(_, nil, _, _, _), do: {:ok, nil}
+  def load(_, %Ash.ForbiddenField{} = value, _, _, _), do: {:ok, value}
 
   def load({:array, type}, values, loads, constraints, context) do
     load(type, values, loads, constraints[:items] || [], context)
--- a/lib/ash/type/union.ex
+++ b/lib/ash/type/union.ex
@@ -122,6 +122,9 @@ defmodule Ash.Type.Union do
   def storage_type(_), do: :map
 
   @impl true
+  def load(unions, [], _, _), do: {:ok, unions}
+
+  @impl true
   def load(unions, load, constraints, context) do
     unions
     |> Stream.with_index()


barnabasJ:
> I think the better fix is to not try to load values on an `%Ash.ForbiddenField{}` perhaps short circuiting in Ash.Type.load?

That sounds good to me. If that has no other side effects, it might be okay to leave this change in there too. As it doesn't make sense to loop over the values if the load is empty?

zachdaniel:
I think it's important to do actually, because we want to trigger a minimal load on those values to do field policies on them, although maybe I'm just now realizing that isn't necessary...does this change not impact your field policies at all?

barnabasJ:
No, I didn't see any changes in my policy tests after this change.

zachdaniel (approved):

zachdaniel:

This does point out an interesting issue: field policies are going to interact with calculation dependencies in an unanticipated way that will need to be dealt with under the hood somehow. Specifically, calculations need to get the values without field policies applied, but then they need to be returned with them applied after-the-fact. We will need to do one of two things:

1. store the actual value of the fields that were hidden via policy, and "rehydrate" them when provided as a calculation dependency.
2. Make field policies apply to type loads in a second-pass via a new callback, something like `def apply_field_policies`. I'm considering that we may actually need to change up the `def load` pattern on types to something like `def get_contained_resources` which returns a list or a map of keys to lists (for unions) to the resources contained inside the type. Then Ash will handle the loading and call something like `set_contained_resources` on the type to say "okay we worked on them, now put them back". Its complex though, will need to mull it over.

barnabasJ:
Yeah, I already have a case where this would happen to one of my calculations. let me know where I can help.

zachdaniel:
Damn, was hoping I could put it off for a bit 😆 I don't really have a great answer yet. Need to think about it a bit.

barnabasJ:
> Damn, was hoping I could put it off for a bit 😆 I don't really have a great answer yet. Need to think about it a bit.

It's not something sensitive, just something to clean up the data a bit. So it's not too bad in my case. I could also rewrite the calc to circumvent the problem if I really needed to. Take your time.

PR #715, opened 2023-10-01
improvement: support an `api` option to `use Ash.Resource`

zachdaniel:
improvement: add functions to `Ash` for resources w/ configured apis
improvement: default code_interface.define_for to resource's ash api

So what does this give us?

```elixir
defmodule Resource do
  use Ash.Resource,
    api: YourApi
end
```

Given that resource, you can then use the functions defined in `Ash` that mirror the functions defined in the api module,

```elixir
Resource
|> Ash.Changeset.for_create()
|> Ash.create!()
```

Additionally, options like `code_interface.define_for` default to the value of that api, allowing you to omit them.

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -5,6 +5,14 @@ defmodule Ash do
   Currently only contains setters/getters for process context.
   """
 
+  for {function, arity} <- Ash.Api.Functions.functions() do
+    unless function in Ash.Api.Functions.no_opts_functions() do
+      args = Macro.generate_arguments(arity, __MODULE__)
+
+      defdelegate unquote(function)(unquote_splicing(args)), to: Ash.Api.GlobalInterface
+    end
+  end
+
   @doc """
   Converts a context map to opts to be passed into an action.
   """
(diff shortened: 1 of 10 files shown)


zachdaniel:
We need to stew on this, but putting it up for feedback.

jimsynz:
I love this

I know we had a big conversation about this on discord where I was very grumpy sounding but I think this is a **huge** improvement to Ash's DX and I'm all for it.

lib/ash.ex
```diff
@@ -5,6 +5,14 @@ defmodule Ash do
   Currently only contains setters/getters for process context.
```
jimsynz, line 5: pants on fire.

zachdaniel:
The biggest issue with this proposal is that it *drastically* changes the way that most of our documentation, examples and the general usage pattern of Ash looks. It may be good to do this in 3.0 just so that we have the ability to properly document it and update all of our guides.

rgraff:
Big fan

zachdaniel:
Alright, I'm merging this. The previous way of doing things will continue to work going forward, but when we do our 3.0 documentation review, we will switch the default to being this. Then, calling individual Api modules directly will be something that you'd only need to do in special cases.

zachdaniel:
Although undocumented, if you are adventurous feel free to try this pattern out :)

PR #718, opened 2023-10-02
fix: fix types for `min`/`max` constraints in `Type.Float`

vonagam:
Seems like `Type.Float` was copied from `Type.Decimal` but the types were left as is from decimal. Unless I am missing something it should not work as there is no `decimal` method in `Type.Float`.


zachdaniel (approved):

zachdaniel:

PR #721, opened 2023-10-03
improvement: add code_interface arg/attr names/descriptions in docs

rellen:
WIP

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -407,8 +407,55 @@ defmodule Ash.CodeInterface do
           """
         end
 
+        arg_docs =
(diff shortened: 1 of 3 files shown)


zachdaniel:
Awesome! The main things I'd suggest: pull out what is positional and what isn't (which is a bit complex but there should be examples of it elsewhere in this file), which would be sets or accepted attributes. Then the positional arguments can get there own section and the rest go after. We shouldn't necessarily group them by arg/attribute as from the callers perspective it's not as important. Great stuff!

PR #735, opened 2023-10-14
Add ash.gen.resource generator

wintermeyer:
Generator for a resource.

$ mix ash.gen.resource Shop Product products name:string price:integer 

generator

### DIFF
--- a/lib/mix/tasks/ash.gen.resource.ex
+++ b/lib/mix/tasks/ash.gen.resource.ex
@@ -44,11 +44,13 @@ defmodule Mix.Tasks.Ash.Gen.Resource do
   defp parse_args(args) do
     {opts, positional_args, _} = OptionParser.parse(args, switches: [no_code_interface: :boolean])
 
-    with [api_name, resource_name, table_name | attribute_args] <- positional_args do
-      attributes = parse_attributes(attribute_args)
-      {:ok, api_name, resource_name, table_name, attributes, opts[:no_code_interface] || false}
-    else
-      _ -> :error
+    case positional_args do
+      [api_name, resource_name, table_name | attribute_args] ->
+        attributes = parse_attributes(attribute_args)
+        {:ok, api_name, resource_name, table_name, attributes, opts[:no_code_interface] || false}
+
+      _ ->
+        :error
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/ash.gen.resource.ex
```diff
@@ -0,0 +1,235 @@
...
+  end
+
+  defp uses_ash_postgres? do
+    mix_file = "mix.exs"
+    content = File.read!(mix_file)
+
+    # Regex to find ash_postgres in deps
+    # This approach has limitations, and there are many ways in which deps can be defined,
+    # but it should work for standard dep definitions.
+    Regex.match?(~r/ash_postgres/, content)
```
zachdaniel, line 223: I'd use `Code.ensure_loaded?(AshPostgres)` as opposed to reading the mix file. Also, I've thought about this a lot and I think we shouldn't make assumptions like this, but rather make it interactive. i.e "choose a data layer, here are the available data layers", that kind of thing.

zachdaniel: and support `--data-layer AshPostgres.DataLayer` otherwise.

lib/mix/tasks/ash.gen.resource.ex
```diff
@@ -0,0 +1,230 @@
...
+      if String.downcase(response) not in ["y", "yes", ""] do
+        IO.puts("Aborted!")
+        exit(:normal)
+      end
+    end
+
+    File.write!(resource_file_path, resource_file_content)
+    IO.puts("Generated #{resource_file_path}")
+
+    if uses_ash_postgres?() do
```
zachdaniel, line 204: Honestly, I'm still pretty torn about putting `ash_postgres` specific code generation in Ash core. The idea that we're working towards behind the scenes in terms of generators is that we would compose generator modifiers from various extensions that would be able to add their own snippets. i.e `--with postgres,graphql` or something like that. While having `postgres` configured automatically would be great I'd rather, at least to start, generate the resource without `ash_postgres` related code for now. We can always add it back in in a follow up.

zachdaniel:
We want to use `Mix.shell()` to print, as opposed to `IO.puts`

zachdaniel:
Closing this for now, as we are working on igniter which will be released before long

Issue #725, opened 2023-10-04
Error on unknown keys in 3.0


zachdaniel:
When calling an action and providing extra input, you should get an error when passing in unknown keys.

fcapovilla:
I've created a pull request for this issue.

zachdaniel:
The PR linked above solves this issue for generic actions, but similar logic will need to be added for queries and changesets to account for other action types.

zachdaniel:
This is complete in the 3.0 branch

PR #750, opened 2023-10-22
feat: Add unknown argument error when passing an unknown key to an action.

fcapovilla:
Solves #725 

When encountering unknown arguments, unknown keys are added to the `invalid_keys` set and an `Ash.Error.Action.InvalidArgument` with a custom message is added to the ActionInput.

Running the action will then return an `Ash.Error.Invalid` error.

### DIFF
--- a/test/actions/generic_actions_test.exs
+++ b/test/actions/generic_actions_test.exs
@@ -84,6 +84,19 @@ defmodule Ash.Test.Actions.GenericActionsTest do
         |> Api.run_action!()
       end
     end
+
+    test "generic actions don't accept unknown keys" do
+      assert {:error, %Ash.Error.Invalid{}} =
+               Post
+               |> Ash.ActionInput.for_action(:hello, %{name: "fred", one: 1})
+               |> Api.run_action()
+
+      assert_raise Ash.Error.Invalid, ~r/Input Invalid/, fn ->
+        Post
+        |> Ash.ActionInput.for_action(:hello, %{name: "fred", one: 1})
+        |> Api.run_action!()
+      end
+    end
   end
 
   describe "authorization" do
--- a/test/actions/generic_actions_test.exs
+++ b/test/actions/generic_actions_test.exs
@@ -2,6 +2,8 @@ defmodule Ash.Test.Actions.GenericActionsTest do
   @moduledoc false
   use ExUnit.Case, async: true
 
+  require Ash.Flags
+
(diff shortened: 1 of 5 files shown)


zachdaniel:
👋 Hey there! So this looks exactly right for generic actions, however we want this to happen for changesets and queries as well. With that said, this PR does not have to include those implementations as well. However, we only want to have this take effect in Ash 3.0. What I'd like to do is create a configuration flag that allows users to opt-in to 3.0 related behaviors, so that we can start introducing these breaking changes. Would you be interested in adding that configuration flag, and then using it here? We would add a flag in this file like `ash_three?`: https://github.com/ash-project/ash/blob/main/lib/ash/flags.ex

fcapovilla:
No problems, I've added the `ash_three?` flag in this pull request so the new behaviors only activate when the flag is `true`.

I had some problems making my test pass because the flag is `false` by default at compile time, so I've forced `ash_three?` to `true` when tests are running.

fcapovilla:
I thought about it a little more, and I guess it's better to test all cases, with and without the flag.

So I removed the forced config, and we can now test the two cases by using the `FLAG_ASH_THREE` environment variable. For example : `FLAG_ASH_THREE=true mix test`

zachdaniel (dismissed):

test/actions/generic_actions_test.exs
```diff
@@ -84,6 +86,26 @@ defmodule Ash.Test.Actions.GenericActionsTest do
         |> Api.run_action!()
       end
     end
+
```
zachdaniel, line 89: Let's do this with a tag. `@tag :ash_three`

zachdaniel: Approved the PR, but realized we should do this with tags. So then we should update the test helper file to exclude this by default I think.

fcapovilla: Ok, I'll add the tag tonight. :+1:

fcapovilla: I've added the tag. I've also updated the test_helper so the ash_three tests are excluded unless the `FLAG_ASH_THREE` is set. This way, we only need to specify the environment variable when running the tests. For example: `FLAG_ASH_THREE=true mix test`

zachdaniel: Nice :)

zachdaniel (approved):

zachdaniel:

Issue #760, opened 2023-10-27
Add :replace_all and :replace_all_except to upsert_fields


sezaru:
**Is your feature request related to a problem? Please describe.**
`upsert_fields` only accepts a list of fields to update, if you have a bunch of fields, then this option will make your action very big. For example, this is the create action of one of my resources:

```elixir
    create :create do
      primary? true

      upsert? true
      upsert_identity :unique_id

      upsert_fields [
        :external_id,
        :external_id_type,
        :cbsa_code,
        :msa_code,
        :minor_civil_division_code,
        :house_number,
        :street_direction,
        :street_post_direction,
        :street_name,
        :street_suffix,
        :unit_prefix,
        :unit_value,
        :city,
        :county,
        :county_fips,
        :state,
        :zip,
        :zip_4,
        :country,
        :geography,
        :year_built,
        :lot_square_feet,
        :lot_depth,
        :lot_width,
        :gross_area,
        :living_area,
        :garage_area,
        :building_area,
        :baths,
        :partial_baths,
        :bedrooms,
        :rooms,
        :units,
        :stories,
        :parking_spaces,
        :zoning_code,
        :type,
        :roof_type,
        :construction_type,
        :air_conditioning_type,
        :garage_types,
        :basement_type,
        :water_source,
        :home_owners_association_type,
        :multiple_listing_service_status,
        :property_classification,
        :last_sale,
        :mortgage,
        :assessment,
        :tax,
        :first_owner_name,
        :second_owner_name,
        :contact,
        :publication_date,
        :upload_date,
        :transaction_time,
        :deleted?
      ]
    end
 ```

**Describe the solution you'd like**
I believe we can do the same thing Ecto does, we replace the ecto list with the following options:

```elixir
    create :create do
      upsert? true
      upsert_identity :unique_id
      upsert_fields :replace_all
    end
```

or

```elixir
    create :create do
      upsert? true
      upsert_identity :unique_id
      upsert_fields {:replace_all_except, [:some_field]}
    end
```

or

```elixir
    create :create do
      upsert? true
      upsert_identity :unique_id
      upsert_fields {:replace, [:some_field]}
    end
```

PR #761, opened 2023-10-27
feat: Expand upsert_fields option

sezaru:
With this change, upsert_fields accepts the following options:

* `nil`: upsert fields not set;
* `:replace_all`: all fields from resource will be updated;
* `{:replace_all_except, [fields]}`: all fields from resource except fields inside `fields` will be updated;
* `{:replace, [fields]}`: only fields inside `fields` will be updated.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/dsls/DSL:-Ash.Resource.cheatmd
+++ b/documentation/dsls/DSL:-Ash.Resource.cheatmd
@@ -4771,7 +4771,7 @@ end
       
   </td>
   <td style="text-align: left">
-    <code class="inline">list(atom)</code>
+    <code class="inline">:replace_all | {:replace, atom | list(atom)} | {:replace_all_except, atom | list(atom)} | atom | list(atom)</code>
   </td>
   <td style="text-align: left">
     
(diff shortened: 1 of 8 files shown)


sezaru:
This is a PR for issue https://github.com/ash-project/ash/issues/760

One thing that I don't know how to do is translate this `nil | :replace_all | {:replace, list(atom)} | {:replace_all_except, list(atom)}` type spec into one for spark, for now I used `:any` just to make it compile.

@zachdaniel any suggestion on how to fix that?

zachdaniel:
Also, we need to keep supporting the old syntax, otherwise this would be a breaking change.
Have you had a chance to test this with `ash_postgres`? 

The type for that would look something like this:
```elixir
{:or, [
  {:wrap_list, :atom}, # preserve the old behavior
  {:literal, :replace_all},
  {:tuple, [{:literal, :replace}, {:wrap_list, :atom}]},
  {:tuple, [{:literal, :replace_all_except}, {:wrap_list, :atom}]},
  ]
}
```

zachdaniel:
Would also want to see some tests for the newly supported variations of this.

sezaru:
Done @zachdaniel

jimsynz:
This looks great @sezaru.  

CI is failing because your DSL changes would change the DSL cheat sheets - you can run `mix spark.cheat_sheets` in the project root and commit the changes to fix it.

Can you confirm that the correct behaviour also present with AshPostgres?

zachdaniel (approved):

Issue #765, opened 2023-11-07
Log a warning when `writable? false` causes foreign key updates to be ignored


col:
**Is your feature request related to a problem? Please describe.**
I was stuck for some time trying to update a relationship. In the end I discovered it was because the relationship with `writable? false` by default. Feel like a more targeted error message here would potentially save me and others a lot of ttime.

**Describe the solution you'd like**
A clear and descriptive warning message when an update is blocked by the `writable? false` option.

**Describe alternatives you've considered**
This could also be mitigated by clearer documentation but once I knew what I was looking forward it was documented. This issue here is that's it's still a slightly surprising behaviour for new comers so the warning message may be the most helpful.

**Express the feature either with a change to resource syntax, or with a change to the resource interface**

For example

```elixir
iex(5)> CartItem |> Ash.Changeset.for_create(:create, %{cart_id: cart.id, product_id: product.id, quantity: 1}) |> Shop.create!()
** (Ash.Error.Invalid) Input Invalid 
* relationship product is required <- The product_id was provided by was ignored
* relationship cart is required <- The cart_id was provided by was ignored 
    (ash 2.16.1) lib/ash/api/api.ex:2309: Ash.Api.unwrap_or_raise!/3
```

**Additional context**
Discussion on Discord: https://discord.com/channels/711271361523351632/711271361523351636/1171058521874378772

zachdaniel:
Closing as duplicate of #725

PR #766, opened 2023-11-07
Add a warning when not updating an attribute due to writable option

col:
This is related to this issue: https://github.com/ash-project/ash/issues/765 

Small change to provide a warning log when attempting to update an attribute that is not writable. 

Example:
```
00:41:19.924 [warning] Cannot change `product_id` because it's not writable.
```

The problem with this change is that it creates a lot of noise in the tests relating to the `id` attribute. 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1150,6 +1150,9 @@ defmodule Ash.Changeset do
           if attr.writable? do
             change_attribute(changeset, attr.name, value)
           else
+            Logger.warning(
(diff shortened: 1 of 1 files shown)


zachdaniel:
I think the noise for this is too high. We're discussing making it an error outright to provide additional unaccepted keys in 3.0.

zachdaniel:
Closing as it will be handled by #725

PR #770, opened 2023-11-12
fix: separated the relationship's name from its resolution 

simpers:
This is necessary so that the name can be used for the error message, should the relationship resolve to nil.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/policy/check/relates_to_actor_via.ex
+++ b/lib/ash/policy/check/relates_to_actor_via.ex
@@ -73,21 +73,21 @@ defmodule Ash.Policy.Check.RelatesToActorVia do
 
   defp relationship_info(resource, path, to_many? \\ false)
 
-  defp relationship_info(resource, [rel], to_many?) do
-    rel = Ash.Resource.Info.relationship(resource, rel)
+  defp relationship_info(resource, [rel_key], to_many?) do
+    rel = Ash.Resource.Info.relationship(resource, rel_key)
 
     if !rel do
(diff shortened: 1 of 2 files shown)


lib/ash/policy/check/relates_to_actor_via.ex
```diff
@@ -73,23 +73,26 @@ defmodule Ash.Policy.Check.RelatesToActorVia do
...
 
-    if !rel do
-      raise "No such relationship #{rel} for #{resource}, required in `relates_to_actor` check"
-    end
+    raise_if_nil(rel, rel_key, resource)
 
     relationship_info(rel.destination, rest, to_many? || rel.cardinality == :many)
   end
+
+  defp raise_if_nil(nil, rel_key, resource) do
```
jimsynz, line 92: I'm not sure that `raise_if_nil` is the right name for this function - can you think of something better?

jimsynz (commented):
Thanks @simpers for the contribution 🎉

lib/ash/policy/check/relates_to_actor_via.ex
simpers: I wasn't sure if you wanted to go down this route at all. But if you think it's fine to do it this way then maybe something like `raise_if_relationship_not_found`, `guard_relationship_exists`, or something along those lines?

zachdaniel:

PR #786, opened 2023-11-30
Api.get! ignores the action option

totaltrash:
I've just updated an old app from Ash 2.6.1 to 2.17.7 and found an interesting bug I think. Providing the action option to `get!` seems to now ignore the provided action and uses the primary?

```elixir
# Used to work as expected in 2.6.1, now gets a record without the loads provided in the details action
Portal.App.get!(Portal.App.Timesheet, id, action: :details)
```

### DIFF
--- a/test/actions/read_test.exs
+++ b/test/actions/read_test.exs
@@ -61,6 +61,10 @@ defmodule Ash.Test.Actions.ReadTest do
         prepare(PostPreparation)
       end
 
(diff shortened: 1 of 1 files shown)


zachdaniel:
Thanks for the reproduction test, going to merge and see what I can do.

zachdaniel:
This is fixed in main

PR #790, opened 2023-12-02
Clarify how to use Enums in attributes

ibarakaiev:
I was struggling for a bit with using Enums for a reason that now seems trivial, so I wanted to clarify the documentation in case anyone else would find this non-intuitive. This PR just changes the documentation to include an example of Enum usage:

```elixir
attribute :status, MyApp.TicketStatus
```

(I was originally trying to just run `attribute :status, :atom, constraints: [one_of: MyApp.TicketStatus]` instead.)


zachdaniel (approved):

zachdaniel:

PR #792, opened 2023-12-05
Fix filter test when relationship is nil

jechol:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

In the expression length(roles || []), one might expect the result to be 0 when roles is nil, as it seems to evaluate to length([]). However, in reality, it results in nil.


zachdaniel:
As of latest `main`, this should actually be fixed. In `iex`:

```elixir
Ash.Expr.eval(length(nil || []))
{:ok, 0}
```

PR #800, opened 2023-12-09
Add a cross-API many-to-many example to the documentation

ibarakaiev:
With the current documentation, it's unclear how to set up a cross-API `many_to_many` relationship if the join resource is in the source API but the destination resource is in a different API (the current example only works for when both of them are in the destination API). I added an example connecting Organizations and Users with guidance on how to set up the additional `has_many` association. The current error message was helpful in helping me realize how to do it, but it still took me a little while to nail everything.

In addition, I added a `postgres` and `actions` block to the example of a join resource, as currently the documentation doesn't list them but shows the full module code, which to me implied they weren't necessary. The error I faced because I left them out wasn't quite on point either: it just said that there was no read action on the `User` resource, which wasn't true, and the real problem was that there was no read action on the join resource.

### DIFF
--- a/documentation/topics/relationships.md
+++ b/documentation/topics/relationships.md
@@ -194,7 +194,7 @@ See the docs for more: `d:Ash.Resource.Dsl.relationships.many_to_many`
 
 ### Relationships across APIs
 
-You will need to specify the `api` option in the relationship if the destination resource and the join table are both parts of a different API:
+You will need to specify the `api` option in the relationship if the destination resource and/or the join table are parts of a different API:
 
 ```elixir
 many_to_many :authors, MyApp.OtherApi.Resource do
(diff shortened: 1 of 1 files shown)


documentation/topics/relationships.md
```diff
@@ -185,7 +194,7 @@ See the docs for more: `d:Ash.Resource.Dsl.relationships.many_to_many`
 
 ### Relationships across APIs
 
-You will need to specify the `api` option in the relationship if the destination resource is part of a different API:
+You will need to specify the `api` option in the relationship if the destination resource and the join table are both parts of a different API:
```
zachdaniel, line 197:
```suggestion
You will need to specify the `api` option in the relationship if the destination resource and/or the join table are parts of a different API:
```

zachdaniel:

PR #801, opened 2023-12-12
Embedded resource changesets should have access to the tenant, actor, and authorize? values

rgraff:
I'm not sure this is an ideal testing strategy but it reproduces the issue.  

In short, I'm using embedded resources as arguments and attributes on resources. When the validations and changes run for the embedded resources, I need to be able to access the tenant, actor and authorize? values to query the data layer.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Fixed in `main`.

Issue #634, opened 2023-06-28
`%Ash.NotSelected{}`


zachdaniel:
Currently when a field is not selected, we fill it in with a `nil` value. This is unideal and uncommunicative. We should instead fill them in with `%Ash.NotSelected{}`.

barnabasJ:
Maybe something like `%Ash.NotProvided` would also be nice when using embedded types as input arguments and not all values were provided. Right now, it is impossible to distinguish between a field that should be set to null and a value that was not provided.

ghalestrilo:
I'm looking into this one - there are multiple levels in the callback stack where this `nil` could be changed. Is [`Ash.Actions.Read:1471`](https://github.com/ash-project/ash/blob/d09c6941327082ca42c569ee45d9fe7781d80cf8/lib/ash/actions/read.ex#L1471) the best place to hook this logic or does it belong somewhere else?

zachdaniel:
I think the most concise place to put it would be here:
https://github.com/ash-project/ash/blob/d09c6941327082ca42c569ee45d9fe7781d80cf8/lib/ash/actions/helpers.ex#L656

We're going to need to flag this feature behind 3.0, as there will certainly be some places to adjust this logic and various downstream effects that need to be handled.

zachdaniel:
This has been merged into the 3.0 branch.

PR #804, opened 2023-12-14
Adds %Ash.NotSelected{} to ash 3 (!634)

ghalestrilo:
Introduces an [`%Ash.NotSelected{}`](https://github.com/ash-project/ash/issues/634) struct to ash >3.0.0, replacing `nil` in unselected fields during a read operation.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Nice! What we should also do is update the GitHub CI file to run tests against ash_three as well, so we can see what breaks (if anything).

lib/ash/actions/helpers.ex
```diff
@@ -653,7 +654,14 @@ defmodule Ash.Actions.Helpers do
       end
     end)
     |> Enum.reduce(result, fn key, record ->
-      Map.put(record, key, nil)
+      default_field_value =
+        if Ash.Flags.ash_three?() do
+          %Ash.NotSelected{}
```
zachdaniel, line 659: This should include the field that is not selected.

ghalestrilo:
I noticed that the ash three tests are missing in CI!  But they'll require a dedicated build step as well 😬 
Does it makes more sense to be a separate MR?

zachdaniel:
Yeah, I think a separate PR is good :)

zachdaniel (approved):

PR #807, opened 2023-12-19
fix: use source if attribute of upsert field has one

barnabasJ:
# Contributor checklist

Without this, the generated SQL in AshPostgres used the wrong column name for the upsert part. 

I can add a test to AshPostgres later, but a test in this repo probably doesn't work because the ETS DataLayer used in tests doesn't support the source option, if I'm not mistaken. 

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
`ETS` doesn't support the source option, correct. But ultimately I think this has to be fixed in `ash_postgres`, not here. Ash core always deals in attributes, never the source of the attribute. Also this sounds like a bug in ecto core to me, which is frustrating to say the least.

PR #813, opened 2023-12-26
Adds a test to ensure authorization occurs when the actor is set and …

hanrelan:
It seems that `authorize?: true` isn't being set when the actor is set, even `authorization` is `:by_default`.

I wasn't sure how to fix the issue, but this adds a (sort of janky) failing test for that case.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ x ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/authorizer/authorizer_test.exs
+++ b/test/authorizer/authorizer_test.exs
@@ -17,6 +17,14 @@ defmodule Ash.Test.Changeset.AuthorizerTest do
 
     actions do
       defaults [:create, :read, :update, :destroy]
+
+      create :title_is_authorization do
+        accept []
+
+        change fn changeset, context ->
+          Ash.Changeset.change_attribute(changeset, :title, context.authorize?)
+        end
+      end
     end
 
     attributes do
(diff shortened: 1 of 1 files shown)


zachdaniel:

zachdaniel:
This is....actually kind of a conundrum/problem. The basic issue here is that we don't *know* what the api is yet.

```elixir
Post
|> Ash.Changeset.for_create(:title_is_authorization, %{}, actor: :an_actor)
```

Nowhere here have we been given an api, so we don't know if we're authorizing yet. This is actually a pretty significant issue, because the general pattern is this:

```elixir
def change(changeset, _opts, context) do
  Ash.Changeset.before_action(changeset, fn changeset -> 
    # do something with context
    context.authorize?
  end)
end
```

so we close over the initial value and there is no good way to get the proper value (technically, inside the before action hook you could do `changeset.context.private.authorize?` but thats of course not a good idea.

I'm going to have to think about this. I fear that any fix to tighten this up is going to require a non-trivial breaking change.

zachdaniel:
Yeah, unfortunately, a true fix for this is going to have to wait until some redesigns that will have to happen with 3.0. In the meantime, the only short term solution we have is to *remove* the `authorize?` key from the context unless it was explicitly provided.

zachdaniel:
Nope, can't do that, as it is in itself a breaking change (just tried it, had to revert). For now, I've added an `api` option to `Ash.Changeset.for_*` that will allow you to pass the api, which should produce the correct values there in all cases. This will be fixed without the need to do that in the 3.0 release.

zachdaniel:
Okay, so there is a way that you can get around this using some upcoming 3.0 work that is currently in `main`, although not documented. You'd be using a small part that is unlikely to change in any meaningful way before 3.0 comes out.

In each resource, do the following:

```elixir
use Ash.Resource, ..., api: TheApiTheResourceIsIn
```

hanrelan:
Got it - makes sense. Will set it up to pass the Api to the resource everywhere (or potentially just use `authorize?: true`, though I'm not sure if that doesn't help for extensions like ash_graphql )

zachdaniel:
You can use the `use Ash.Resource, ..., api: Api` currently (if you are using `main`) I will release the changes tomorrow. That will allow you to avoid passing the api to all changesets.

PR #821, opened 2024-01-07
fix: only put keywords to be interpolated in message in error vars

### LINKED ISSUE #820: InvalidAttribute `vars` keyword list should not include `:field`, `:message` or `:path`
If my understanding is correct, the `vars` attribute of `Ash.Error.Changes.InvalidAttribute` is supposed to include values that need to be interpolated in the error message string. However, extra values are also being included.

In the following example, `min: 3` should be present in `vars`, however `field` and `message` are superfluous. I noticed this as the values are leaking through to my frontend via my `ash_graphql` API.

Please could you confirm that this is a bug, then I will submit a PR to fix? Thanks!

```elixir
%Ash.Error.Changes.InvalidAttribute{
  field: :string_e,
  message: "length must be greater than or equal to %{min}",
  private_vars: nil,
  value: "xy",
  changeset: nil,
  query: nil,
  error_context: [],
  vars: [
    field: :string_e,
    message: "length must be greater than or equal to %{min}",
    min: 3
  ],
  path: [],
  stacktrace: #Stacktrace<>,
  class: :invalid
}
```

ahey:
This PR resolves #820 – I think I found and fixed this in all places it was occurring.

During the course of this work, I [discovered](https://github.com/ash-project/ash/compare/main...ahey:error-vars?expand=1#diff-32cc05145e05358ce83d202848ab062c4748ca83784a0f6a7e1854f5aac02594R151) and fixed another issue ([here](https://github.com/ash-project/ash/compare/main...ahey:error-vars?expand=1#diff-5c6d681025e312c3caa3d2600b733fa250caa0878be36aa94a6f49b5aeeaa078R157) and [here](https://github.com/ash-project/ash/compare/main...ahey:error-vars?expand=1#diff-9b75dd399fae543dfa8cd09e97e8e6c11a969c91aa1f7d83b85631ff0cc52670R178)) where `vars` was missing a keyword to be interpolated into the error string.

A few tests were incorrectly expecting the superflouous keywords to be included in vars. These have been updated, and the suite is now passing, as well as our product's test suite.

### DIFF
--- a/lib/ash/error/exception.ex
+++ b/lib/ash/error/exception.ex
@@ -55,7 +55,11 @@ defmodule Ash.Error.Exception do
             }
           end)
 
-        super(opts)
+        super(opts) |> Map.update(:vars, [], &clean_vars/1)
+      end
+
+      defp clean_vars(vars) do
+        vars |> Keyword.drop([:field, :message, :path])
       end
 
       defoverridable exception: 1, message: 1
--- a/lib/ash/error/exception.ex
+++ b/lib/ash/error/exception.ex
@@ -59,7 +59,7 @@ defmodule Ash.Error.Exception do
       end
 
       defp clean_vars(vars) do
-        vars |> Keyword.drop([:field, :message, :path])
+        vars |> Kernel.||([]) |> Keyword.drop([:field, :message, :path])
       end
 
       defoverridable exception: 1, message: 1
--- a/lib/ash/type/keyword.ex
+++ b/lib/ash/type/keyword.ex
@@ -175,7 +175,7 @@ defmodule Ash.Type.Keyword do
             {:ok, [{field, field_value} | result]}
 
           {:error, errors} ->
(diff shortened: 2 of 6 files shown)


ahey:
Thanks @zachdaniel, that is a better approach. I've updated my branch in line with this.

lib/ash/error/exception.ex
```diff
@@ -55,7 +55,11 @@ defmodule Ash.Error.Exception do
             }
           end)
 
-        super(opts)
+        super(opts) |> Map.update(:vars, [], &clean_vars/1)
+      end
+
+      defp clean_vars(vars) do
+        vars |> Keyword.drop([:field, :message, :path])
```
zachdaniel, line 62:
```suggestion
        vars |> Kernel.||([]) |> Keyword.drop([:field, :message, :path])
```

zachdaniel:

lib/ash/error/exception.ex
ahey: Piping into `||`, TIL 😄

PR #828, opened 2024-01-11
Add code interface support in Info and Builder.

jechol:
# Contributor checklist

- [x] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/builder.ex
+++ b/lib/ash/resource/builder.ex
@@ -568,4 +568,96 @@ defmodule Ash.Resource.Builder do
       Keyword.merge(opts, name: name, relationship_path: List.wrap(relationship_path))
     )
   end
+
+  @doc """
+  Builds and adds an interface unless an interface with that name already exists
+  """
+  @spec add_new_interface(
+          Spark.Dsl.Builder.input(),
+          name :: atom,
+          opts :: Keyword.t()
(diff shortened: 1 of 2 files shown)


jechol:
I tried to create an interface with the same name but different arity, but it didn't work. 
Should I really base it on arity?

```elixir
  calculations do
    calculate :full_name,
              :ci_string,
              expr(
                if corp_is_prefix,
                  do: string_join([~i"(주)", name], " "),
                  else: string_join([name, ~i"(주)"], " ")
              )

    calculate :full_name_with_joiner,
              :ci_string,
              expr(
                if corp_is_prefix,
                  do: string_join([~i"(주)", name], ^arg(:joiner)),
                  else: string_join([name, ~i"(주)"], ^arg(:joiner))
              ) do
      argument :joiner, :string, allow_nil?: false
    end
  end

  code_interface do
    define_calculation :full_name, calculation: :full_name, args: []
    define_calculation :full_name, calculation: :full_name_with_joiner, args: [:joiner]
  end
```

zachdaniel:
🤔 hmm....yeah that might be an issue...lets do what you're doing for now :) We can always add one that also takes an arity later.

zachdaniel:

PR #833, opened 2024-01-12
feat: Allow `:self` as an option when requesting a relative page from a given page

sevenseacat:
This will re-run the same query that was used to fetch the given page, updating the data in it

I was hoping there would be some tests for the existing options that I could copy and modify, but there were not :( I manually tested it with both offset and keyset pagination, and they both behaved as expected.

### Use case:

When viewing a page of data in a web browser, and then deleting a record on it, reloading the page (data page, not web page) will correctly shift the data up/down to fill the gap.


zachdaniel:
I think we have relevant pagination tests here: https://github.com/ash-project/ash/blob/main/test/actions/pagination_test.exs

sevenseacat:
Ah hah. I'll add some for this then :)

zachdaniel:

PR #839, opened 2024-01-22
fix: do not check primary key fields in filter

barnabasJ:
I had a test in my project error out because the pagination added the `id` as a filter `id > last_entry_from_last_page_id` and primary keys do not have field policies if I'm not mistaken.

### DIFF
--- a/lib/ash/policy/authorizer/authorizer.ex
+++ b/lib/ash/policy/authorizer/authorizer.ex
@@ -540,7 +540,6 @@ defmodule Ash.Policy.Authorizer do
       _ ->
         altered =
           sort
-          # primary key doesn't have policies on it, and so is nil here
           |> Enum.with_index()
           |> Enum.reduce(
             {[], authorizer_acc(authorizer, authorizer.resource, context)},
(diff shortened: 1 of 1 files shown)


lib/ash/policy/authorizer/authorizer.ex
```diff
@@ -775,38 +776,43 @@ defmodule Ash.Policy.Authorizer do
...
-
-        {:filter_and_continue, filter, _authorizer} ->
-          raise """
-          Was given a partial filter for a field policy for field #{inspect(field)}.
-
-          Filter: #{inspect(filter)}
-
-          Field policies must currently use only filter checks or simple checks.
-          """
+      if is_nil(policies) do
```
zachdaniel, line 779: I think we should do something like check if the field is a primary key, to be more defensive instead. That way if there is a bug at some point we don't accidentally not apply field policies. Its a better failure case.

barnabasJ: That sounds good. Let me see if I can figure it out.

lib/ash/policy/authorizer/authorizer.ex
```diff
@@ -830,12 +835,14 @@ defmodule Ash.Policy.Authorizer do
           Ash.Changeset.accessing(changeset, [:attributes, :calculations, :aggregates])
       end
 
+    pkey = Ash.Resource.Info.primary_key(query_or_changeset.resource)
+
     accessing_fields
+    # primary keys are always accessible
+    |> Enum.reject(&(&1 in pkey))
```
barnabasJ, line 842: As this was similar logic, I tried to make the same change here, but this led to tests failing. Before, if there were no field_policies at all, it would allow access, and now it behaves similarly to actions where you need to have a policy. 

How do you feel about this? Maybe something for 3.0?

barnabasJ: Also not sure how this plays together with the sort/filter when there are no policies at all.

zachdaniel: 🤔 its a good question. We can consider it for 3.0, but will need to keep the current behavior until then.

zachdaniel: We should potentially have a top level check here for "is there at least one field policy" to make this decision explicit.

zachdaniel (approved):

zachdaniel:

Issue #842, opened 2024-01-23
Actor missing in Api.can call


barnabasJ:
The opts are passed here in place of the actor https://github.com/ash-project/ash/blob/d229ade9fcca9543386a4126b3bdc6145a1231fe/lib/ash/actions/read/relationships.ex#L170

PR #843, opened 2024-01-23
fix: pass in actor to api.can call

barnabasJ:
Addresses #842

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -165,7 +165,9 @@ defmodule Ash.Actions.Read.Relationships do
             |> Map.put(:api, join_relationship.api || related_query.api)
             |> hydrate_refs(relationship.source)
 
-          case through_query.api.can(through_query,
+          case through_query.api.can(
+                 through_query,
+                 source_query.context[:private][:actor],
                  alter_source?: true,
                  run_queries?: false,
                  base_query: through_query
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -165,35 +165,50 @@ defmodule Ash.Actions.Read.Relationships do
             |> Map.put(:api, join_relationship.api || related_query.api)
             |> hydrate_refs(relationship.source)
 
-          case through_query.api.can(
-                 through_query,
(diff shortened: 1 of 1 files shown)


barnabasJ:
Still need to handle the `{:ok, false}`  and `{:ok, false, exception}` case in some way

zachdaniel:
Ah, right. I think adding the error to the main query should do it. I believe there is an option to tell it to return the exception so that we never get `{:ok, false}`, it will always have the exception.

barnabasJ:
The test in my project is still failing, I'm doing a query without authorization. I think if `authorize?` is `false` we maybe need to skip the call to `Api.can`?

zachdaniel:
Yep, you're totally right.

lib/ash/actions/read/relationships.ex
```diff
@@ -165,33 +165,50 @@ defmodule Ash.Actions.Read.Relationships do
...
-            {:ok, true, authorized_through_query} ->
-              {:ok,
-               [
-                 {clear_lateral_join_source(source_query), relationship.source_attribute,
-                  relationship.source_attribute_on_join_resource, relationship},
-                 {clear_lateral_join_source(authorized_through_query),
-                  relationship.destination_attribute_on_join_resource,
-                  relationship.destination_attribute, join_relationship}
-               ]}
+          if source_query.context[:private][:authorize?] do
```
barnabasJ, line 168: It feels like I'm adding a lot of `if` statements recently, but I was unable to come up with something better in the short term.

Do you have some rule of thumb for things like this? Would you refactor some things out into smaller functions to circumvent this or is there another way?

zachdaniel: Yeah, if there is a need to refactor I'd focus on extracting something out into a named function. Often I don't end up doing this, much to the chagrin of others. Feel free to do so if you like :)

barnabasJ: I'm ok with merging it like this for now, I still have a couple other tests that fail with the new read action code. That has priority for me right now.

zachdaniel:

PR #844, opened 2024-01-23
fix: make sure query has action

barnabasJ:
Without the action name, I ran into an exception in the policy logger:

https://github.com/ash-project/ash/blob/189c6659c17baca46ab95a8e0fad6d88756849e8/lib/ash/policy/authorizer/authorizer.ex#L1182

### DIFF
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -155,6 +155,14 @@ defmodule Ash.Actions.Read.Relationships do
 
           through_query =
             relationship.through
+            |> Ash.Query.for_read(
+              relationship.read_action ||
+                Ash.Resource.Info.primary_action!(relationship.through, :read).name,
+              %{},
+              authorize?: source_query.context[:private][:authorize?],
+              actor: source_query.context[:private][:actor],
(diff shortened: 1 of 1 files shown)


lib/ash/actions/read/relationships.ex
```diff
@@ -155,6 +155,14 @@ defmodule Ash.Actions.Read.Relationships do
 
           through_query =
             relationship.through
+            |> Ash.Query.for_read(
+              relationship.read_action ||
```
zachdaniel, line 159: This needs to use the `through` relationship read action.

zachdaniel: Sorry, the `join_relationship`

lib/ash/actions/read/relationships.ex
```diff
@@ -156,7 +156,7 @@ defmodule Ash.Actions.Read.Relationships do
           through_query =
             relationship.through
             |> Ash.Query.for_read(
-              relationship.read_action ||
+              relationship.through.read_action ||
```
zachdaniel, line 159: `through` is a module, needs to be `join_relationship.read_action`

zachdaniel (approved):

PR #846, opened 2024-01-23
fix: `join_relationship` in many to many can make `through` optional

vonagam:
Now if existing `join_relationship` is provided in `many_to_many` then it takes priority, overrides `through` (making it optional) and `destination_attribute_on_join_resource`.

It is still possible to specify non-matching `through` and `destination_attribute_on_join_resource` but at least now the relationship will use actual ones in the end. I can add errors for such cases but a new transformer will be needed for that since it needs to be run before `ManyToManySourceAttributeOnJoinResource`, I tried to reuse this one as was suggested.


zachdaniel (approved):

zachdaniel:

zachdaniel:
FYI I had to undo part of this. We can derive the `through`, but not the `destination_attribute_on_join_resource`. What we can derive is the `source_attribute_on_join_resource`.

vonagam:
Yeah, hah, in discord I was writing it properly but mixed it up here.

PR #847, opened 2024-01-23
fix: load fails because of empty list instead of query

barnabasJ:
This gets rid of the exeption in my test. The behaviour is still a bit different as the calculation is now returning back `nil` instead of `[]` as before

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/calculations.ex
+++ b/lib/ash/actions/read/calculations.ex
@@ -317,6 +317,15 @@ defmodule Ash.Actions.Read.Calculations do
       load_through_attribute_rewrites(ash_query, calculation, path)
   end
 
+  defp get_rewrites([], calculation, path, :relationships) do
+    dbg()
+
+    []
+    |> Enum.flat_map(fn {name, query} ->
+      get_all_rewrites(query, calculation, path ++ [name])
+    end)
+  end
+
   defp get_rewrites(ash_query, calculation, path, :relationships) do
     ash_query.load
     |> Enum.flat_map(fn {name, query} ->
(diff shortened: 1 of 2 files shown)


zachdaniel:
Interesting...is `[]` the default value of the calculation?

barnabasJ:
> Interesting...is `[]` the default value of the calculation?

I didn't specify a default value on the resource, the return type for the calculation is `{:array, :struct}`

These might be two different things tough

zachdaniel:
Hmm....it seems like a bug then. Like that your calculation value isn't properly being added.

zachdaniel:
The new calculation system is much less complicated than the old, but still complicated. We set conflicting calculation dependencies to a special name that tells us to rewrite them later when evaluating a given calculation. It seems like an error in that rewriting logic.

barnabasJ:
I have to stop now for today. I will try to find why the return value changed tomorrow.

zachdaniel:
I'm also happy to address this, ideally if a failing test can be written.

barnabasJ:
> I'm also happy to address this, ideally if a failing test can be written.

So, the nil vs [] issue was actually on our end; I returned a Stream instead of a list from the calculation, which led to the wrong result. 

So this should be good to go, I think.

barnabasJ:
> > I'm also happy to address this, ideally if a failing test can be written.
> 
> So, the nil vs [] issue was actually on our end; I returned a Stream instead of a list from the calculation, which led to the wrong result.
> 
> So this should be good to go, I think.

We could potentially change the guard here https://github.com/ash-project/ash/blob/a3b8efd6479a3d40bfc615971107f0059ff792d7/lib/ash/actions/read/calculations.ex#L80 to include `or is_struct(values, Stream)` if you'd like to support returning streams. In my case, the `is_map(value)` guard of the other implementation matched.

Might be a better experience for the user. The current way kinda masks the error and makes it hard to find?

zachdaniel:
I'll merge this, would you be able to PR that change?

barnabasJ:
> I'll merge this, would you be able to PR that change?

sure, I'll do it tomorrow

PR #852, opened 2024-01-25
add test for selecting policy

barnabasJ:
This test should mirror the behaviour I tested in my project before the refactor. 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,7 +2,6 @@ import Config
 
 config :ash,
   flags: [
-    read_uses_flow?: System.get_env("FLAG_READ_USES_FLOW", "false") == "true",
     ash_three?: System.get_env("FLAG_ASH_THREE", "false") == "true"
   ]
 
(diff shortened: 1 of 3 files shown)


barnabasJ:
I ran this test with ash from the commit in my project before updating and it didn't do what I was expecting. Maybe this only works with AshPostgres DataLayer.

I'm checking further.

barnabasJ:
In the end the problem was a different one, I also had this condition in my policy

forbid_if(selecting(:internal_notes))

but the field was deselected here https://github.com/ash-project/ash/blob/f84d85cf20148f90afad9e1e44d4a2a115340257/lib/ash/actions/read/calculations.ex#L1026 before the query was authorized

barnabasJ:
I used the `selecting` policy before field policies were available, but never changed it after introducing them into my project.

zachdaniel:

PR #855, opened 2024-01-26
fix: use correct option for relationship in atomic `relate_actor`

vonagam:
Fixes:
- `opts[:field]` was used where `opts[:relationship]` is supposed to.
- actor was not get by `opts[:field]`, added `actor_or_field` call to use field if specified.

Note: I see that there is a check for case of actor being nil. But this check does not cover a case where `opts[:field]` is specified. In that case an `actor` can be an empty map, so actual actor is still nil and it slips past the check.

### DIFF
--- a/lib/ash/resource/change/relate_actor.ex
+++ b/lib/ash/resource/change/relate_actor.ex
@@ -14,69 +14,81 @@ defmodule Ash.Resource.Change.RelateActor do
     end
   end
 
-  def change(changeset, opts, %{actor: nil}) do
-    validate_type!(changeset, opts)
-
-    if opts[:allow_nil?] do
-      changeset
-    else
-      Changeset.add_error(
-        changeset,
-        InvalidRelationship.exception(
-          relationship: opts[:relationship],
-          message: "could not relate to actor, as no actor was found (and :allow_nil? is false)"
-        )
-      )
-    end
-  end
-
   def change(changeset, opts, %{actor: actor}) do
-    validate_type!(changeset, opts)
-    field = opts[:field]
+    relationship = resolve_relationship!(changeset, opts)
+    actor = resolve_actor(actor, opts)
+    allow_nil = opts[:allow_nil?]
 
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:
Yeah, we should probably be checking that the specific field value is not `nil`, as opposed to just the actor. Would you mind doing that in this PR as well?

vonagam:
Ok, will add it. Just initially it was self-explanatory one line change, now I will need to do more than that.

vonagam:
On that topic, I do not see a nil check in atomic version. I assume it just missing for a moment, but need to be noted that it needs to be added.

zachdaniel:
Yes, you are right. The atomic version should be able to use exactly the same logic as the non-atomic one since actor values are always known up front, even though the check itself is atomic.

vonagam:
Okay, rewrote things to handle nil in a field and for atomic to use the same logic as non-atomic change.

Two things:

First, there was a test that explicitly allowed passing actor with a field of nil even without `allow_nil?` set to true. Had to change it. It still means that the change is a little breakish.

Second, I don't know how to properly handle cases for atomic when no changes needed (allow_nil - I wrote it as `{:atomic, %{}}` and case when there is an error (I just marked it as non-atomic with error message, I hope it means that non-atomic version will be called and error will be created properly).

lib/ash/resource/change/relate_actor.ex
```diff
@@ -14,69 +14,81 @@ defmodule Ash.Resource.Change.RelateActor do
...
-
-    if relationship.type == :belongs_to do
-      {:atomic,
-       %{
-         relationship.source_attribute => Map.get(actor, relationship.destination_attribute)
-       }}
-    else
-      {:not_atomic, "Can only use `relate_actor` atomically with a belongs_to relationship."}
+      nil ->
+        {:not_atomic,
```
zachdaniel, line 55:
```suggestion
        {:error,
```

PR #861, opened 2024-01-31
fix: union merge load logic

barnabasJ:
before it was adding the acc as the load for a union type that had no loads.


zachdaniel:
Wow, nice find! There is a slightly better fix here though, which is not to return `{:ok, acc}` on empty load above. I'll make that change.

PR #863, opened 2024-01-31
fix: change names of code interface methods for calculations

### LINKED ISSUE #857: Calculation names that end with a "?" cannot be called with the "!" variant.
**Describe the bug**
Defining a calculation that ends in `?` in the code interface of a resource causes the bang version to fail. E.g: `Resource.is_true?(data)` will return `{:ok, true}`, but `Resource.is_true?!(data)` will raise a function clause error.

**To Reproduce**
```elixir
defmodule Thing.Resource do
  use Ash.Resource

  attributes do
    uuid_primary_key :id
  end

  calculations do
    calculate :is_true?, :boolean, expr(true) 
  end

  code_interface do
    define_for Thing.Api
    define_calculation :is_true?, args: [:_record]
  end
end
```

**Expected behavior**
```
iex(6)> Thing.Resource.is_true?!(%Thing.Resource{})
true
```
**Actual behavior**
```
iex(6)> Thing.Resource.is_true?!(%Thing.Resource{})
** (FunctionClauseError) no function clause matching in Thing.Resource."-inlined-is_true?/2-"/2    
    
    The following arguments were given to Thing.Resource."-inlined-is_true?/2-"/2:
    
        # 1
        false
    
        # 2
        []
    
    (ash_calculation_bug_reproduced 0.1.0) Thing.Resource."-inlined-is_true?/2-"/2
    iex:6: (file)
```

**Runtime**
 - Elixir version
   - 1.15.4
 - Erlang version
   - 26.0.2
 - OS
   - Mac OS Sonoma 14.1.1
 - Ash version
   - 2.14
   - N/A

vonagam:
Closes #857.

As proposed in [the last comment](https://github.com/ash-project/ash/issues/857#issuecomment-1917978189) now for calculations with names that end with the question mark the following convention applies:
```elixir
is_true(...) # safe method
is_true?(...) # throwing method
``` 

This is a breaking change since it changes behavior of theoretical `is_safe?` from safe one to throwing one.

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -183,6 +183,20 @@ defmodule Ash.CodeInterface do
     end
   end
 
+  @doc false
+  # sobelow_skip ["DOS.BinToAtom"]
+  def resolve_calc_method_names(name) do
+    if name |> to_string() |> String.ends_with?("?") do
+      safe_name = name |> to_string() |> String.trim_trailing("?") |> String.to_atom()
+      bang_name = name
+      {safe_name, bang_name}
+    else
+      safe_name = name
+      bang_name = "#{name}!" |> String.to_atom()
+      {safe_name, bang_name}
+    end
+  end
+
   @doc """
   Defines the code interface for a given resource + api combination in the current module. For example:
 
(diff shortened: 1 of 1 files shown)


zachdaniel:
🔥

zachdaniel:

PR #866, opened 2024-02-02
improvement: add `join_keys` option to `manage_relationship`

vonagam:
Idea is to allow this for many to many:
```elixir
[
  on_match: :update,
  on_lookup: :relate,
  on_no_match: :create,
  join_keys: [:some, :key],
]
```
With the same meaning as this:
```elixir
[
  on_match: {:update, :update, :update, [:some, :key]},
  on_lookup: {:relate, :create, :read, [:some, :key]},
  on_no_match: {:create, :create, :create, [:some, :key]},
]
```

Additional things:

- Cleaned up `ManagedRelationshipHelpers.sanitize_opts` - made variable names and white space usage consistent. Added a missing case for `on_no_match: {:create, create, join_create}` and removed one redundant case in `on_lookup`. In  `:on_lookup` `primary_action_name` method was used to get an action name unlike other ones which were using `Ash.Resource.Info.primary_action!`. The difference is that the latter is throwing and former is not. For consistency I converted all calls to throwing ones. Except for update action in `on_lookup: :relate` for `belongs_to` relationship. Seems like that combo does not actually need an update action. Or does it?

- Added clause for `defp split_join_keys(input, [])` - I assume it is most common one so makes sense to simplify.

- Removed mention of `[:join, :keys]` from `on_missing: :destroy` full form documentation. It does not support it.

### DIFF
--- a/lib/ash/actions/managed_relationships.ex
+++ b/lib/ash/actions/managed_relationships.ex
@@ -1441,6 +1441,10 @@ defmodule Ash.Actions.ManagedRelationships do
     {metadata[:join_keys] || %{}, input}
   end
 
+  defp split_join_keys(input, []) do
+    {%{}, input}
+  end
+
   defp split_join_keys(input, :all) do
     {input, %{}}
   end
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -3277,7 +3277,7 @@ defmodule Ash.Changeset do
       * `:ignore` (default) - those inputs are ignored
       * `:destroy` - the record is destroyed using the destination's primary destroy action
       * `{:destroy, :action_name}` - the record is destroyed using the specified action on the destination resource
-      * `{:destroy, :action_name, :join_resource_action_name, [:join, :keys]}` - the record is destroyed using the specified action on the destination resource,
+      * `{:destroy, :action_name, :join_resource_action_name}` - the record is destroyed using the specified action on the destination resource,
         but first the join resource is destroyed with its specified action
       * `:error`  - an error is returned indicating that a record would have been updated
       * `:unrelate` - the related item is not destroyed, but the data is "unrelated", making this behave like `remove_from_relationship/3`. The action should be:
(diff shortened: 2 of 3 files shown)


zachdaniel (approved):
nice :)

vonagam:
So, two things:

- It does not matter but just want to mention that maybe `join_params` is better than `join_keys`. I used `join_keys` to be consistent with internal code that called them that, but `join_keys` might sound like keys you join on. It can be changed if you think it is better.

- To repeat the question about `on_lookup: :relate` for `belongs_to`. The documentation states that in the form of `{:relate, update_action}` and `{:relate, update_action, read_action}` `update_action` means "an update action on the source resource" for `belongs_to` (same in `:relate_and_update`). But if I understand code properly that's incorrect. Belongs to relationship does not seem to use any update action to change a relationship attribute. If that so, does it not makes sense for the long form of relate instruction to be in order of `{:relate, read_action, update_action, join_keys}`? Because all need `:relate`, all can accept read action, some can accept update/create action and even less so can accept join params. I assume it is a breaking change but maybe Ash 3 release is a good time to make such change (and it's not that big one).

zachdaniel:
Yeah, you're right, for `on_lookup: :relate` for a `:belongs_to` actually just adds a change to the current action.

As for option order, I think we need to switch to a nested configuration, and I'd like to DSL-ify it at some point.

```elixir
manages_relationship :arg_name, :relationship_name do
  type :create
  on_missing :destroy do
    read_action :read_action ....
  end
end
```

vonagam:
> if there is an update action configured we should error if its not the update action that is currently being called

Hm... but lookup can be called not only in an update action, but create too. Actually in tests there are two cases ([1](https://github.com/ash-project/ash/blob/e331e24c7becf9bb7eace83d65cf09c07f741c81/test/resource/changes/relate_actor_test.exs#L18), [2](https://github.com/ash-project/ash/blob/e331e24c7becf9bb7eace83d65cf09c07f741c81/test/support/flow/child.ex#L17)) with `manage_relationship` in a create action to set `belongs_to` without any primary update action.

And it seems as unneeded verbosity if to configure a read action you need to repeat a current action name (or specify a `nil`?):
```elixir
create :create_with_account do
  argument :account, :map
  change manage_relationship(:account, on_lookup: {:relate, :create_with_account, :custom_read})
end
```

Ideally belongs to should support only instructions `:relate` and `{:relate, read_action}`. It is possible to add action type check for second element in a tuple and error if it is not a read action.

zachdaniel:
Yeah, I see your point. You're right.

PR #868, opened 2024-02-06
fix: set right `accessing_from` name in many to many managed create

vonagam:
Right now when managed create for many to many relationship is happening the name in `accessing_from` is for a through relationship. 

Example: 

Resources `Post`, `Tag` and `PostTag`. In `Tag` we add such policy:
```elixir
policies do
  policy action_type(:create) do
    authorize_if accessing_from(Post, :tags)
  end
end
```
This will not work because actual name in `accessing_from` context will be `tags_join_assoc` (if the through name is generated automatically).

### DIFF
--- a/lib/ash/actions/managed_relationships.ex
+++ b/lib/ash/actions/managed_relationships.ex
@@ -1132,7 +1132,7 @@ defmodule Ash.Actions.ManagedRelationships do
             relationship.destination
             |> Ash.Changeset.new()
             |> Ash.Changeset.set_context(%{
-              accessing_from: %{source: relationship.source, name: relationship.join_relationship}
+              accessing_from: %{source: relationship.source, name: relationship.name}
             })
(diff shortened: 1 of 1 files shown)


zachdaniel:
🤔 this is interesting. We actually use `accessing_from` for more than just authorization cases, so I don't think we can make this change, unfortunately. It is technically correct that it is being accessed from the join relationship. We can potentially enrich the custom check to be able to figure this out, but I don't think that we can make this suggested change.

vonagam:
But in update it is [set correctly](https://github.com/ash-project/ash/blob/f7024f2b6ce90dc2f298da0f698fe32d5da1bab2/lib/ash/actions/managed_relationships.ex#L1312).

That's the only place where the `name` is incorrect.

zachdaniel:
Oh, sorry. I misunderstood. I see now. You're right, we're using the join relationship in a case where we shouldn't. Good catch!

zachdaniel:

PR #870, opened 2024-02-06
Update validate_related_resource_inclusion.ex

BryanJBryce:
The error message had an error because it referred to a option name that doesn't exist

# Contributor checklist

- [ ] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

PR #871, opened 2024-02-06
chore: continue update validate_related_resource_inclusion.ex

vonagam:
Continues  #870.

Saw the PR - looked if `join_relationship_name` is mentioned anywhere else and the exact same thing was in another place.


zachdaniel:

PR #873, opened 2024-02-07
fix: continue cleanup in `ManagedRelationshipHelpers`

vonagam:
This continues cleanup in `ManagedRelationshipHelpers`:

- Now `:unrelate` is properly resolved in `sanitize_opts`. It removes the need to do it in `ManagedRelationships` and allows to use resolved actions in `on_match_destination_actions` and `on_missing_destination_actions`. Those changes detected a test where many to many used `:append_and_remove` but no destroy action was present on a join resource.

- Added documentation for `:missing` in `:on_match`.

- In helper methods use the fact that `sanitize_opts` is called at the start. So no need to handle all possible unsanitized variants again.

- Do not resolve update action for `:relate` in `:on_lookup` in belongs_to.

- Remove non-throwing `primary_action_name`. Remove unneeded `source`. Remove unneeded `unwrap` calls.

### DIFF
--- a/lib/ash/actions/managed_relationships.ex
+++ b/lib/ash/actions/managed_relationships.ex
@@ -1630,9 +1630,6 @@ defmodule Ash.Actions.ManagedRelationships do
        ) do
     tenant = changeset.tenant
 
-    action_name =
-      action_name || Ash.Resource.Info.primary_action(relationship.through, :destroy).name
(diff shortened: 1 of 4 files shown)


zachdaniel (approved):

zachdaniel:

PR #875, opened 2024-02-07
improvement: add `on_match: :update_join` to `manage_relationship`

vonagam:
Support for updating only join resource in `manage_relationship` without updating destination one in many to many.

Can be specified as `:update_join`, `{:update_join, :action_name}` or `{:update_join, :action_name, [:join, :param]}`.

Under the hood it just gets transformed into `{:update, nil, :action_name, [:join, :param]}`. So the PR adds such alias and checks if an update action for destination is nil or not before doing it.

I will need to add it to documentation and ideally a test. But first want to confirm that such implementation is acceptable. Because it adds support for `{:update, nil}` (for many to many only, as an alternative way to do the same thing) but that's the least disrupting way to add the feature. The code is simple.

Note: saw [a wrong line](https://github.com/ash-project/ash/blob/d47eb086a254a4ada965ae1d90f0be4abc857deb/lib/ash/actions/managed_relationships.ex#L1301) and added a fix (at least in a way that I assume it is intended) for it here.

### DIFF
--- a/lib/ash/actions/managed_relationships.ex
+++ b/lib/ash/actions/managed_relationships.ex
@@ -1298,7 +1298,7 @@ defmodule Ash.Actions.ManagedRelationships do
               {regular_params, %{}}
 
             is_struct(regular_params) ->
-              Map.from_struct(regular_params)
+              {match, Map.from_struct(regular_params)}
 
             true ->
               {match, regular_params}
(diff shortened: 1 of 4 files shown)


zachdaniel (dismissed):
LGTM! Docs and a test and it's good to go. Awesome work as always.

vonagam:
Ok added a test and documentation mention. Let me know if anything is not quite right.

zachdaniel (approved):

PR #886, opened 2024-02-13
docs: Correct example used for manual actions

sevenseacat:
In my testing, arguments to a manual action are part of the `arguments` in the input to the `run` function - which is the action struct itself.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/dsls/DSL:-Ash.Resource.md
+++ b/documentation/dsls/DSL:-Ash.Resource.md
@@ -776,7 +776,7 @@ For calling this action, see the `Ash.Api` documentation.
 action :top_user_emails, {:array, :string} do
   argument :limit, :integer, default: 10, allow_nil?: false
   run fn input, context ->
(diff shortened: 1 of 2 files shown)


documentation/dsls/DSL:-Ash.Resource.md
```diff
@@ -775,8 +775,8 @@ For calling this action, see the `Ash.Api` documentation.
 ```
 action :top_user_emails, {:array, :string} do
   argument :limit, :integer, default: 10, allow_nil?: false
-  run fn input, context ->
-    with {:ok, top_users} <- top_users(input.limit) do
+  run fn action, context ->
```
zachdaniel, line 778: We would still call this input, something like input.arguments.foo

zachdaniel:

PR #888, opened 2024-02-13
Fix/replace refs for relationships

barnabasJ:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/policy/authorizer/authorizer.ex
+++ b/lib/ash/policy/authorizer/authorizer.ex
@@ -715,18 +715,18 @@ defmodule Ash.Policy.Authorizer do
   defp do_replace_ref(
          %{
            attribute: %struct{name: name},
-           relationship_path: relationship_path,
+           relationship_path: [relationship_path],
            resource: resource
          } = ref,
-         %{stack: [{_, _path, action} | _]} = acc
+         %{stack: [{parent, _path, _action} | _]} = acc
        )
-       when struct in [Ash.Resource.Attribute, Ash.Resource.Aggregate, Ash.Resource.Calculation] and
-              not is_nil(relationship_path) do
+       when struct in [Ash.Resource.Attribute, Ash.Resource.Aggregate, Ash.Resource.Calculation] do
+    action =
+      Ash.Resource.Info.relationship(parent, relationship_path).read_action ||
+        Ash.Resource.Info.primary_action!(resource, :read)
+
     {expr, acc} =
-      expression_for_ref(resource, name, action, ref, %{
-        acc
-        | stack: [{resource, relationship_path, action} | acc.stack]
-      })
+      expression_for_ref(resource, name, action, ref, acc)
 
     {expr, acc}
   end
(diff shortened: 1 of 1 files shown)


lib/ash/policy/authorizer/authorizer.ex
```diff
@@ -712,6 +712,25 @@ defmodule Ash.Policy.Authorizer do
...
+         } = ref,
+         %{stack: [{parent, _path, _action} | _]} = acc
+       )
+       when struct in [Ash.Resource.Attribute, Ash.Resource.Aggregate, Ash.Resource.Calculation] do
+    action =
+      Ash.Resource.Info.relationship(parent, relationship_path).read_action ||
+        Ash.Resource.Info.primary_action!(resource, :read)
+
+    {expr, acc} =
+      expression_for_ref(resource, name, action, ref, acc)
```
barnabasJ, line 729: We might also need to update the stack here. But doing it caused an error for me. 

I hit this case when coming from the `%{__function__?: true, ... }` above. Returning the original stack there might be the correct way to do things. What do you think?

zachdaniel: 🤔 I'm pretty sure the way you have it here is the right way :)

zachdaniel:
This looks good, *but* I don't understand exactly how we got to the scenario, a failing test would be great. If you have time. If not, we can merge.

zachdaniel (dismissed):

zachdaniel:
I think it should be fine because `expression_for_ref` doesn't use the `stack` IIRC.

barnabasJ:
I tried adding a test. I did it in AshPostgres because I needed a filter function e.g `ilke` to be available. But I'm still having a hard time triggering the problem. Right now it does not call do_replace_refs because the `input?` flag on the Ref is not true. How do I need to set the filter to make it so?

barnabasJ:
> I tried adding a test. I did it in AshPostgres because I needed a filter function e.g `ilke` to be available. But I'm still having a hard time triggering the problem. Right now it does not call do_replace_refs because the `input?` flag on the Ref is not true. How do I need to set the filter to make it so?

I was able to figure it out by looking at the ash_graphql resolver, still having issues replicating the error though

barnabasJ:
I added this test in ash_postgres https://github.com/ash-project/ash_postgres/pull/206

The behaviour is a bit strange over all, but this way, you should be able to get a better picture of it.

barnabasJ:
> I added this test in ash_postgres [ash-project/ash_postgres#206](https://github.com/ash-project/ash_postgres/pull/206)
> 
> The behaviour is a bit strange over all, but this way, you should be able to get a better picture of it.

I added another case that is two relationships deep. This still fails even with my current change.

lib/ash/policy/authorizer/authorizer.ex
```diff
@@ -712,6 +712,25 @@ defmodule Ash.Policy.Authorizer do
     end
   end
 
+  defp do_replace_ref(
+         %{
+           attribute: %struct{name: name},
+           relationship_path: [relationship_path],
```
zachdaniel, line 718:
```suggestion
           relationship_path: relationship_path,
```

zachdaniel:

This looked like some hard work to narrow this done. Fixed one small issue in your PR but otherwise this is good to go!

PR #893, opened 2024-02-14
chore: format some `with` clauses to be on a single line

vonagam:
Seems formatter does not enforce a single style in that case and allows to write both:
```elixir
with a <- b do
end
```
and:
```elixir
with a <-
       b do
end
```

Saw a second variant where first one makes more sense (especially if it does not match neighbors). But I assume that some of changed lines might be a deliberate stylistic choice though.


zachdaniel (approved):

zachdaniel:

PR #900, opened 2024-02-16
Reproducing a bug where the flow does not honor tenant

jechol:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Flow doesn't seem to honor tenant.
This PR is to reproduce the bug.

```
mix test test/flow/tenant_test.exs
```


jechol:
@zachdaniel Could you confirm if this is a bug or my misunderstanding?

zachdaniel:
This does indeed look like a bug. Will merge and debug.

zachdaniel:
Okay, so this isn't actually a bug, just a weird quirk of how destroy/update steps work in `Ash.Flow`. If the record being updated or destroyed evaluates to `nil`, the update/destroy step itself returns `nil`. So it is actually functioning as expected. I pushed up a fix just now for something else that is related. What you can do is specify `not_found_error? true` on a read step with `get? true` if you want the record not being present to be an error.

zachdaniel:
Actually, nevermind, `not_found_error?` defaults to `true`. So the bug was that we weren't honoring that behavior.

PR #901, opened 2024-02-16
Reproduce flow custom step has no tenant in context

jechol:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

When run
```
mix test test/flow/tenant_test.exs
```

From `GetPost.run/3`
```
opts = context |> Ash.context_to_opts() |> IO.inspect()
```
outputs
```
[authorize?: false, actor: nil]
```
and there's no tenant.

### DIFF
--- a/test/flow/tenant_test.exs
+++ b/test/flow/tenant_test.exs
@@ -51,6 +51,21 @@ defmodule Ash.Flow.TenantTest do
     end
   end
 
+  defmodule GetPost do
+    use Ash.Flow.Step
+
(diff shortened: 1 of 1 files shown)


jechol:
@zachdaniel Could you confirm if this is a bug or my misunderstanding?

zachdaniel:
I'm not sure if it's a "bug" per-se, but I think we ought to send the tenant in as context. I'll add it.

zachdaniel:
fixed in `2.19.1`

PR #907, opened 2024-02-19
fix: correct `for_read` calls in update and destroy actions

vonagam:
`for_read` opts were provided as a third argument, meaning they were treated as an input arguments instead and ignored (unless somebody had arguments with same names).

In non-bulk destroy there was a propagation of context from a changeset to a query through `Ash.Query.set_context(changeset.context)` but in other three places there wasn't. Meaning that important context like table for polymorphic relationships was missing.

Reordered some lines to be in the same order.


zachdaniel:

PR #910, opened 2024-02-22
docs: Improvements to the bulk_update and bulk_destroy functions

simpers:
Improvements to the bulk_update and bulk_destroy functions of the Ash.Api module.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/api/api.ex
+++ b/lib/ash/api/api.ex
@@ -610,7 +610,7 @@ defmodule Ash.Api do
                               resource: [
                                 type: {:spark, Ash.Resource},
                                 doc:
-                                  "The resource being updated. This must be provided if the input given is a stream, so we know ahead of time what the resource being updated is."
+                                  "The resource being destroyed. This must be provided if the input given is a stream, so we know ahead of time what the resource being updated is."
                               ],
                               stream_batch_size: [
                                 type: :integer,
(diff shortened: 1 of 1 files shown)


lib/ash/api/api.ex
```diff
@@ -1958,14 +1958,16 @@ defmodule Ash.Api do
...
               input :: map,
-              Keyword.t()
+              opts :: Keyword.t()
             ) ::
               Ash.BulkResult.t() | no_return
 
   @doc """
-  Updates all items in the provided enumerable or query with the provided input.
+  Updates all items in the provided enumerable or query with the provided input for the action, which
+  is a map, where the keys are the primary keys of the records to update, and the values are maps
```
zachdaniel, line 1969: The keys to the input map aren't the primary keys of the records to update. That comes from the enumerable or query in the first argument.

lib/ash/api/api.ex
```diff
@@ -1989,14 +1991,16 @@ defmodule Ash.Api do
...
               input :: map,
-              Keyword.t()
+              opts :: Keyword.t()
             ) ::
               Ash.BulkResult.t() | no_return
 
   @doc """
-  Destroys all items in the provided enumerable or query with the provided input.
+  Destroys all items in the provided enumerable or query with the provided input for the action, which
+  is a map, where the keys are the primary keys of the records to update, and the values are maps
```
zachdaniel, line 2002: same here.

lib/ash/api/api.ex
simpers: Okay, great! What does it mean "that comes"? Like, what could it look like? Will fix when I understand it better.

zachdaniel: What I mean is that with `bulk_update`, you don't provide primary keys, you provide a query or stream to be updated. For example:

```elixir
Resource
|> Ash.Query.filter(active == true)
|> MyApi.bulk_update(:deactivate)
```

or

```elixir
list_of_records = [%Resource{}, %Resource{}, %Resource{}]

MyApi.bulk_update(list_of_records, :deactivate)
```

Primary keys don't come into the equation in bulk updates or destroys.

simpers: Okay, so should the action need input, the input map will be the same input for all records in the list/stream?

zachdaniel: Correct.

simpers: Gotcha! Fixing :)

zachdaniel (approved):

PR #916, opened 2024-02-25
fix: prevent leakage of stale async_limiter's pid 

tw4452852:
the pid of async_limiter in the query's context maybe invalid due to we stop it at the end of read. So adding another checking to make sure it's still alive.

BTW, I found this issue according to the following crash:

```
[error] GenServer #PID<0.647.0> terminating
** (stop) exited in: GenServer.call(#PID<0.648.0>, {:get_and_update, #Function<0.21809672/1 in Ash.Actions.Read.AsyncLimiter.async_or_inline/3>}, 5000)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir 1.16.0) lib/gen_server.ex:1114: GenServer.call/3
    (ash 2.19.3) lib/ash/actions/read/async_limiter.ex:26: Ash.Actions.Read.AsyncLimiter.async_or_inline/3
    (elixir 1.16.0) lib/enum.ex:1708: anonymous fn/3 in Enum.map/2
    (elixir 1.16.0) lib/enum.ex:4399: anonymous fn/3 in Enum.map/2
    (elixir 1.16.0) lib/stream.ex:1816: anonymous fn/3 in Enumerable.Stream.reduce/3
    (elixir 1.16.0) lib/enum.ex:4842: Enumerable.List.reduce/3
    (elixir 1.16.0) lib/stream.ex:1828: Enumerable.Stream.do_each/4
    (elixir 1.16.0) lib/enum.ex:4399: Enum.map/2
    (ash 2.19.3) lib/ash/actions/read/relationships.ex:39: Ash.Actions.Read.Relationships.fetch_related_records/2
    (ash 2.19.3) lib/ash/actions/read/relationships.ex:23: Ash.Actions.Read.Relationships.load/3
    (ash 2.19.3) lib/ash/actions/read/read.ex:237: Ash.Actions.Read.do_run/3
    (ash 2.19.3) lib/ash/actions/read/read.ex:49: anonymous fn/3 in Ash.Actions.Read.run/3
    (ash 2.19.3) lib/ash/actions/read/read.ex:48: Ash.Actions.Read.run/3
    (ash 2.19.3) lib/ash/api/api.ex:2301: Ash.Api.page/3
    (ash 2.19.3) lib/ash/api/api.ex:2258: Ash.Api.page!/3
...
```

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -256,6 +256,11 @@ defmodule Ash.Actions.Read do
                opts[:authorize?],
                false
              ) do
+        query =
+          query
+          # prevent leakage of stale pid as we stop it at the end of reading
+          |> clearup_async_limiter()
+
         data
         |> Helpers.restrict_field_access(query)
         |> add_tenant(query)
@@ -417,6 +422,15 @@ defmodule Ash.Actions.Read do
       end
     end)
   end
+
+  defp clearup_async_limiter(
+         %{context: %{private: %{async_limiter: async_limiter}} = ctx} = query
+       )
+       when is_pid(async_limiter) do
+    Ash.Query.set_context(query, put_in(ctx[:private][:async_limiter], nil))
+  end
+
+  defp clearup_async_limiter(query), do: query
 
   def add_async_limiter(
         %{context: %{private: %{async_limiter: async_limiter}}} = query,


zachdaniel:
This is interesting...It should not need to be restarted, because it shouldn't ever be reused after a given query. Can you tell me more about what you were doing when you got this error?

tw4452852:
Normally yes, but it's not the case during the pagination due to we cache the query in it. And this is what trigger this bug, I do something like this:

```
cur_page = API.read!(..., page: [limit: ...])
...
cur_page = API.page!(cur_page, :next) # this will tigger the exception
```

BTW, the reason why we didn't catch this bug during testing is due to [the condition of using `async_limiter`](https://github.com/ash-project/ash/blob/f9af1307a1141055712e122ccb82d0c1e9e51ec2/lib/ash/actions/read/read.ex#L450-L451) and we usually set `config :ash, :disable_async?, true` as recommended [in the documentation](https://ash-hq.org/docs/guides/ash/latest/topics/testing#testing).

zachdaniel:
Ah..... okay, so what we need to do is make sure that the query we return has the async limiter *removed* from its context. Would you mind adjusting this PR to do that? In `read.exs` we have a function called `add_page` and `add_query`, and the `query` needs to be adjusted in that block to be the same query but with the async limiter removed from its context.

tw4452852:
I thought about this way and I also want to consolidate the stopping the process and clearing context in the same place. But it's a bit of tricky as we close the process [in the after block](https://github.com/ash-project/ash/blob/f9af1307a1141055712e122ccb82d0c1e9e51ec2/lib/ash/actions/read/read.ex#L280) while the value is returned.

tw4452852:
@zachdaniel Got it. I've updated the changes, hope you like it.

zachdaniel (approved):

PR #920, opened 2024-03-04
improvement: Make undo actions for `update` and `destroy` actions more explicit.

jimsynz:


lib/ash/reactor/steps/destroy_step.ex
```diff
@@ -58,18 +60,12 @@ defmodule Ash.Reactor.DestroyStep do
...
 
-    attributes =
-      options[:resource]
-      |> Ash.Resource.Info.attributes()
-      |> Map.new(&{&1.name, record[&1]})
-      |> then(&Map.merge(arguments[:input], &1))
-
     options[:resource]
-    |> Changeset.for_create(options[:undo_action], attributes, changeset_options)
+    |> Changeset.for_create(options[:undo_action], %{record: record}, changeset_options)
```
zachdaniel, line 64: we should also set the changeset metadata on this record right?

zachdaniel (approved):
had one comment, but otherwise LGTM

lib/ash/reactor/steps/destroy_step.ex
jimsynz: Should it? Does a changeset for `destroy` contain anything useful?  The only reason we do it on the `update` step is because it's the only way to pass the changeset to the `undo` function if we need it later.

zachdaniel: A destroy action can be "soft" in which case it is an update masquerading as a destroy.

PR #925, opened 2024-03-11
chore: remove unneeded `persist` call

vonagam:
There is no need to do `Spark.Dsl.Transformer.persist(dsl, :api, api)` since `api` comes from `Spark.Dsl.Transformer.get_persisted(dsl, :api)`.


jimsynz (approved):
details matter

PR #935, opened 2024-03-20
fix: update set/get/update_context function specs to use maps

skanderm:
`Ash.PlugHelpers` context function specs were mistakenly using `Ash.Resource.record()` types instead of `map()`


zachdaniel (approved):

zachdaniel:

PR #937, opened 2024-03-21
improvement: add `matches` built-in policy check

vonagam:
Adds `matches` built-in check for making simple checks using functions without need to define a separate module for that.

Example from the test (where `bar` is required argument):
```elixir
policy action(:create_bar) do
  authorize_if matches("bar is even", fn _actor, context ->
                 rem(context.changeset.arguments.bar, 2) == 0
               end)
end
```

Also included some typespec corrections - unless I'm mistaken `actor` is not required to be a struct (even though it is most likely is) and also can be nil.

### DIFF
--- a/lib/ash/policy/check.ex
+++ b/lib/ash/policy/check.ex
@@ -9,6 +9,7 @@ defmodule Ash.Policy.Check do
   for an easy way to write that check.
   """
 
+  @type actor :: any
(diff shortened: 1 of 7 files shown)


zachdaniel (approved):

zachdaniel:

PR #942, opened 2024-03-22
fix: corrections for `reload`/`reload!`

vonagam:
- `reload!` was not exposed on `Ash`, only `reload` was.
- `reload` was marked as not having `opts` while it does.


zachdaniel:

PR #945, opened 2024-03-23
fix: show that `Policy.Check.strict_check` can return an error tuple

vonagam:
Type that `strict_check` is [allowed to return](https://github.com/ash-project/ash/blob/0afb9ade26ba540e61fa9db27f4f5bb77bce1071/lib/ash/policy/policy.ex#L190) `{:error, error}` tuple.


zachdaniel:

PR #946, opened 2024-03-24
improvement: support `load` option for create/update/destroy

vonagam:
`load` option was supported in read and bulk actions. Now it is applicable to usual create/update/destroy actions too.

Noting that create/update/soft-destroy and non-soft destroy handle loads different.

Small changes besides that - consistent usage of `Helpers` alias, remove double call to `add_process_context` in destroy action, remove empty `@create_update_opts_schema`.


zachdaniel (approved):

zachdaniel:

PR #951, opened 2024-03-27
fix: fix missing tenant in some bulk contexts

vonagam:
There were two identical contexts on the same level in nesting. Both were missing tenants. Then in some PR definition for one of contexts was moved up and added tenant. So by removing second context a tenant also gets passed properly there.


zachdaniel:

PR #956, opened 2024-03-28
improvement: allow simple check to return error tuple

vonagam:
Allows simple check to return `{:error, error}` or explicit `{:ok, boolean}` in addition to simply returning a boolean.

Had to silence dialyzer because it is used in macro, so when it is used it knows what `match?` specifically returns in a module and complains about never matching pattern.

This is PR to main. Ash 2 lives on a separate branch now. If this is merged in, what do I need to do to add it to second version too (the code and change is the same in this case)?


zachdaniel:
I wouldn't worry about it. Not everything needs to be ported back to 2.0.

PR #957, opened 2024-03-29
refactor: erroneous else

mhanberg:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

Hi!

In my Elixir parser CI, I noticed ash was failing to parse (problem for me!) but it lead me to notice you had this odd (yet valid!) code.

I'm not entirely sure if this is supposed to be "else []" or just drop the else, but figured I'd drop you a PR.

Thanks!

### DIFF
--- a/lib/ash/actions/helpers.ex
+++ b/lib/ash/actions/helpers.ex
@@ -772,6 +772,8 @@ defmodule Ash.Actions.Helpers do
   def select(results, query) do
     if Enumerable.impl_for(results) do
       Enum.map(results, &select(&1, query))
+    else
+       results
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/ash/actions/helpers.ex
```diff
@@ -772,7 +772,6 @@ defmodule Ash.Actions.Helpers do
   def select(results, query) do
     if Enumerable.impl_for(results) do
       Enum.map(results, &select(&1, query))
```
zachdaniel, line 774:
```suggestion
      Enum.map(results, &select(&1, query))
    else
       results
```

zachdaniel:
It's interesting because I thought you weren't supposed to have empty blocks like that. Like in a case statement for example. Feels like I should have gotten a compiler warning of some kind about this 😆 

PR #963, opened 2024-03-31
fix: handle `sensitive?` option in query aggregate/calculation

vonagam:
In `Ash.Query.Aggregate` and `Ash.Query.Calculation` options included `sensitive?` but it was not used (also one had wrong default value, other had no default value). Now it is handled.

Also in calculation there was `sortable?: opts[:filterable?]`.


zachdaniel:

PR #964, opened 2024-04-01
improvement: support `allow_nil_input` dsl option in update/destroy

vonagam:
Create/update/destroy support `accept` and `require_attributes` but only create supports `allow_nil_input` (strange name though, why "input" instead of "attributes"?). With it comes support for `allow_nil_input` in update/destroy actions.

Also simplified (and, I assume, corrected some issues in) private method `Changeset.attributes_to_require`.


zachdaniel:
We shouldn't need `allow_nil_input` on any other action type. The reason it's called `allow_nil_input` is because, although the *input* is allowed to be `nil`, we don't ever allow the attribute itself to actually end up as `nil`. It basically says "my action will set the value, if you don't give it to me". But for updates, `allow_nil?` is already accounted for on the existing record, so there is no need for `allow_nil_input?`.

vonagam:
So let's say we have an attribute name "foo" and it does not allow nulls. Do I understand correctly that update action should allow to pass in `foo: nil` with `accept [:foo]` (with the intention to handle it somehow in a change before data layer) without any `allow_nil_input`? (Because I think it did not work for me like that, need to double check then if true.)

zachdaniel:
Hmm...yeah, I guess you're right, there is a subtle difference. The reason that `allow_nil_input` was added for create actions was actually to handle when it was *missing*, not explicitly set to `nil`. We could support `allow_nil_input` on update/destroy in that case.

So the last thing is that I'm not sure of is the change from `:create` to `:update`. I think that was actually intentional (although looks weird), because `:update` only validates new changes. Let's leave it until we know there is a bug for now. Otherwise, good to go 👍

vonagam:
Ok, removed change from `:update` to `:create`.

PR #965, opened 2024-04-01
improvement: support `allow_nil_input` dsl option in update/destroy (Ash 2 backport)

vonagam:
Backport of #964 for Ash 2.

The change is almost the same except two things. First, take into account `reject` (which does not exist in Ash 3). Second, do remove an attribute from being required because it has `private?: true` - seems like a bug, because `private?` is not about that.

Edit: Put back `private?` check. Just to avoid any breaking changes.


zachdaniel (approved):

zachdaniel:
Yeah, the `private?` bit there is actually used because we basically do two phases. We check public attributes that are accepted by the action in an earlier step, since the user can't provide them, we know they must be set by the action.

PR #976, opened 2024-04-04
feat: add `update_change` function and builtin change

rbino:
Allows mapping a change with a function

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -4138,6 +4138,41 @@ defmodule Ash.Changeset do
   end
 
   @doc """
+  Updates an existing attribute change by applying a function to it.
+
+  This is useful for applying some kind of normalization to the attribute.
+
+  ```elixir
+  Ash.Changeset.update_change(changeset, :serial, &String.downcase/1)
+  ```
+
+  The update function gets called with the value already cast to the correct type.
+
+  ```elixir
+  changeset
+  |> Ash.Changeset.change_attribute(:integer_attribute, "3")
+  |> Ash.Changeset.update_change(:integer_attribute, fn x -> x + 1 end)
+  ```
+
+  ## Invalid value handling
+
+  If `update_change` is called with a changeset that has not been validated yet, the update
+  function must handle potentially invalid and `nil` values.
+
+  To only deal with valid values, you can call `update_change` in a `before_action` hook.
+  """
+  @spec update_change(t(), atom, (any -> any)) :: t()
+  def update_change(changeset, attribute, fun) do
+    case Ash.Changeset.fetch_change(changeset, attribute) do
+      {:ok, change} ->
+        Ash.Changeset.force_change_attribute(changeset, attribute, fun.(change))
+
+      :error ->
+        changeset
+    end
+  end
+
+  @doc """
   Add an argument to the changeset, which will be provided to the action.
   """
   def set_argument(changeset, argument, value) do
--- a/lib/ash/resource/change/builtins.ex
+++ b/lib/ash/resource/change/builtins.ex
@@ -132,6 +132,23 @@ defmodule Ash.Resource.Change.Builtins do
   end
 
   @doc """
+  Updates an existing attribute change by applying a function to it.
+
+  The update function gets called with the value already cast to the correct type, and only gets called
+  on valid changesets, so the value is guaranteed to have passed validations and constraints.
+  """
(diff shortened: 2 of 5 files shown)


lib/ash/changeset/changeset.ex
```diff
@@ -4137,6 +4137,51 @@ defmodule Ash.Changeset do
...
+
+  If the attribute has `allow_nil?: false`, the update function is guaranteed not to be called if the
+  attribute value is `nil`, even if it is set to `nil` explicitly.
+
+  If the attribute has `allow_nil?: true` the update function must consider `nil` as a possible input
+  if it is passed explicitly.
+  """
+  @spec update_change(t(), atom, (any -> any)) :: t()
+  def update_change(changeset, attribute, fun) do
+    case Ash.Changeset.fetch_change(changeset, attribute) do
```
zachdaniel, line 4167: I think we should not call the function if the changeset is invalid as well. That way if a validation is applied on a value that doesn't pass, you don't try to update it. For instance:

```elixir
accept [:...]
validate compare(points, greater_than: 0)
change update_change(:points, &(&1 - 1))
```

rbino: This doesn't work (e.g. for `allow_nil` validations) if it gets called like this:

```elixir
    test "does not call the update function on explicit `nil` attribute with `allow_nil?: false`" do
      Tag
      |> Ash.Changeset.new()
      |> Ash.Changeset.change_attribute(:name, nil)
      |> Ash.Changeset.update_change(:name, fn _ -> raise "I should not be called" end)
      |> Ash.Changeset.for_create(:create)
    end
```

this actually raises (and `:name` has `allow_nil?: false`)

It works moving the call to `update_change` downwards, but then the implementation has to use `force_change_attribute`. I'm not sure if the function should be designed to be used before or after `Changeset.for_*`, or both.

zachdaniel: Yeah, I think if we do everything you’re doing *and* check for validity that would do what we want. It will work best-effort before the action, and fully during the action.

rbino: I've tried to do that, but it's still brittle (for example, action validations run before `update_change` if they're before it _but_ global validations run after, so the update function can be called with a value that fails a global validation).

Given this not so obvious pieces of behavior, I think it's best to drop the function in the `Changeset` module and just provide a builtin that gets called in a `before_action` hook, like it's already done in, e.g., [Increment](https://github.com/ash-project/ash/blob/main/lib/ash/resource/change/increment.ex)

zachdaniel: global validations should happen after action-specific changes.

zachdaniel: I think we can do it in a before action hook 👍 only issue is that it could throw users for a loop expecting it to be updated.

rbino:
Updated with the changes discussed on Discord, cc @vonagam

lib/ash/resource/change/builtins.ex
```diff
@@ -131,6 +131,18 @@ defmodule Ash.Resource.Change.Builtins do
...
+  @doc """
+  Updates an existing attribute change by applying a function to it.
+
+  The update function gets called with the value already cast to the correct type, and only gets called
+  on valid changesets, so the value is guaranteed to have passed validations and constraints.
+  """
+  @spec update_change(attribute :: atom, update_function :: (any -> any)) ::
+          Ash.Resource.Change.ref()
+  def update_change(attribute, function) do
+    {Ash.Resource.Change.UpdateChange, attribute: attribute, function: function}
```
vonagam, line 143: Shouldn't `Spark.CodeHelpers.lift_functions` be used here like in other built-ins that accept functions?

rbino: `set_argument` and `set_new_argument` accept functions but do not lift them, `after_action` etc do lift them.
I guess it would make sense to lift them here so the caller can define the function inline, which helps keep the relevant code in a single place

zachdaniel: Yeah, I think so 👍

rbino: Done

vonagam (approved):

zachdaniel:

PR #979, opened 2024-04-04
improvement: make Ash.Resource.Validation.Match partially atomic

rbino:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/validation/match.ex
+++ b/lib/ash/resource/validation/match.ex
@@ -56,6 +56,34 @@ defmodule Ash.Resource.Validation.Match do
   end
 
   @impl true
+  def atomic(changeset, opts, context) do
+    cond do
+      argument?(changeset.action, opts[:attribute]) ->
+        validate(changeset, opts, context)
+
+      not Ash.Changeset.changing_attribute?(changeset, opts[:attribute]) ->
+        {:not_atomic, "can't atomically match an attribute that is not changing"}
+
+      atomic_expr_change?(changeset.atomics, opts[:attribute]) ->
+        {:not_atomic, "can't match on an atomic expression"}
+
+      true ->
+        validate(changeset, opts, context)
+    end
+  end
+
+  defp argument?(%{arguments: arguments}, attribute) do
+    Enum.find(arguments, &(&1.name == attribute))
+  end
+
+  defp atomic_expr_change?(atomics, attribute) do
+    case Keyword.get(atomics, attribute) do
+      nil -> false
+      value -> Ash.Expr.expr?(value)
+    end
+  end
+
+  @impl true
   def describe(opts) do
     [
       message: opts[:message],
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1913,7 +1913,16 @@ defmodule Ash.Changeset do
         end
 
       %{always_atomic?: true, validation: _} = change, changeset ->
-        run_atomic_validation(changeset, change, context)
+        case run_atomic_validation(changeset, change, context) do
+          {:not_atomic, reason} ->
+            Ash.Changeset.add_error(
+              changeset,
+              "Validation #{change.module} must be run atomically, but it could not be: #{reason}"
+            )
+
+          changeset ->
+            changeset
+        end
 
       %{change: {module, opts}, where: where} = change, changeset ->
         if module.has_change?() do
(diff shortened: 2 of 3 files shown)


zachdaniel:
In order to be atomic, a validation has to be able to work even on atomic updates (i.e in the data layer). We can only support `match` being atomic if its talking about arguments, so we'd need to do something like check if its an argument, or check if its being changed to a known value.

rbino:
If I understand correctly the problem is that the validation will check if the attribute matches the regex even if it's not changing, and if it's not changing the match should be done in the data layer, where there's currently no support for a regex match operation.

So `atomic` should check:
- If the validation is on an argument _or_ it's on an attribute that is currently changing to a known value, return `{:atomic, validate(changeset, opts, context)}`
- Otherwise, return `{:not_atomic, "some error"}`

Is this right?

zachdaniel:
Yep!

zachdaniel:
You can use `Ash.Expr.expr?` to see if a value is a literal or an expression?

rbino:
@zachdaniel I've updated the PR as requested, but the tests are failing in Ash internals, with this stacktrace (for both tests that should return `:not_atomic`):

```
     ** (KeyError) key :action not found in: {:not_atomic, "can't atomically match an attribute that is not changing"}

     If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
     code: |> Ash.Changeset.for_update(:atomic_update, %{other: "foo"})
     stacktrace:
       (ash 3.0.0-rc.12) lib/ash/resource/validation/match.ex:61: Ash.Resource.Validation.Match.atomic/3
       (ash 3.0.0-rc.12) lib/ash/changeset/changeset.ex:647: Ash.Changeset.run_atomic_validation/3
       (elixir 1.16.2) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
       (ash 3.0.0-rc.12) lib/ash/changeset/changeset.ex:1449: Ash.Changeset.do_for_action/4
       test/resource/validation/match.ex:91: (test)
```

Maybe the case of a partially atomic validation is not currently handled?

rbino:
Ok found and fixed the issue

lib/ash/resource/validation/match.ex
```diff
@@ -55,6 +55,34 @@ defmodule Ash.Resource.Validation.Match do
     end
   end
 
+  @impl true
+  def atomic(changeset, opts, context) do
+    cond do
+      argument?(changeset.action, opts[:attribute]) ->
+        validate(changeset, opts, context)
+
+      not Ash.Changeset.changing_attribute?(changeset, opts[:attribute]) ->
```
zachdaniel, line 64: So, we can do the match in an expression as long as it's not a regex. What it would look like is something like this:

```elixir
argument?(...) ->
  validate(...)

!is_expr(new_value) -> # get new value from attributes[attr] || atomics[attr]
  do_match_logic_inline(new_value)
  
is_regex(opts[:match]) ->
   {:not_atomic, "can't do regexes atomically"}
 
is_expr(new_value) -> # get new value from attributes[attr] || atomics[attr]
  {:atomic, ...., expr(contains(^atomic_ref(attr), ^opts[:match]))}
```

this kind of thing. I need to write a guide on how atomics work, thats part of what I'll be doing soon.

rbino: I'm not understanding how `:match` can be anything other than a Regex given it gets passed as second argument to `String.match?/2`, which only accepts a regex ad second argument 🤔

Another thing I've noticed is that Match returns `:ok` if the attribute [is not changing](https://github.com/ash-project/ash/blob/2a585662adbbb11c85125f01927312a69917142c/lib/ash/resource/validation/match.ex#L53), so I guess also the "not changing" case is atomic-capable

zachdaniel: That is wrong actually 😞it needs to be applied even if not changing.

rbino: Should this be fixed in this PR or in a separate one?

rbino: I think this is wrong in other validations too (e.g. OneOf does the right thing in `atomic` but only validates changes and arguments in `validate`), would it make sense to fix those in a separate PR?
This PR should be good otherwise (since it already does the right thing if the attribute is not changing, which is declaring the validation not atomic)

zachdaniel: Yeah, we definitely need to fix any other validations 😢 They need to check current data if things aren't changing (unless the validation is explicitly checking for changes). Also yeah you're right I don't know why I thought `String.match` accepted a string and would check for it being contained 😆 that was dumb.

PR #988, opened 2024-04-07
fix: remove `internal?` and `stacktraces?` from common opts

vonagam:
Unused.


zachdaniel:

PR #989, opened 2024-04-07
fix: undo adding primary key argument for methods with `get?`

vonagam:
In Ash 3 code interface started to automatically add an argument for primary key to methods with `get?: true`. 

Seems like it was unintended consequence of refactoring `if` to `case` statement.

This PR undoes this change. To get things by id one can use `get_by: :id`.


zachdaniel (approved):

zachdaniel:

PR #993, opened 2024-04-08
improvement: Allow Reactor modules to be run directly from generic actions

jimsynz:
- Extends the generic action runner to detect when the `run` module is a Reactor and automatically executes it in the correct way.
- Adds a verifier that detects said actions and ensures that action arguments are present for all expected inputs.

# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel (approved):
SICK 🔥

PR #997, opened 2024-04-09
fix: Validate all conditions in a `numericality` validation instead of only the last

sevenseacat:
The previous `reduce` loop did not break on the first failure and return the error, it returned the last iteration of the loop

❗ ❗ Some more work may need to be done on the atomic side of the validation, I'm not familiar with how they work!

Without the code change, the new test fails:

```
  1) test validate against a range of values (Ash.Test.Resource.Validation.CompareTest)
     test/resource/validation/compare_test.exs:139
     match (=) failed
     code:  assert {:error, %{message: message, vars: vars}} = Compare.validate(changeset, opts, %{})
     left:  {:error, %{message: message, vars: vars}}
     right: :ok
     stacktrace:
       test/resource/validation/compare_test.exs:157: Ash.Test.Resource.Validation.CompareTest.assert_error/3
       test/resource/validation/compare_test.exs:145: (test)
```

Because -1 passes the last check (less than or equal to 10), the validation passed when it should not have.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/validation/compare.ex
+++ b/lib/ash/resource/validation/compare.ex
@@ -34,41 +34,30 @@ defmodule Ash.Resource.Validation.Compare do
 
     case value do
       {:ok, value} ->
-        Enum.reduce(
-          Keyword.take(opts, [
-            :greater_than,
-            :less_than,
-            :greater_than_or_equal_to,
-            :less_than_or_equal_to
-          ]),
-          :ok,
-          fn validation, _ ->
-            case validation do
-              {:greater_than, attribute} ->
-                if Comp.greater_than?(value, attribute_value(changeset, attribute)),
-                  do: :ok,
-                  else: invalid_attribute_error(opts, value)
+        opts
+        |> Keyword.take([
+          :greater_than,
+          :less_than,
+          :greater_than_or_equal_to,
+          :less_than_or_equal_to
+        ])
+        |> Enum.find_value(fn
+          {:greater_than, attribute} ->
+            if !Comp.greater_than?(value, attribute_value(changeset, attribute)),
+              do: invalid_attribute_error(opts, value)
 
-              {:greater_than_or_equal_to, attribute} ->
-                if Comp.greater_or_equal?(value, attribute_value(changeset, attribute)),
-                  do: :ok,
-                  else: invalid_attribute_error(opts, value)
+          {:greater_than_or_equal_to, attribute} ->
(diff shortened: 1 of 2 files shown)


lib/ash/resource/validation/compare.ex
```diff
@@ -34,41 +34,30 @@ defmodule Ash.Resource.Validation.Compare do
...
-          end
-        )
+        opts
+        |> Keyword.take([
+          :greater_than,
+          :less_than,
+          :greater_than_or_equal_to,
+          :less_than_or_equal_to
+        ])
+        |> Enum.find_value(fn
```
zachdaniel, line 44: So the original point of the reduce (although not working) was to gather up multiple errors. So for this, lets turn this into a `flat_map`, where the `:ok` case returns `[]` and the error returns `[the_error]`, and then check do something like:

```elixir
|> case do
  [] -> :ok
  errors -> {:error, errors}
end
```

zachdaniel:
🤦 thanks for the fix. Had a comment but otherwise 👍

lib/ash/resource/validation/compare.ex
sevenseacat: Gotcha. Will make that change tomorrow 👍

sevenseacat: Actually, thinking about this a bit more - how could a `compare` check (in a `numericality` validation) generate more than one error on a field? All of the options seem pretty mutually exclusive, unless you have odd conditions like `[greater_than: 5, greater_than_or_equal_to: 10]` and you pass in a number that fails both checks.

The only case I can think of is if we wanted to validate that a number was _outside_ a given range, eg x < 5 || x > 10, but that would have to be implemented with negating a standard numericality validation, wouldn't it?

The `message` function also prints out the full conditions (eg. "must be greater than 1 and must be less than 100") for the error, regardless of which part failed.

zachdaniel: You could be validating a range, like greater than 1 and less than 10.

zachdaniel: But the fact that the message contains all the info does suggest that we could just add the one error.

sevenseacat: Even if you are validating a range, there's only going to be one error - you can't fail both checks at once.

zachdaniel:

PR #1003, opened 2024-04-10
fix: swap the parameters in Ash.ToTenant.to_tenant/2

rbino:
This is technically a breaking change, but without this the protocol would never work as intended since the funtion gets dispatched based on the type of the first parameter and the resource parameter will always be an atom (precisely, an alias) so the implementation for Atom will always be called

I'll add regression tests in another PR where I call the protocol in actions, I've wanted to leave this separate to highlight the need for a breaking change

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -104,7 +104,7 @@ defmodule Ash.Changeset do
         if changeset.tenant do
           concat(
             "tenant: ",
-            to_doc(Ash.ToTenant.to_tenant(changeset.resource, changeset.tenant), opts)
+            to_doc(Ash.ToTenant.to_tenant(changeset.tenant, changeset.resource), opts)
           )
         else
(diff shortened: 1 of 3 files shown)


rbino:
I've realized that we _could_ work around that in a non-breaking way if we define a new function (e.g. `from_record`, or some other name, naming things is hard etc etc) with the parameters in the correct order and make the Atom implementation of `to_tenant` call that with swapped argument while printing some kind of deprecation warning.

Should we go down this road to be extra careful?

rbino:
Actually, I guess that this can be considered a _non_ breaking change, since no one can be relying on the current behavior because the call will _never_ be dispatched to custom `defimpl`s in the current status

zachdaniel (approved):

PR #1012, opened 2024-04-11
Properly set action_type when preparing changesets

ryanrborn:
In trying to implement bulk actions, we encountered an issue where the `action_type` on a `bulk_delete` was being set as `:update`. We have an after action that depends on the `action_type` being correct. This MR attempts to ensure that the `action_type` is set correctly.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1436,7 +1436,7 @@ defmodule Ash.Changeset do
   def prepare_changeset_for_action(changeset, action, opts) do
     changeset
     |> Map.put(:action, action)
-    |> Map.put(:action_type, action.type || :update)
+    |> Map.put(:action_type, action.type)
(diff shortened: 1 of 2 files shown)


lib/ash/changeset/changeset.ex
```diff
@@ -1436,6 +1436,7 @@ defmodule Ash.Changeset do
   def prepare_changeset_for_action(changeset, action, opts) do
     changeset
     |> Map.put(:action, action)
+    |> Map.put(:action_type, action.type || :update)
```
zachdaniel, line 1439:
```suggestion
    |> Map.put(:action_type, action.type)
```

It shouldn't ever be possible to have an action without a type

zachdaniel:

PR #1023, opened 2024-04-15
fix: union array argument/attribute handling when current attribute is nil

jeremygrant:
Seems like a few regressions have occurred when handling union arrays if the current value is nil as well as handling resetting an allow_nil? union array back to nil

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/union.ex
+++ b/lib/ash/type/union.ex
@@ -700,6 +700,7 @@ defmodule Ash.Type.Union do
        end) do
       old_values_by_type =
         old_values
+        |> List.wrap()
         |> Stream.with_index()
         |> Stream.map(fn {item, index} ->
           Map.put(item, :__index__, index)
@@ -860,6 +861,7 @@ defmodule Ash.Type.Union do
        end) do
       old_values_by_type =
         old_values
+        |> List.wrap()
         |> Stream.with_index()
         |> Stream.map(fn {item, index} ->
           Map.put(item, :__index__, index)
--- a/test/type/union_test.exs
+++ b/test/type/union_test.exs
@@ -72,7 +72,7 @@ defmodule Ash.Test.Type.UnionTest do
           new_thing =
             Ash.Changeset.get_argument(changeset, :new_thing)
 
-          things = Ash.Changeset.get_attribute(changeset, :things)
+          things = Ash.Changeset.get_attribute(changeset, :things) || []
 
           Ash.Changeset.change_attribute(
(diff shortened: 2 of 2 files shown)


zachdaniel (dismissed):

lib/ash/type/union.ex
```diff
@@ -695,11 +695,13 @@ defmodule Ash.Type.Union do
         new_values,
         constraints
       ) do
-    if Enum.any?(constraints[:types] || [], fn {_name, config} ->
-         Ash.Type.prepare_change_array?(config[:type])
-       end) do
+    if !is_nil(new_values) &&
```
zachdaniel, line 698: Hm...actually I approved this, but I'm thinking that we shouldn't call `prepare_change_array` with a `nil` value ever, and that is probably what changed here is that somehow we started doing this. I think we should solve it at the call site instead of here perhaps.

jeremygrant: @zachdaniel ok, so you're thinking we'd check here?

https://github.com/ash-project/ash/blob/49958975a18758bdc4ee74e8bdb67028cff58e68/lib/ash/type/type.ex#L593

and here for handle_change_array?

https://github.com/ash-project/ash/blob/49958975a18758bdc4ee74e8bdb67028cff58e68/lib/ash/type/type.ex#L572

zachdaniel: Actually...I think I might be changing my mind here. It makes more sense to let the underlying types handle `nil`. The problem with this is that if the value is `nil`, we still need to attempt to call the underlying `prepare_change_array` with the `nil` value. So instead of skipping, we need to skip the reindexing logic and cast each group of old values against a `nil` value for the type.

It's more complex, but also more accurate.

jeremygrant: @zachdaniel I've removed this from this PR, since as I mentioned it wasn't required for our current issue. I'll put that in a different PR.

test/type/union_test.exs
```diff
@@ -81,6 +81,11 @@ defmodule Ash.Test.Type.UnionTest do
           )
         end
       end
+
+      update :add_things do
+        require_atomic? false
```
jeremygrant, line 86: @zachdaniel the rebase from the original version required me to add this. TBH I'm not sure why 🤷‍♂️

zachdaniel:

PR #1024, opened 2024-04-15
fix: union array argument/attribute handling when current attribute is nil (Backport for 2.0)

jeremygrant:
Seems like a few regressions have occurred when handling union arrays if the current value is nil as well as handling resetting an allow_nil? union array back to nil

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/union.ex
+++ b/lib/ash/type/union.ex
@@ -665,6 +665,7 @@ defmodule Ash.Type.Union do
        end) do
       old_values_by_type =
         old_values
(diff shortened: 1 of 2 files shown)


jeremygrant:
@zachdaniel this has had the resetting an array attribute to nil case removed. So inline with the changes on PR1023

zachdaniel:

PR #1040, opened 2024-04-22
feat: leverage resource domain for eager/pre checking

### LINKED ISSUE #1022: feat: leverage resource domain for eager checking

Now that the domain is usually specified at the resource level, it is a bit redundant to specify it via `eager_check_with`.

It would be nice for Ash to automatically use the domain if it is specified at the resource level, only requiring `eager_check_with` if it is not available, and of course allowing overriding the default if desired.

We should add an `eager_check?` option which defaults to `false`. This would be a non-breaking change, and if set to `true`, it would attempt to use the resource domain, warning if it is not defined. If `eager_check_with` is set, `eager_check?` should automatically be set to true.

For example

```elixir
  identities do
    identity :foo, [:bar], eager_check?: true
  end
```

Or

```elixir
  identities do
    identity :foo, [:bar], eager_check_with: Some.Different.Domain
  end
```

frankdugan3:
closes #1022

# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/resource/transformers/set_eager_check_with.ex
```diff
@@ -0,0 +1,45 @@
...
+          Transformer.replace_entity(
+            dsl_state,
+            [:identities],
+            %{identity | eager_check_with: domain},
+            &(&1.name == identity.name)
+          )
+
+        {:ok, new_dsl_state}
+
+      %{eager_check?: false} = identity, {:ok, dsl_state} ->
```
zachdaniel, line 33: Why are we overwriting `false` with `true` here?

frankdugan3: To maintain consistency if `eager_check_with` is manually specified. It may not be necessary, but it made sense to me.

frankdugan3: The idea being that requiring a user to specify `eager_check? true` with a different domain would be a breaking change.

zachdaniel: ah, right I didn't see the `Enum.reject/2` at the top 👍

zachdaniel (approved):

PR #1053, opened 2024-04-24
Don't filter by `public?` in `Ash.Resource.Transformers.DefaultAccept`

nallwhy:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Based on the [documentation](https://github.com/ash-project/ash/blob/44c38277e5848bf0abf1b5a98b057443627bfe99/documentation/topics/security/sensitive-data.md), the `public?` option has no effect internally within Ash. However, the current implementation of `Ash.Resource.Transformer.DefaultAccept` only adds public attributes with `:*`. As `Ash.Resource.Verifiers.ValidateAccept` has been modified to no longer consider whether attributes are public, it seems appropriate to apply the same principle to `Ash.Resource.Transformer.DefaultAccept`.

### DIFF
--- a/lib/ash/resource/transformers/default_accept.ex
+++ b/lib/ash/resource/transformers/default_accept.ex
@@ -6,10 +6,10 @@ defmodule Ash.Resource.Transformers.DefaultAccept do
   alias Spark.Dsl.Transformer
 
   def transform(dsl_state) do
-    public_attribute_names =
+    attribute_names =
       dsl_state
       |> Transformer.get_entities([:attributes])
-      |> Enum.filter(&(&1.public? && &1.writable?))
+      |> Enum.filter(& &1.writable?)
       |> Enum.map(& &1.name)
 
     default_default_accept =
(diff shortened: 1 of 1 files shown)


nallwhy:
@zachdaniel 
I got it. However, this means that using `:*` requires that the attribute must be exposed externally as `public?: true`, which I think makes its use very limited. If `default_accept` is intended as a temporary feature to facilitate migration from v2 to v3, that makes sense. However, if that's not the case, I think it would be better to change it since it is being used explicitly and not implicitly.

zachdaniel:
Yeah, `default_accept :*` is primarily a backwards compatibility tool, and I think we will remove it in `4.0` in favor of `default_accept @default_accept` where you specify an explicit list.

zachdaniel:
Or just `default_accept [:foo, :bar, :baz]`

nallwhy:
@zachdaniel I agree with that! Thank you :)

Issue #1015, opened 2024-04-12
Split absent to attribute_absent and argument_absent.


jechol:
**Is your feature request related to a problem? Please describe.**
```elixir
validations do
  validate attribute_equals(:s_date, nil)
  validate absent(:s_date)
end
```

The above two have different behaviors:

* `attribute_equals`: It only looks at the attribute.
* `absent`: It first looks at the argument given in the changeset, and then it looks at the attribute.

In most cases, `attribute_equals` is the expected behavior. `absent` has an unexpected meaning that can be confusing.

**Describe the solution you'd like**
Split absent to attribute_absent and argument_absent.

**Express the feature either with a change to resource syntax, or with a change to the resource interface**

```elixir
validations do
  validate attribute_absent(:s_date)
  validate argument_absent(:s_date)
end
```

zachdaniel:
🤔 I can see how this might be confusing. I'm not sure what the best way is to eliminate the confusion, primarily because some folks intentionally leverage this behavior to do arguments.

We could potentially support some syntax like `validate absent({:attribute, :s_date})`, but that doesn't solve for the potential confusion if you have an argument w/ the same name as an attribute. The reason why splitting them up won't be a solve-all is because you can mix and match arguments and attributes `validate present([:arg1, :arg2, :attr3], at_least: 1)`.

jechol:
In that case, wouldn't it work to create three versions of every validation function?

non-prefix version: checks both argument and attribute (ex. present, equals)
argument version: checks only the argument (ex. argument_present, argument_equals)
attribute version: checks only the attribute (ex. attribute_present, attribute_equals)
Doing it this way would make the purpose clear from the function name, which seems like it would reduce the potential for confusion.

zachdaniel:
Hmmmmmmm...perhaps just having an attributes present one would be enough. The current one applies to arguments or attributes, and attributes applies only to attributes. I'd accept that PR. It should be implemented by passing an additional option into the current validation.

PR #1055, opened 2024-04-24
feat: Ash.create and Ash.update oneliner

pierrelegall:
See #1015.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -21,6 +21,10 @@ defmodule Ash do
           | Keyword.t()
           | list(atom | {atom, atom | Keyword.t()})
 
+  @type resource_with_arguments :: {Ash.Resource.t(), map() | Keyword.t()}
+
+  @type record_with_arguments :: {Ash.Resource.record(), map() | Keyword.t()}
+
   @global_opts [
     domain: [
       type: {:spark, Ash.Domain},
@@ -1953,11 +1957,22 @@ defmodule Ash do
   Create a record or raises an error. See `create/2` for more information.
   """
   @doc spark_opts: [{1, @create_opts_schema}]
-  @spec create!(Ash.Changeset.t(), Keyword.t()) ::
+  @spec create!(Ash.Changeset.t() | resource_with_arguments(), Keyword.t()) ::
           Ash.Resource.record()
           | {Ash.Resource.record(), list(Ash.Notifier.Notification.t())}
           | no_return
-  def create!(changeset, opts \\ []) do
+  def create!(changeset_or_resource_with_argument, opts \\ [])
+
+  def create!({resource, arguments}, opts) do
+    action = opts[:action] || Ash.Resource.Info.primary_action!(resource, :create).name
+    arguments = Enum.into(arguments, %{})
+
+    resource
+    |> Ash.Changeset.for_create(action, arguments)
+    |> create!(opts)
+  end
+
+  def create!(changeset, opts) do
     Ash.Helpers.expect_changeset!(changeset)
     Ash.Helpers.expect_options!(opts)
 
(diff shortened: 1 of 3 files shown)


lib/ash.ex
```diff
@@ -1953,11 +1957,22 @@ defmodule Ash do
...
           Ash.Resource.record()
           | {Ash.Resource.record(), list(Ash.Notifier.Notification.t())}
           | no_return
-  def create!(changeset, opts \\ []) do
+  def create!(changeset_or_resource_with_argument, opts \\ [])
+
+  def create!({resource, arguments}, opts) do
+    action = opts[:action] || Ash.Resource.Info.primary_action!(resource, :create).name
+    arguments = Enum.into(arguments, %{})
+
```
zachdaniel, line 1969: Because we construct the changeset here, we also need to pass the relevant opts there. What this would look like is something like:

```elixir
changeset_opts = Keyword.take(opts, Keyword.keys(Ash.Changeset.for_create_opts()))
create_opts = Keyword.take(opts, Keyword.keys(@create_opts_schema))
```

We'll need to do this in each place.

zachdaniel:
Looks good, will need some adjustments for option handling, but otherwise seems like a nice quality of life addition :)

lib/ash.ex
pierrelegall: There is no `for_update_opts` :thinking:

pierrelegall: There is no `Ash.Changeset.for_update_opts()` :thinking:

zachdaniel: We will need to add it then :) It should return the schema used in `Ash.Changeset.for_update`, and should be `@doc false`

pierrelegall: I allowed myself to:

```elixir
  @for_update_opts @for_create_opts

  @doc false
  def for_update_opts, do: @for_update_opts
```

for now, because I saw no differences :)

zachdaniel (approved):

zachdaniel:

vonagam:
Hm, but why tuple? Shouldn't args be more ergonomic? Especially for update, like `Ash.get!(...) |> Ash.update(...)`.

First argument (atom for create or record for update) clearly allows to pattern match without ambiguity.

The only difference is that you cannot pass params as keyword list, but that's already currently present limitation.

Also it seems that the linked issue is not right one.

zachdaniel:
Well, the current limitation is that there is no way to pass params to `Ash.create` and `Ash.update`, I guess the alternative would be adding an option like `params` or `input` option, instead of using a tuple. Then it would be:

```elixir
Ash.create!(Resource, params: %{...})
```

and

```elixir
Ash.update!(%Record{}, params: %{...})
```

That would probably be more ergonomic in general.

zachdaniel:
Yeah, I think we should do it that way, good call. @vonagam do you have a few to make that change? I can add it to my list if not.

zachdaniel:
Also perhaps @pierrelegall would you be willing to do it? Just make sure whoever decides to do it comments saying so so we don't have multiple people do it :)

vonagam:
Why not allow three args with second being params, like in code interface?

zachdaniel:
Well, I was thinking that would be a much bigger change, but I guess it isn't really? Its backwards compatible if we do the whole "if its a keyword then its opts" thing. I'm interested. That also enables us to support the bulk versions i.e `Ash.update!(Query, %{input: :value})` which is currently only supported in the code interface.

If we're sure its backwards compatible and doesn't screw anything up, then I'm down for it.

zachdaniel:
If there are *small* breakages that may be okay as well as this is still technically in a release candidate.

vonagam:
"if its a keyword then its opts" - right, the same as in code interface. Because of that rule plus the fact that atom/record were not supported before it should not be breaking in any way. 

As for bulk - interesting, that would also be consistent with code interface.

For record - I'm busy at the moment, so if you want to do it - do it, especially if you want to add that bulk thing.

zachdaniel:
I'll add it to my list, won't add the bulk stuff quite yet I don't think, but maybe :) @pierrelegall just pinging you again so you can see the changes to this feature.

pierrelegall:
I suppose it is much better this way!

This is the benefits:
- pipeable easily
- use a map always, not a keyword list, this is more consistente
- empty args can be implicit : no params/input/args means an empty hash

I'm ok to do it. Thanks @vonagam for your feedback :)

zachdaniel:
nice! That sounds great. Just remember the `if Keyword.keyword?` but to distinguish opts from params.

pierrelegall:
Propositions for the option name above is `params` and `input`.

However, this input arguments to the action. So I vote for `args` :hand:

```elixir
Ash.create!(Resource, args: %{...})
```

zachdaniel:
We won't need the option anymore, since it will be an optional second argument to the function.

zachdaniel:
But also, `args` is a bit overloaded, because actions have arguments *and* accepted attributes. What we typically call that is `input`.

pierrelegall:
I do not see a simple solution to make the `input` as a second argument and not an option.

The problem is that `opts` is an optional argument, that was the initial reason why we choose to use a tuple.

zachdaniel:
This is where the keyword check I was talking about comes into play. This is how we do it in the code interface:

```elixir
def function(initial, params_or_opts \\ %{}, opts \\ []) do
  {params, opts} =
    if opts == [] && Keyword.keyword?(params_or_opts) do
      {%{}, params_or_opts}
    else
      {params_or_opts, opts}
    end          
end
```

This allows:

```elixir
function(initial)
function(initial, %{input: :val})
function(initial, opt: :val}
function(initial, %{input: :val}, opt: :val)
```

PR #1057, opened 2024-04-25
fix: handle resource being passed as query code_interface.ex

barnabasJ:
This bit me today because I had code  like this and it worked in 2.0

```elixir
                Resource
                |> filter_board_codes(board_codes)
                |> filter_min_price(min_price)
                |> filter_max_price(max_price)
                |> filter_refundable_only(refundable_only)
```

and all the `filter_*` functions looked like this

```elixir
def filter_something(query, nil), do: query

def filter_something(query, something) do
  query
  |> Ash.Query.filter(....)
end
```

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -569,7 +569,7 @@ defmodule Ash.CodeInterface do
                         raise ArgumentError,
                               "Query resource #{inspect(other_resource)} does not match expected resource #{inspect(unquote(resource))}."
 
-                      resource when is_atom(resource) ->
+                      resource when is_atom(resource) and not is_nil(resource) ->
                         resource
 
                       query ->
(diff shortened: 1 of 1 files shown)


zachdaniel:
Right, makes sense. We should use `Ash.Query.new(Resource)`, and we should probably add something like a check to make sure it is actually the same resource as the resource this is declared for.

barnabasJ:
> Right, makes sense. We should use `Ash.Query.new(Resource)`, and we should probably add something like a check to make sure it is actually the same resource as the resource this is declared for.

I'll check why the tests fail later and add the check.

lib/ash/code_interface.ex
```diff
@@ -569,6 +569,9 @@ defmodule Ash.CodeInterface do
                         raise ArgumentError,
                               "Query resource #{inspect(other_resource)} does not match expected resource #{inspect(unquote(resource))}."
 
+                      resource when is_atom(resource) ->
```
zachdaniel, line 572:
```suggestion
                      resource when is_atom(resource) and not is_nil(resource) ->
```

zachdaniel (approved):

Issue #1051, opened 2024-04-23
Ash.create() and Ash.update() oneliner


pierrelegall:
I'm always frustrated by the verbosity of the code below when I have no domain/resource code action for it.

```elixir
MyResource
|> Ash.Changeset.for_action(:create, %{data: value})
|> Ash.create()
```

Maybe something like this could do the job:

```elixir
Ash.create({MyResource, data: value})
# or
Ash.create({MyResource, data: value}, action: :not_primary_create)
```

And it could be the same for an update action (with a record instead of an atom):

```elixir
Ash.update({record, data: value})
# or
Ash.update({record, data: value}, action: :not_primary_update)
```

I'll be happy to implement it myself :)

PS: still not an Ash 3.0 user ;p

zachdaniel:
I'm down 👍 Probably not worth back porting to 2.0 though.

sevenseacat:
This seems weird to me - it feels like another way of making a code interface (but not as succinctly and missing most of the benefits of doing so)?

zachdaniel:
Hm...I mean, it can be useful for cases when you are writing generic code (which is primarily what this interface is for). You're right that it does bring them closer to code interfaces, but it's much less verbose for the same result, so feels weird to artificially limit these interfaces just to push users elsewhere.

pierrelegall:
Note this have already been change to something more elegant on `main`:

```elixir
Ash.create(User, %{name: "John"}, action: :register_with_password)
```

pierrelegall:
It always seems weird to me do `Ash.create(changesets)`. A changeset is not created here: a resource is created from a changeset. Now we can do both, a changeset or a resource with its params.

Obviously, the execution path is now less self-evident.

PR #1060, opened 2024-04-25
improvement: better create/update first argument

pierrelegall:
`See` #1051 and #1055.

First argument of Ash.create (with a resource) and Ash.update (with a record) no longer are in a tuple with its arguments.

Arguments are now the second argument if it's not a keyword list.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -1981,33 +1981,12 @@ defmodule Ash do
   Create a record or raises an error. See `create/2` for more information.
   """
   @doc spark_opts: [{1, @create_opts_schema}]
-  @spec create!(Ash.Changeset.t() | resource_with_args(), Keyword.t()) ::
+  @spec create!(Ash.Changeset.t() | Ash.Changeset.t(), Keyword.t()) ::
           Ash.Resource.record()
           | {Ash.Resource.record(), list(Ash.Notifier.Notification.t())}
           | no_return
-  def create!(changeset_or_resource_with_args, opts \\ [])
-
-  def create!({resource, args}, opts) do
-    Ash.Helpers.expect_resource!(resource)
-    Ash.Helpers.expect_options!(opts)
-
-    changeset_opts = Keyword.take(opts, Keyword.keys(Ash.Changeset.for_create_opts()))
-    create_opts = Keyword.take(opts, Keyword.keys(@create_opts_schema))
-
-    action = opts[:action] || Ash.Resource.Info.primary_action!(resource, :create).name
-    args = Enum.into(args, %{})
-
-    resource
-    |> Ash.Changeset.for_create(action, args, changeset_opts)
-    |> create!(create_opts)
-  end
-
-  def create!(changeset, opts) do
-    Ash.Helpers.expect_changeset!(changeset)
-    Ash.Helpers.expect_options!(opts)
-
-    changeset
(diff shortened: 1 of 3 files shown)


zachdaniel:
So to make this change we should make `input` an optional second argument, not an option.

pierrelegall:
Yes sorry, this is a draft PR :wink:

lib/ash.ex
```diff
@@ -1973,15 +1977,18 @@ defmodule Ash do
...
           opts :: Keyword.t()
         ) ::
           {:ok, Ash.Resource.record()}
           | {:ok, Ash.Resource.record(), list(Ash.Notifier.Notification.t())}
           | {:error, term}
-  def create(changeset_or_resource, opts \\ [])
+  def create(changeset_or_resource, params_or_opts \\ %{}, opts \\ [])
+
+  def create(%Ash.Changeset{} = changeset, params_or_opts, opts) do
+    {_params, opts} = Ash.Helpers.get_params_and_opts(params_or_opts, opts)
```
zachdaniel, line 1990: 🤔 I think we should do this kind of thing here:

```elixir
if changeset.__validated_for_action__ && params != %{} do
  Logger.warn("input is being provided for a changeset that has already been validated for an action. params are being ignored"
else
  Ash.Changeset.for_create(changeset, params, ....)
end
```

lib/ash.ex
```diff
@@ -2350,14 +2361,20 @@ defmodule Ash do
...
           {:ok, Ash.Resource.record()}
           | {:ok, Ash.Resource.record(), list(Ash.Notifier.Notification.t())}
           | {:error, term}
-  def update(changeset_or_record, opts \\ [])
+  def update(changeset_or_record, params_or_opts \\ %{}, opts \\ [])
 
   @doc spark_opts: [{1, @update_opts_schema}]
-  def update(%Ash.Changeset{} = changeset, opts) do
+  def update(%Ash.Changeset{} = changeset, params_or_opts, opts) do
+    {_params, opts} = Ash.Helpers.get_params_and_opts(params_or_opts, opts)
```
zachdaniel, line 2376: Same note here as the above.

lib/ash.ex
pierrelegall: Or maybe raise an error?

I see no case where we could call `Ash.create(changeset, params, opts)`.

lib/ash.ex
pierrelegall: Replied above :)

lib/ash.ex
zachdaniel: We can raise an error, sure 👍 but we should handle that second case, where the changeset has not yet been validated for an action. That way you can
```elixir
Resource
|> Ash.Changeset.new()
|> Ash.Changeset.set_context(...)
|> Ash.create!(%{foo: :bar})
```

pierrelegall: 561574b should solve this issue.

zachdaniel:
Rebasing should resolve CI issues

pierrelegall:
> Rebasing should resolve CI issues

Done :wink:

zachdaniel:

PR #1070, opened 2024-04-29
improvement: pass context to anonymous validate functions

vonagam:
Breaking change for Ash 3 that aligns anonymous validate functions with those of changes, preparations and `before_action`/`after_action` built-ins helpers. Now anonymous validate functions should expect two arguments - a changeset and a context.


zachdaniel:

PR #1076, opened 2024-04-29
Add documentation on disabling pagination

bopm:
# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

This PR adds documentation on ways one can disable pagination in `required?` section of pagination API.


zachdaniel:
DSL documentation must be edited in the DSL definition, and then `mix spark.cheat_sheets` generates the files.

https://github.com/ash-project/ash/blob/main/lib/ash/resource/actions/read.ex#L121 is the place in question. Also, IIRC if you have `required?: false` and `default_limit: nil`, then pagination will only be done if you opt in.

bopm:
@zachdaniel thanks for pointing me in that direction. About second edit, I prefer to keep it in separate PR, which I'll probably offer after trying it in action.

zachdaniel:
My worry about the current wording is that this wording implies that you *must* supply `page: false` if you have `required?: false`. But technically if you have `required?: false, default_limit: nil`, it allows you to do it with no page parameters. How about adding "for example":

> Whether or not pagination can be disabled (by passing `page: false` to `Ash.Api.read!/2`, for example). Only relevant if some pagination configuration is supplied.

or something like that?

bopm:
This makes more sense, how about new version, does it work?

zachdaniel (approved):

zachdaniel:

PR #1077, opened 2024-04-29
Add doc text to the `default` option

moxley:
I spent a couple hours trying to understand the `default` option of the `attributes.attribute` DSL. The hope is this PR will help other developers facing similar issues as I did.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
So, I think it might be worth explaining that this happens because the default is casted to the type it contains, and *therefore*, you may want to consider the things you mentioned.

moxley:
> So, I think it might be worth explaining that this happens because the default is casted to the type of the attribute, and _therefore_, you may want to consider the things you mentioned.

What does that mean? In particular, "the default is casted to the type of the attribute". In my case, the attribute type was `:string`, and the default was an empty string. An empty string casted to a string is an empty string.

This came about when I tried to upgrade Ash in my project. The upgrade caused several GraphQL tests to fail. The code was trying to insert a record into the database with a null column value, on a column that was non-nullable. The `attribute` definition had `default: ""`. The test was not passing a key for that attribute, so I expected a `""` to be used in the INSERT statement. It worked before. Having to pass `constraints: [allow_empty?: true]` was the solution, unintuitive as it was.

zachdaniel:
Hm...not really sure why it was working before, but this actually is the intended behavior. With the default string type, an empty string casts to `nil`. https://hexdocs.pm/ash/3.0.0-rc.38/Ash.Type.String.html you can read more about it in the module docs. `allow_empty?` is a constraint on the string type. So the reason the default is going from `""` to `nil` is because we cast the default to the corresponding type (and constraints).

moxley:
Okay, got it. Maybe the best approach for this `:default` option doc is to point folks to the documentation for the appropriate Ash.Type.* for the the attribute type they're using. Something like:

> Note: The `default` value may not always be the actual value saved to the resource record for the default case. See the Ash.Type.* documentation for attribute's type to learn how empty values are handled.

zachdaniel:
TBH the only time it is likely to be problematic is in this exact case, so your specific example is good I think. But people need to be aware that the `allow_empty?` option is string specific, and because we are casting the value.Clearly a sharp edge that needs to be documented, but want to make sure its clear.

moxley:
Yes, it is an exact case, and a common one. I pushed a commit with my second proposed text.

zachdaniel:
Two things:
1. I think the wording can be made a bit more specific.

> Note: The `default` value is casted to the type. For `:string`, for example, if `allow_empty?` is `false`, the value `""` will be cast to `nil`. See the `:constraints` option, the `:allow_nil?` option, and the revelant Ash.Type.* documentation more.

2. we actually need to update it in the DSL file, and then generate this file with `mix spark.cheat_sheets`. https://github.com/ash-project/ash/blob/main/lib/ash/resource/attribute.ex#L124

moxley:
Thank you. I adjusted your example, putting it in the correct file, and I generated the .md files.

zachdaniel:

PR #1080, opened 2024-04-30
fix: streaming with limit/offset and no pagination

rbino:
If there were 0 results, it just kept reading and reading forever

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/stream.ex
+++ b/lib/ash/actions/read/stream.ex
@@ -140,7 +140,7 @@ defmodule Ash.Actions.Read.Stream do
 
           results = Ash.read!(query, Keyword.put(opts, :domain, domain))
 
-          if Enum.count(results) == limit do
+          if Enum.count(results) < limit do
             {results, false}
           else
             {results, offset + limit}
--- a/test/actions/stream_test.exs
+++ b/test/actions/stream_test.exs
@@ -58,6 +58,44 @@ defmodule Ash.Test.Actions.StreamTest do
     assert count == 10
   end
 
+  test "records can be streamed using limit/offset strategy with no pagination, count % batch_size != 0" do
+    1..10
+    |> Stream.map(&%{title: "title#{&1}"})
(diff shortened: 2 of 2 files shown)


zachdaniel:
@rbino this is in draft, but looks right to me. Good to merge?

rbino:
I've drafted it because reading it back I had the doubt it now returned the wrong number of results when the limit was _not_ reached, I'm verifying

lib/ash/actions/read/stream.ex
```diff
@@ -133,14 +133,14 @@ defmodule Ash.Actions.Read.Stream do
...
-            |> Ash.Query.limit(limit)
+            |> Ash.Query.limit(limit + 1)
             |> Ash.Query.offset(offset)
+            |> Ash.read!(Keyword.put(opts, :domain, domain))
+            |> Enum.split(limit)
 
-          results = Ash.read!(query, Keyword.put(opts, :domain, domain))
-
-          if Enum.count(results) == limit do
+          if Enum.empty?(rest) do
```
zachdaniel, line 143: Hmm....is there a way to accomplish this without having to read an additional record? For really large streams this means reading one additional record potentially many many times. Does `Enum.count(results) <= limit` not work?

rbino: I went with this approach since it's the one used in pagination, but I guess there it needs to be like this to return the correct value for the `more?` field. I guess here we can read the right number and just do an extra empty read at the end.

rbino: Actually, `Enum.count(results) <= limit` is always true by construction (the results will always be less than or equal to the limit). I guess the check would have to be `Enum.count(results) < limit`

zachdaniel: ah, yeah right :)

zachdaniel:

PR #1094, opened 2024-05-02
fix: fixes to some atomic validations

vonagam:
`attribute_in` and `changing` were using non-inverted condition in atomic implementation.

`present` was ignoring arguments for `at_most`/`at_least` options.


zachdaniel (approved):

zachdaniel:
Now that ETS supports atomics we can test these atomics in core.

PR #1103, opened 2024-05-02
improvement: normalise `List.wrap` usage for policy checks

vonagam:
1) Adds `List.wrap` call to `changing_attributes`. So one can write `changing_attributes(:my_field)`. 
2) Same for `changing_relationships`. So `changing_relationship` becomes just an alias for `changing_relationships`.
3) Move some wrap calls from runtime `match?` to options creation site.

If you prefer having separate methods I can undo 1 and 2 change and add `changing_attribute` instead. But I think one method is more consistent since you don't have `action` and `actions` for example.

3 is a small breaking change in case somebody was using those check modules directly without normalizing options.

### DIFF
--- a/lib/ash/policy/check/action.ex
+++ b/lib/ash/policy/check/action.ex
@@ -4,14 +4,13 @@ defmodule Ash.Policy.Check.Action do
 
   @impl true
   def describe(options) do
-    operator =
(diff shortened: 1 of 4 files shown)


zachdaniel (approved):

zachdaniel:

PR #1107, opened 2024-05-03
Fix missing tenant for many_to_many read

jechol:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -176,6 +176,7 @@ defmodule Ash.Actions.Read.Relationships do
               actor: source_query.context[:private][:actor],
               tracer: source_query.context[:private][:tracer]
             )
+            |> Ash.Query.set_tenant(source_query.tenant)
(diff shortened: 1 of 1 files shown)


lib/ash/actions/read/relationships.ex
```diff
@@ -176,6 +176,7 @@ defmodule Ash.Actions.Read.Relationships do
               actor: source_query.context[:private][:actor],
               tracer: source_query.context[:private][:tracer]
```
zachdaniel, line 177: We should set it here in the options list actually.

jechol: source_query.context[:private] doesn't have tenant, so I had to set it with source_query.tenant, which looks a bit awkward.

zachdaniel: Thats fine :)

zachdaniel:

PR #1111, opened 2024-05-03
fix: make lateral join filtering in ETS data layer work

rbino:
One of the two branches was using the non-bang version of a function, which returned `{:ok, filter}`. That value was then directly put in the Query struct, resulting in an invalid filter that was never applied.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #1114, opened 2024-05-05
Add documentation for Ash.Filter, linking to Expressions guide

moxley:
I found no mention in the documentation about the connection between `Ash.Filter` and the `Ash.Filter.Functions.*` modules. Seemed like an important connection to make.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -182,6 +182,19 @@ defmodule Ash.Filter do
 
   Maps are also accepted, as are maps with string keys. Technically, a list of
   `[{"string_key", value}]` would also work.
+
+  ### Filter Functions
+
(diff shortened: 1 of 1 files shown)


zachdaniel:
So I don't really think we want folks poking around those builtin function modules directly. We have the full filter api expressed in the `expressions.md` guide, but I can definitely see how this stuff is sprawling and the docs/interfaces need to be consolidated.

I think that we should, instead of pointing out `Ash.Filter.Functions.*` we should point folks to the expressions guide.

moxley:
Got it. That makes sense. I updated the doc accordingly. I also added an example on how to filter by a joined relationship.

zachdaniel (approved):

PR #1115, opened 2024-05-05
fix: run array type nil check after item constraints

sheerlox:
Following [my discussion with Zach on the Elixir Forum](https://elixirforum.com/t/array-type-empty-values-cannot-remove-all-empty-values/63330/3), this PR makes sure to run the item constraints before checking for `nil` values in an array type.

The main goal is to address empty strings being converted to `nil` *after* we checked for `nil` values in the array, effectively resulting in `nil` values in the data layer.

# Notes/Questions
- Moving the `nil` check has the side-effect of not running it anymore when `type.custom_apply_constraints_array?()` is true. In the Ash codebase that only impacts `Ash.EmbeddableType`. I don't know what the implications of that change are, so it's worth making sure it's not breaking any existing behavior for anyone.
- I've added a `test/type/array_test.exs` test file, which breaks the 1-to-1 mapping. Should I instead move its content to `test/type/type_test.exs`, given the special array type logic is contained in `ash/type/type.ex`?

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/type.ex
+++ b/lib/ash/type/type.ex
@@ -813,7 +813,11 @@ defmodule Ash.Type do
         |> Enum.with_index()
         |> Enum.reduce({[], []}, fn {item, index}, {items, errors} ->
           if type.custom_apply_constraints_array?() do
-            {[item | items], errors}
+            if is_nil(item) && not nil_items? do
+              {[item | items], [[message: "no nil values", index: index] | errors]}
+            else
+              {[item | items], errors}
+            end
           else
             case apply_constraints(type, item, item_constraints) do
               {:ok, value} ->
--- a/lib/ash/type/type.ex
+++ b/lib/ash/type/type.ex
@@ -812,31 +812,31 @@ defmodule Ash.Type do
         term
         |> Enum.with_index()
         |> Enum.reduce({[], []}, fn {item, index}, {items, errors} ->
-          if is_nil(item) && not nil_items? do
-            {[item | items], [[message: "no nil values", index: index] | errors]}
+          if type.custom_apply_constraints_array?() do
+            {[item | items], errors}
           else
-            if type.custom_apply_constraints_array?() do
-              {[item | items], errors}
-            else
-              case apply_constraints(type, item, item_constraints) do
-                {:ok, value} ->
+            case apply_constraints(type, item, item_constraints) do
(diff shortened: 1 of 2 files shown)


lib/ash/type/type.ex
```diff
@@ -812,31 +812,31 @@ defmodule Ash.Type do
         term
         |> Enum.with_index()
         |> Enum.reduce({[], []}, fn {item, index}, {items, errors} ->
-          if is_nil(item) && not nil_items? do
-            {[item | items], [[message: "no nil values", index: index] | errors]}
+          if type.custom_apply_constraints_array?() do
+            {[item | items], errors}
```
zachdaniel, line 816: We will need to add a check here for `item` being `nil`, to match the behavior below (as you pointed out)

zachdaniel:
I think the `array_test` is good, as we'll want to test the new constraint of removing empty items that we've talked about, and that gives array-oriented tests a nice place to live 👍

lib/ash/type/type.ex
```diff
@@ -812,31 +812,35 @@ defmodule Ash.Type do
         term
         |> Enum.with_index()
         |> Enum.reduce({[], []}, fn {item, index}, {items, errors} ->
-          if is_nil(item) && not nil_items? do
-            {[item | items], [[message: "no nil values", index: index] | errors]}
-          else
-            if type.custom_apply_constraints_array?() do
-              {[item | items], errors}
+          if type.custom_apply_constraints_array?() do
+            if is_nil(item) && not nil_items? do
```
sheerlox, line 816: Is the check duplication OK, or is there a better/cleaner way to do this? Still kinda new to Elixir so I don't master the best practices yet :sweat_smile:

sheerlox: Well I'm working on the next PR (`remove_nil_items?`) and created a private function in there to centralize the duplicated code :+1:

zachdaniel: A little duplication never hurt anybody 😆

zachdaniel:
Will merge once build passes

zachdaniel (approved):

zachdaniel:

PR #1124, opened 2024-05-07
feat: Add `cascade_destroy` to builtin changes.

jimsynz:
# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -297,6 +297,10 @@ defmodule Ash.Actions.Destroy.Bulk do
 
               %{bulk_result | notifications: notifications}
             else
+              Ash.Actions.Helpers.warn_missed!(atomic_changeset.resource, action, %{
+                resource_notifications: notifications
+              })
+
               %{bulk_result | notifications: []}
             end
           end
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -284,24 +284,20 @@ defmodule Ash.Actions.Destroy.Bulk do
           if opts[:return_notifications?] do
             %{bulk_result | notifications: notifications}
           else
-            if opts[:return_notifications?] do
-              bulk_result
+            if notify? do
+              notifications =
+                (bulk_result.notifications || []) ++ Process.get(:ash_notifications, [])
+
+              remaining_notifications = Ash.Notifier.notify(notifications)
+              Process.delete(:ash_notifications) || []
+
+              Ash.Actions.Helpers.warn_missed!(atomic_changeset.resource, action, %{
+                resource_notifications: remaining_notifications
+              })
+
+              %{bulk_result | notifications: notifications}
             else
-              if notify? do
-                notifications =
-                  (bulk_result.notifications || []) ++ Process.get(:ash_notifications, [])
-
-                remaining_notifications = Ash.Notifier.notify(notifications)
-                Process.delete(:ash_notifications) || []
-
-                Ash.Actions.Helpers.warn_missed!(atomic_changeset.resource, action, %{
-                  resource_notifications: remaining_notifications
-                })
-
-                %{bulk_result | notifications: notifications}
-              else
-                %{bulk_result | notifications: []}
-              end
+              %{bulk_result | notifications: []}
             end
           end
         after
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -259,6 +259,17 @@ defmodule Ash.Actions.Destroy.Bulk do
                   data_layer_context: opts[:data_layer_context] || %{}
                 }
               )
+              |> case do
+                {:ok, bulk_result} ->
+                  bulk_result
+
+                {:error, error} ->
+                  %Ash.BulkResult{
+                    status: :error,
+                    error_count: 1,
+                    errors: [Ash.Error.to_error_class(error)]
+                  }
+              end
             else
               do_atomic_destroy(query, atomic_changeset, has_after_batch_hooks?, input, opts)
             end
--- a/lib/ash/resource/change/builtins.ex
+++ b/lib/ash/resource/change/builtins.ex
@@ -286,6 +286,42 @@ defmodule Ash.Resource.Change.Builtins do
     {Ash.Resource.Change.Select, target: value, ensure?: true}
   end
 
+  @doc """
+  Cascade this resource's destroy action to a related resource's destroy action.
+
+  Adds an after-action hook that explicitly calls destroy on any records related
+  via the named relationship.  It will optimise for bulk destroys where
+  possible.
+
+  > #### Beware database constraints {: .warning}
+  >
+  > Think carefully before using this change with data layers which enforce
+  > referential integrity (ie PostgreSQL and SQLite) and you may need to defer
+  > constraints for the relationship in question.
+  >
+  > See also:
+  >   1. [`postgres.references.reference.deferrable` DSL](https://hexdocs.pm/ash_postgres/dsl-ashpostgres-datalayer.html#postgres-references-reference-deferrable)
+  >   2. [`sqlite.references.reference.deferrable` DSL](https://hexdocs.pm/ash_sqlite/dsl-ashsqlite-datalayer.html#sqlite-references-reference-deferrable)
+  >   3. [PostgreSQL's `SET CONSTRAINTS` documentation](https://www.postgresql.org/docs/current/sql-set-constraints.html)
+  >   4. [SQLite's `PRAGMA defer_foreign_keys` documentation](https://www.sqlite.org/pragma.html#pragma_defer_foreign_keys)
+
+  > #### Cascading notifications {: .tip}
+  >
+  > By default notifications are disabled for the related destroy. This is to avoid potentially sending a **lot** of notifications for high-cardinality relationships.
+
+  ## Options
+
+  #{Ash.Resource.Change.CascadeDestroy.opt_schema() |> Keyword.delete(:resource) |> Spark.Options.docs()}
+
+  ## Example
+
+      change cascade_destroy(:relationship)
+  """
+  @spec cascade_destroy(relationship :: atom) :: Ash.Resource.Change.ref()
+  def cascade_destroy(relationship, opts \\ []) do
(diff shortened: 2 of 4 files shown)


lib/ash/resource/change/builtins.ex
```diff
@@ -286,6 +286,27 @@ defmodule Ash.Resource.Change.Builtins do
     {Ash.Resource.Change.Select, target: value, ensure?: true}
   end
 
+  @doc """
+  Cascade this resource's destroy action to the related resource's destroy action.
+
+  ## Example
+
```
zachdaniel, line 293: Is it worth pointing out somewhere that this might not work with foreign key constraints in ash_postgres? You might need to defer constraints to use this strategy. And also if you're using ash_postgres you'd probably prefer the automatic cascade using PG?

zachdaniel (dismissed):

lib/ash/resource/change/cascade_destroy.ex
```diff
@@ -0,0 +1,204 @@
...
+  >   1. [PostgreSQL's `SET CONSTRAINTS` documentation](https://www.postgresql.org/docs/current/sql-set-constraints.html)
+  >   2. [SQLite's `PRAGMA defer_foreign_keys` documentation](https://www.sqlite.org/pragma.html#pragma_defer_foreign_keys)
+
+  ## Options
+
+  #{Spark.Options.docs(@option_schema)}
+
+  ## Example
+
+      change {Ash.Resource.Change.CascadeDestroy, resource: __MODULE__, relationship: :comments, action: :destroy}
```
zachdaniel, line 44: do you actually need the `resource` constraint to be set in the resource? Is this something we do for you? If not, you should always have a way to get the resource in the callbacks surely.

sevenseacat:
Will this handle any notifications emitted from calling the cascaded action?

zachdaniel:
ah, right, no it won't. We need to pass `notify?: true` as an option. At which point I think our process context notification picker-upper should do the job. But we should *probably* make that an option to the change and default it to false. Because this could be used on high cardinality relationships and that could be tons of notifications. I can't decide which way is the footgun. Notify but it could be way too many notifications, or don't but you might have to hunt down why you aren't getting delete notifications.

zachdaniel:
We should probably add a test for the notifications here though to confirm either way.

lib/ash/resource/change/builtins.ex
```diff
@@ -286,6 +286,45 @@ defmodule Ash.Resource.Change.Builtins do
...
+  > #### Beware database constraints {: .warning}
+  >
+  > Think carefully before using this change with data layers which enforce
+  > referential integrity (ie PostgreSQL and SQLite) and you may need to defer
+  > constraints for the relationship in question.
+  >
+  > See also:
+  >   1. [PostgreSQL's `SET CONSTRAINTS` documentation](https://www.postgresql.org/docs/current/sql-set-constraints.html)
+  >   2. [SQLite's `PRAGMA defer_foreign_keys` documentation](https://www.sqlite.org/pragma.html#pragma_defer_foreign_keys)
```
sevenseacat, line 304: This is why I originally had it set up as a before_action hook, not an after_action hook, to avoid needing to do this - would that help here?

zachdaniel: It depends on the relationship type. For cascading to a belongs to then you'd want to do this before the action. If cascading to other relationship types then you'd want to do it after. But also, doing this as a before action hook doesn't work for atomics (because we can only know what we *affected* with an atomic, not what *will be* affected).

sevenseacat: setting the FK as deferrable initially deferred (to make this work) should be doable with AshPostgres, eg: 

```elixir
    references do
      reference :album, deferrable: :initially
    end
```

AshSqlite says it supports the same option, though I have not tested it

zachdaniel: I totally forgot that we did that. Probably worth bringing up in this explanation There are some performance implications, but TBH I'm not entirely sure if they are enough to care or not.

lib/ash/resource/change/builtins.ex
```diff
@@ -286,6 +286,47 @@ defmodule Ash.Resource.Change.Builtins do
...
+  possible.
+
+  > #### Beware database constraints {: .warning}
+  >
+  > Think carefully before using this change with data layers which enforce
+  > referential integrity (ie PostgreSQL and SQLite) and you may need to defer
+  > constraints for the relationship in question.
+  >
+  > See also:
+  >   1. [`postgres.references.reference.deferrable` DSL](`e:ash_postgres:dsl-ashpostgres-datalayer.html#postgres-references-reference-deferrable`)
```
zachdaniel, line 303: What is the `e:` syntax?

jimsynz: https://hexdocs.pm/ex_doc/readme.html#auto-linking

turns out it emits a warning if the package isn't in the deps, so I've converted it to a normal link.

lib/ash/resource/change/cascade_destroy.ex
jimsynz: yeah, but we can't validate the options in the `init` callback if we don't have it.

zachdaniel: We should probably do something in the framework to provide that constraint or something then TBH. Don't want people to have to pass the current resource in. Also, unfortunately, at the moment the init callback is not actually called to initialize and then persisted or anything. The init is called every time that change is called. I intend to improve that in the future.

jimsynz:
> We should probably add a test for the notifications here though to confirm either way.

Updated to use the resource from the passed in changeset(s).

sevenseacat:
Testing this out on my app (because this will be super-useful!) but I have a couple of issues with it:

I have three resources:

Artist has:

```elixir
    destroy :destroy do
      change cascade_destroy(:albums, notify?: true)
    end
```

Album has:

```elixir
  postgres do
    references do
      reference :artist, deferrable: :initially
    end
  end

  actions do 
    destroy :destroy do
      change cascade_destroy(:notifications, notify?: true)
    end
  end
```

Notification has:

```elixir
  postgres do
    references do
      reference :album, deferrable: :initially
    end
  end

  pub_sub do
    publish :destroy, [:user_id, "destroy"]
  end

  actions do
    defaults [:destroy]
  end
```

1) I'm not seeeing pubsub notifications being sent when a destroy is cascaded.

With pubsub debugging turned on, I can see the destroys in the logs:

```
[debug] QUERY OK source="albums" db=2.1ms
DELETE FROM "albums" AS a0 WHERE (a0."id" = $1) ["9864c0ca-d240-4c0c-85de-5891e551bd52"]
↳ AshPostgres.DataLayer.destroy/2, at: lib/data_layer.ex:2560
[debug] QUERY OK source="notifications" db=0.2ms
DELETE FROM "notifications" AS n0 WHERE (n0."album_id"::uuid IN ($1::uuid)) RETURNING n0."id", 
n0."dismissed_at", n0."inserted_at", n0."user_id", n0."album_id" ["9864c0ca-d240-4c0c-85de-5891e551bd52"]
```

But no pubsub for the Notification destroy.

2. I get a crash when attempting to cascade cascade_destroys (eg. deleting an Artist -> deletes their Albums -> deletes Notifications for the albums). I can see in the logs that all destroys are successfully carried out, but then they roll back after the following error: 

```
14:25:26.520 [debug] QUERY OK source="artists" db=0.5ms
DELETE FROM "artists" AS a0 WHERE (a0."id" = $1) ["5ba7f062-efef-475f-b042-397be555cbb7"]
14:25:26.531 [debug] QUERY OK source="albums" db=0.3ms
DELETE FROM "albums" AS a0 WHERE (a0."artist_id"::uuid IN ($1::uuid)) RETURNING a0."id", 
a0."name", a0."year_released", a0."cover_image_url", a0."inserted_at", a0."updated_at", 
a0."artist_id", a0."created_by_id", a0."updated_by_id" ["5ba7f062-efef-475f-b042-397be555cbb7"]
14:25:26.534 [debug] QUERY OK source="notifications" db=0.2ms
DELETE FROM "notifications" AS n0 WHERE (n0."album_id"::uuid IN ($1::uuid)) RETURNING 
n0."id", n0."dismissed_at", n0."inserted_at", n0."user_id", n0."album_id" 
["c32ad110-50f8-4182-b8b2-ebad4999058b"]
14:25:26.534 [debug] QUERY OK db=0.1ms
rollback []


  1) test destroy destroys any associated albums when the artist is deleted (Tunez.Music.ArtistTest)
     test/tunez/music/artist_test.exs:143

     * %KeyError{key: :notifications, term: {:ok, %Ash.BulkResult{status: :success, errors: [], 
       records: [#Tunez.Music.Album<...>], notifications: [%Ash.Notifier.Notification{...}], 
       error_count: 0}}, message: "key :notifications not found in: {:ok,\n 
       %Ash.BulkResult{\n   status: :success,\n   errors: [],\n   .....}\n\nIf you are using 
       the dot syntax, such as map.field, make sure the left-hand side of the dot is a map"}

       (ash 3.0.0-rc.46) lib/ash/actions/destroy/bulk.ex:270: Ash.Actions.Destroy.Bulk.run/6
       (ash 3.0.0-rc.46) lib/ash.ex:2279: Ash.bulk_destroy!/4
       (ash 3.0.0-rc.46) lib/ash/resource/change/cascade_destroy.ex:67: anonymous fn/4 in Ash.Resource.Change.CascadeDestroy.change/3
       ...
```

I hope these will be straightforward tweaks!

lib/ash/actions/destroy/bulk.ex
```diff
@@ -297,6 +297,10 @@ defmodule Ash.Actions.Destroy.Bulk do
 
               %{bulk_result | notifications: notifications}
             else
+              Ash.Actions.Helpers.warn_missed!(atomic_changeset.resource, action, %{
```
jimsynz, line 300: @zachdaniel I'm not sure about this. do we actually want to just throw away notifications here?

sevenseacat:
The latest updates address both of my issues, cheers! 🙇‍♀️

lib/ash/actions/destroy/bulk.ex
zachdaniel: Yeah, we warn that we couldn't send them and then toss them, seems right to me 👍

zachdaniel (approved):

zachdaniel:
Awesome stuff 🎉

PR #1135, opened 2024-05-08
fix: access to change context in get_and_lock(_for_update) changes

barnabasJ:
context is now a struct and does not implement access behaviour


zachdaniel (approved):

zachdaniel:

PR #1157, opened 2024-05-13
docs: Fix incorrect documentation for Ash.create

simon-nystrom:
Hi,

Found this small error in the documentation for `Ash.create`, it causes the LSP to think the following will not succeed for example:

image

But it does in fact work


zachdaniel (approved):

zachdaniel:

PR #1158, opened 2024-05-13
fix: inifite recursion if query is empty

barnabasJ:
I'm not completely sure why this is happening. But right now, a test of mine fails for a read where I load through a calculation that returns a union, and I load relationships on the union. The load inside the query for the resource in the union looks like this `load: [relationship: []]`, and because it is an empty array instead of a `query`, it hit the last implementation of `clean_up_field_auth` indefinitely.


zachdaniel (approved):

zachdaniel:

PR #1163, opened 2024-05-14
fix: handle nil case

barnabasJ:
Another of my tests failed because nil was returned from `Map.get` above

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -609,6 +609,9 @@ defmodule Ash.Actions.Read do
                     end)
                   end)
               end
+
+            nil ->
+              record
           end
         end)
       end


lib/ash/actions/read/read.ex
```diff
@@ -609,6 +609,9 @@ defmodule Ash.Actions.Read do
                     end)
                   end)
               end
+
+            nil ->
```
zachdaniel, line 613: Hmmmmmm how are we even getting to `nil` here... I guess we're unloading a calculation or som thing somewhere?

zachdaniel (approved):

lib/ash/actions/read/read.ex
barnabasJ: I haven't had much time to dig deeper. I am still working on upgrading to 3.0, and I'm already over my allotted time budget

PR #1168, opened 2024-05-15
Update sensitive-data.md, tyring to clearify that public? applies to …

m0rt3nlund:
Added `, reading a resource, for example using Ash.read/2`, tyring to clearify that `public?` applies to reading operations via extensions and not when using `Ash.read` directly.

Also added some newlines as I thought made it easier to read.

Also thinking that https://hexdocs.pm/ash/3.0.1/dsl-ash-resource.html#attributes-attribute-public? maybe should mention that this property is used in `accept` for the actions as well?

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Hmm...yeah we could mention it there, but it I believe is mentioned in the docs for `accept` or `default_accept` when you use `:*` its "all public attributes"? Would have to check.

zachdaniel:

PR #1178, opened 2024-05-16
fix: make Ash.Resource.Validation.Changing work correctly in atomics

rbino:
- Use the correct option key to select the field
- Fail without hitting the data layer if the attribute is not changing
- Allow checking relationships like the non-atomic counterpart

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/validation/changing.ex
+++ b/lib/ash/resource/validation/changing.ex
@@ -53,16 +53,42 @@ defmodule Ash.Resource.Validation.Changing do
   end
 
   @impl true
-  def atomic(_changeset, opts, context) do
-    {:atomic, [opts[:field]], expr(^atomic_ref(opts[:attribute]) == ^ref(opts[:attribute])),
-     expr(
-       error(^InvalidAttribute, %{
-         field: ^opts[:field],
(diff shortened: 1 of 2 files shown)


lib/ash/resource/validation/changing.ex
```diff
@@ -53,16 +53,42 @@ defmodule Ash.Resource.Validation.Changing do
...
+               field: ^opts[:field],
+               value: ^atomic_ref(opts[:field]),
+               message: ^(context.message || "must be changing"),
+               vars: %{field: ^opts[:field]}
+             })
+           )}
+        end
+
+      %{type: :belongs_to, source_attribute: source_attribute} ->
+        if not Ash.Changeset.changing_attribute?(changeset, source_attribute) do
```
zachdaniel, line 74: lets flip the conditional (I think credo will yell at you)

zachdaniel (approved):

lib/ash/resource/validation/changing.ex
rbino: Done

zachdaniel:

Issue #1073, opened 2024-04-29
Add config to un-redact sensitive attributes and arguments in development


miguel-s:
**Is your feature request related to a problem? Please describe.**
In development, debugging values and errors for sensitive fields is difficult as the actual values will be redacted or filtered.

For example I'm trying to debug passwords generated by iOS autofill and why ash_authentication doesn't like them. Neither iOS nor ash want to tell me what the actual password is though.

**Describe the solution you'd like**
A config similar to `config :ash, :policies, show_policy_breakdowns?: true` for policies.
For un-redacting sensitive values in development only.

**Describe alternatives you've considered**
None 😅 

**Express the feature either with a change to resource syntax, or with a change to the resource interface**
```elixir
# config/dev.exs
config :ash, :resources, show_sensitive?: true
```

zachdaniel:
Sounds useful! PRs welcome :)

nallwhy:
Can I take this one?

zachdaniel:
Yep! Go ahead 🙇

pierrelegall:
Thanks! Can we close this issue?

PR #1180, opened 2024-05-18
feat: Implement `show_sensitive?` config

nallwhy:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

This PR implements #1073.

### DIFF
--- a/lib/ash/helpers.ex
+++ b/lib/ash/helpers.ex
@@ -599,11 +599,9 @@ defmodule Ash.Helpers do
   def deep_merge_maps(_left, right), do: right
 
   @doc false
-  def redact(value) do
-    if Application.get_env(:ash, :show_sensitive?, false) do
-      value
-    else
-      "**redacted**"
-    end
+  if Application.compile_env(:ash, :show_sensitive?, false) do
+    def redact(value), do: value
+  else
+    def redact(_value), do: "**redacted**"
   end
 end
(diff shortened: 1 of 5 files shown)


lib/ash/helpers.ex
```diff
@@ -597,4 +597,13 @@ defmodule Ash.Helpers do
   end
 
   def deep_merge_maps(_left, right), do: right
+
+  @doc false
+  def redact(value) do
```
zachdaniel, line 602: We should make this a compile time configuration, and use `Application.compile_env`. Then, conditionally define the function (for performance).

zachdaniel (changes requested):

lib/ash/helpers.ex
nallwhy: I've changed it to use `Application.compile_env/3`.
However I don't prefer that because `Application.get_env/3` takes less than 1us and the guideline advises avoiding compile-time environments.

> In any case, compile-time environments should be avoided. Whenever possible, reading the application environment at runtime should be the first choice.

https://hexdocs.pm/elixir/Application.html#module-compile-time-environment

zachdaniel: Hmm...that's true, but this could get called thousands of times every time you inspect a list of records. My initial reason for using compile_env was to make it not possible to toggle on at runtime in production though.

nallwhy: I agree :)

zachdaniel (dismissed):

zachdaniel:
Actually, one more thing: could you please add this to the security guide somewhere? Want to make sure this configuration is discoverable.

zachdaniel (approved):

PR #1198, opened 2024-05-23
fix: mixup between atomic upsert vs update capability of datalayers

lejoko:
The problems appeared in the ash_mysql module that I'm currently writing when I removed the `{:atomic, :upsert}` capability from my module but left the `{:atomic, :update}` one : updates with atomics started complaining about the missing upsert capability.

I'm not 100% sure that this is the right fix, and I have no idea how I could write non regression tests for this. However tests that wrongly failed in my module now pass and all tests still pass in ash.


zachdaniel (approved):

zachdaniel:

PR #1214, opened 2024-05-30
fix: use correct boolean operation names in Filter.find/4

rbino:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


rbino:
I've found this by chance while reading the implementation to see what `Filter.find/4` was doing, I don't have enough context on `Filter` to create a regression test but I guess one wouldn't hurt (from what I see this could affect optimized boolean expression generation)

zachdaniel (approved):

PR #1229, opened 2024-06-07
Pagination: support pagination's `count` on actions besides reads

davidebriani:
Read actions support specifying pagination options on `load` statements for relationships.
These options are correctly respected, and the `count: true` option results in Ash counting the total number of related resources and returning the count in the page returned as the loaded relationship.

However, actions such as `Ash.update!/3` do not properly respect the `load` option since the resulting `count` field is left as `nil`. 

The PR adds a sample failing test.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -178,6 +178,8 @@ defmodule Ash.Actions.Read do
 
     query = load_and_select_sort(query)
 
+    query = add_relationship_count_aggregates(query)
+
     pkey = Ash.Resource.Info.primary_key(query.resource)
 
     missing_pkeys? =
--- a/test/actions/create_test.exs
+++ b/test/actions/create_test.exs
@@ -607,6 +607,288 @@ defmodule Ash.Test.Actions.CreateTest do
     end
   end
 
+  describe "load" do
+    test "allows loading has_many relationship on the changeset" do
+      post1 = Ash.create!(Post, %{title: "Post 1"})
+      post2 = Ash.create!(Post, %{title: "Post 2"})
+
+      load_query =
+        Post
+        |> Ash.Query.sort(title: :asc)
+        |> Ash.Query.select([:title])
+
+      author =
+        Author
+        |> Ash.Changeset.new()
+        |> Ash.Changeset.change_attribute(:name, "Author")
+        |> Ash.Changeset.manage_relationship(:posts, [post2, post1], type: :append_and_remove)
+        |> Ash.Changeset.load(posts: load_query)
+        |> Ash.create!()
+
+      assert [%Post{title: "Post 1"}, %Post{title: "Post 2"}] = author.posts
+    end
+
+    test "allows loading has_many relationship on the action options" do
+      post1 = Ash.create!(Post, %{title: "Post 1"})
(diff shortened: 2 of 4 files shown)


zachdaniel:
@rbino would you have a few to look into this?

lib/ash/actions/read/read.ex
```diff
@@ -1972,6 +1972,11 @@ defmodule Ash.Actions.Read do
            (page[:count] != false and action.pagination.countable == :by_default))
 
     cond do
+      # Here it returns nil.
+      # Is it making an assumption that is not satisfied later?
+      # Commenting this and interrogating the datalayer below would result in
+      # the Ets datalayer failing since `run_aggregate_query_with_lateral_join`
+      # is not supported, and Postgres failing when compiling Ecto subqueries.
```
rbino, line 1979: For relationship count it's normal it returns `nil` here, it should add an aggregate from the parent [here](https://github.com/ash-project/ash/blob/b3a3d3d406b71a9f5a0f8e183f144d06f481fb37/lib/ash/actions/read/read.ex#L486). I'm guessing that since the read for the parent happens in the `update` action, it doesn't pass through here?

davidebriani: Thanks, you pointed me in the right direction: I managed to make the test pass.

I'll double check my solution and add a couple more tests before pushing a new commit.

davidebriani: The fix seems actually straightforward, but I took the chance add a bunch of tests on the `load` statements for actions (create, update, destroy).

lib/ash/actions/read/read.ex
```diff
@@ -855,7 +855,7 @@ defmodule Ash.Actions.Read do
 
     if missing_pkeys? ||
          (Enum.empty?(must_be_reselected) && Enum.empty?(query.aggregates) &&
-            Enum.empty?(calculations_in_query)) do
+            Enum.empty?(calculations_in_query) && Enum.empty?(query.load)) do
```
zachdaniel, line 858: I think this check may be problematic. Specifically, the "load" statement should already be "flushed out" to `aggregates` and `calculations_in_query`. What this would say to be is that the work to `add_relationship_count_aggregates/1` needs to be done earlier in the pipeline, before we split off the aggregates/calcs, which would allow us to leave this section of code unchanged.

davidebriani: I followed your advice and moved the change before in the pipeline.
However I am bit lost with what's going on in the code, so maybe double check my guess :)

zachdaniel:

lib/ash/actions/read/read.ex
```diff
@@ -178,6 +178,8 @@ defmodule Ash.Actions.Read do
 
     query = load_and_select_sort(query)
 
+    query = add_relationship_count_aggregates(query)
```
rbino, line 181: If this is now done here shouldn't it be removed from line 435?

zachdaniel: yep :)

zachdaniel: I'll do that and push it up

PR #1232, opened 2024-06-08
Try to fix the compare/2 implementations

adonig:
# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [x] Update dependencies


zachdaniel:
There are some good changes in here, but another interesting bit is that `compare` should actually be defined automatically if `bulk_compare` is defined and should call into `bulk_compare`. I'll merge this and make those changes.

zachdaniel:

zachdaniel:
Nevermind, my changes don't really work. `bulk_compare` returns boolean statements and `compare` returns a "comparison". For now we'll leave it as is although in the future this should probably be redesigned.

PR #1234, opened 2024-06-10
`load` relationships on the results of bulk operations

davidebriani:
On paper, bulk operations (create, update, destroy) currently supports a `load` option used to specify "a load statement to apply to records", provided that the `return_records?` option is also set to `true` and records are returned.

The relevant changes and unit tests are added here to verify that the `load` statement is respected and it correctly loads relationships.

Note: loading relationships with pagination on results for bulk destroys is still not supported. Indeed, relationships are currently queried using a lateral join but after the resource deletion has happened, so it looks like nothing is related.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -184,6 +184,8 @@ defmodule Ash.Actions.Destroy.Bulk do
         {atomic_changeset, opts} =
           Ash.Actions.Helpers.set_context_and_get_opts(domain, atomic_changeset, opts)
 
+        atomic_changeset = Ash.Actions.Helpers.apply_opts_load(atomic_changeset, opts)
+
         atomic_changeset = %{atomic_changeset | domain: domain}
 
         atomic_changeset =
@@ -503,6 +505,21 @@ defmodule Ash.Actions.Destroy.Bulk do
               {results, []}
             end
 
+          {results, errors, error_count} =
+            case load_data(
+                   results,
+                   atomic_changeset.domain,
+                   atomic_changeset.resource,
+                   atomic_changeset,
+                   opts
+                 ) do
+              {:ok, results} ->
+                {results, [], 0}
+
+              {:error, error} ->
+                {[], List.wrap(error), Enum.count(List.wrap(error))}
+            end
+
           notifications =
             if opts[:notify?] do
               notifications ++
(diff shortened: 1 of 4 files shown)


zachdaniel:
So to confirm this is tests of the `load` option for bulk actions, and a fix for it not being honored on bulk destroys, is that right?

davidebriani:
> So to confirm this is tests of the `load` option for bulk actions, and a fix for it not being honored on bulk destroys, is that right?

Yes. For clarification:
- Similar batches of unit tests are added for create/update/destroy to assert the expected behavior
- Bulk updates were already working
- Bulk creates currently break [here](https://github.com/ash-project/ash/blob/8dc0c2cb05e541cf73f0d18b3f7e315c5b0eadae/lib/ash/actions/create/bulk.ex#L189), resulting in errors such as `(Protocol.UndefinedError) protocol Enumerable not implemented for #Ash.Changeset...`, so it requires some troubleshooting
- Bulk destroys were working but not loading relationships. The code changes fix only non-paginated loads for now. Paginated loads currently return `%Ash.Page.Keyset{}` (instead of Offset) with`count: nil` and with no result.

zachdaniel:
Loading relationships on bulk destroys is actually kind of a dodgy proposal. If you're doing anything that requires a lateral join, it will look like nothing is related because we load relationships after-the-fact 🥲. Someday in the future we can support doing it all in one query but that is dodgy for its own reasons.

davidebriani:
Ah that's why, it makes sense! You spared me some good digging :)

I would rather not embark in a big rewriting on this PR, so my proposal is:
- we could keep the fix for non-paginated loads on destroys, if it makes sense. Regarding the paginated loads, for the moment I would just remove the introduced unit tests that are failing.
- we can focus on making bulk creates work.

zachdaniel:
@davidebriani as for it failing there, that is likely just because that is where we "trigger" things w/ the `Stream.run`. Will need to look more into it. Is there anything else in the stacktrace that points to where it might be going wrong? I can look in the next few days otherwise.

Ultimately we allow folks to load relationships on destroy, so we should do our best to honor that. It will just have some unfortunate caveats for now.

davidebriani:
@zachdaniel turns out I cannot read a stack trace :-P

The issue was that the bulk create assumed that `inputs` was a list of maps: a list of changesets was not actually contemplated, regardless of what the doc says.

zachdaniel:
Hmm....I'm not sure that we actually want to accept a list of change sets. Its a pretty big change 🥶

zachdaniel:
Might need to think about this more. Seeing as it was never working accepting change sets, perhaps we should walk it back and adjust the docs, and the current tests/behavior for now to account for that, and we can decide in a separate PR if we want to support that.

davidebriani:
@zachdaniel I left out the changes regarding bulk creates, I'm only adding the unit tests here.

I'll open a new draft PR as a placeholder to discuss how we want to handle changeset inputs in bulk creates.

zachdaniel:

PR #1236, opened 2024-06-11
fix: accept changeset inputs on bulk creates

davidebriani:
As stated in the documentation, bulk creates accept either a list of changesets or a list of maps as inputs.

However, bulk create is not actually working when providing changesets as inputs.

This change proposes a fix to ensure that bulk creates support changesets as the input argument.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -574,18 +574,34 @@ defmodule Ash.Actions.Create.Bulk do
          action,
          opts,
          lazy_matching_default_values,
-         base,
+         base_changeset,
          argument_names
        ) do
-    base
+    changeset =
+      case input do
+        %Ash.Changeset{} ->
+          input
+
+        input ->
+          handle_params(
+            base_changeset,
+            Keyword.get(opts, :assume_casted?, false),
+            action,
+            opts,
+            input,
+            argument_names
+          )
+      end
+
+    changeset =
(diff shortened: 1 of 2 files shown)


lib/ash/actions/create/bulk.ex
```diff
@@ -574,18 +574,34 @@ defmodule Ash.Actions.Create.Bulk do
...
          opts,
          lazy_matching_default_values,
-         base,
+         base_changeset,
          argument_names
        ) do
-    base
+    changeset =
+      case input do
+        %Ash.Changeset{} ->
```
zachdaniel, line 582: So, there are some complexities here. A few of them

* the changeset resource must be validated to be the same as the current resource
* the changeset action must be validated to be the same as the current action, or it must be coerced
* if `changeset.__validated_for_action__` is set to the current action, then we can use the changeset as-is. 
* Otherwise we need to treat it as if it's being called with empty params, which actually has some complex issues of its own. We still need to call `Ash.Changeset.handle_params/4`, but how do we handle any existing modifications that have been made? Perhaps if line 631 does two `Map.merge` calls to merge existing changeset attributes and arguments, that will be sufficient.

zachdaniel:
Will close for now since there hasn't been recent progress, happy to continue the conversation later!

PR #1242, opened 2024-06-14
do not lose constraints context

barnabasJ:
only passing the item constraints led to an forbidden error for a destroy on an embeddable type because the context with the actor was missing.


zachdaniel (approved):

zachdaniel:

PR #1258, opened 2024-06-24
fix: run notifications in the shorthand code interface methods

kernel-io:
Not sure how you'd want to implement this - or if it was excluded for a reason, but when you run an update / destroy action via a code interface with the shorthand 'id' method (i.e: `Resource.do_something(resource_id)`) no notifications get run.

i.e: everything works if you do 
```elixir
123
|> Resource.get_by_id!()
|> Resource.do_something()
```

no notifications get triggered if you do
```elixir
Resource.do_something(123)
```
(do_something is an atomic update action)

There is also some weirdness surrounding the use of `return_notifications?: true` in code interfaces.

i.e: 
```elixir
{:ok, record, notifications} = PredictionTask.reset_processing("02wMwEs82mpgrHN7hlS8vX", return_notifications?: true)
```
throws a match error as it only returns a 2 element tuple. however

```elixir
{:ok, record, notifications} = 
PredictionTask.get_by_id!("02wMwEs82mpgrHN7hlS8vX") 
|> PredictionTask.reset_processing(return_notifications?: true)
```
works correctly and has notifications returned (but still not sent)

(reset_processing is a non atomic update action)

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -832,6 +832,7 @@ defmodule Ash.CodeInterface do
                             bulk_opts
                             |> Keyword.put(:return_records?, true)
                             |> Keyword.put(:return_errors?, true)
+                            |> Keyword.put(:return_notifications?, true)
                           else
                             bulk_opts
                           end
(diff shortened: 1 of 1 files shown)


lib/ash/code_interface.ex
```diff
@@ -832,6 +832,7 @@ defmodule Ash.CodeInterface do
                             bulk_opts
                             |> Keyword.put(:return_records?, true)
                             |> Keyword.put(:return_errors?, true)
+                            |> Keyword.put(:return_notifications?, true)
```
zachdaniel, line 835: I think the missing piece is `notify?: true` here, and then the rest can likely be omitted?

zachdaniel:
@kernel-io can you try `main`? I've added a test and a fix that makes it pass.

PR #1261, opened 2024-06-24
Fix the issue with the order of cases

adonig:
There was an issue with the order of cases in my previous PR that prevented expressions like

```[relation: [in: [:a, :b]]]```

to get parsed correctly. The new changes make sure they will get turned into

```[:relation_id: [in: [:a, :b]]]```

# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [x] Update dependencies

### CODE REVIEW

**Comment by adonig** [comment]
My bad, I forgot the case ```[author: [id: [in: [1,2,3]]]]``` and I forgot to run the tests 😳

**Comment by zachdaniel** [comment]
Sorry, haven't had a chance to get across this. What's the status then? Did the last PR introduce a bug? Or is this just further enhancement on top of that?

**Comment by adonig** [comment]
I don't believe it introduced a bug. Let us quickly look at the code.

We kept the code that previously was the if-case:
```elixir
          is_list(nested_statement) || is_map(nested_statement) ->
            case parse_expression(nested_statement, context) do
              {:ok, nested_expression} ->
                {:ok, BooleanExpression.optimized_new(:and, expression, nested_expression)}

              {:error, error} ->
                {:error, error}
            end
```

This is the newly introduced code:
```elixir
          rel.type != :many_to_many && !Map.get(rel, :no_attributes) &&
              [rel.source_attribute] == Ash.Resource.Info.primary_key(rel.destination) ->
            with attr <- attribute(%{public?: true, resource: rel.source}, rel.source_attribute),
                 %Ash.Resource.Attribute{type: type, constraints: constraints} = attr,
                 {:ok, casted} <- Ash.Type.cast_input(type, nested_statement, constraints) do
              add_expression_part({attr, casted}, %{context | resource: rel.source}, expression)
            end
```

This is the code that was previously the else-case but we now handle the case when attribute is nil graciously:
```elixir
          true ->
            with [field] <- Ash.Resource.Info.primary_key(context.resource),
                 attribute when not is_nil(attribute) <- attribute(context, field),
                 {:ok, casted} <-
                   Ash.Type.cast_input(attribute.type, nested_statement, attribute.constraints) do
              add_expression_part({field, casted}, context, expression)
            else
              _other ->
                {:error,
                 InvalidFilterValue.exception(
                   value: inspect(nested_statement),
                   message:
                     "a single value must be castable to the primary key of the resource: #{inspect(context.resource)}"
                 )}
            end
```

So whenever the nested_statement is a list or a map, we will execute the same code as before. This is for example the case when we parse a filter like `[author: [id: [in: [1,2,3]]]]` or `[category: [in: [:dishwashing, :washing]]`.

When the nested_expression isn't a list or a map, for example when parsing a filter like `[category: :cooking]`, the new code could get executed when `:category` isn't a `:many_to_many` relationship, `Map.get(rel, :no_attributes)` is false and `rel.source_attribute` equals the primary key of `rel.destination`. I took this condition from our discussion in Discord but I now realized it's not what we want, because most of the time the source attribute is `relationship_id` and the destination primary key is just `id`. Maybe here we should check that `rel.destination_attribute` equals the primary key of the destination?

Otherwise we will just execute the old else-case but we graciously handle the case when it previously crashed because the attribute was nil.

So I believe it doesn't introduce a bug, but it's just not yet doing what we want it to do. Right now I make it do what we want and also support filter expressions like `[category: [in: [:dishwashing, :washing]]` which should be the same as `[category: [id: [in: [:dishwashing, :washing]]]]` when the condition is true. This time I will add enough tests to make sure we are good 😅

adonig:
It took me some time to find out how to handle both `[author1: [in: [1, 2, 3]]]` and `[author1: [id: [in: [1, 2, 3]]]` and in the end I came up with this:

```elixir
    cond do
      rel = relationship(context, field) ->
        # if the relation type isn't :many_to_many and :no_attributes isn't true and the destination attribute
        # of the relationship is the primary key of the destination and the source attribute is public try to
        # use the source attribute instead of field. If this attempt returns an error, i.e. when
        # {attr.name, nested_statement} is {author1_id, [id: [in: [1, 2, 3]]]}, just execute the old code.
	with relation_type when relation_type != :many_to_many <- rel.type,
	     true <- !Map.get(rel, :no_attributes),
	     [pk] when pk == rel.destination_attribute <- Ash.Resource.Info.primary_key(rel.destination),
	     attr <- attribute(%{public?: true, resource: rel.source}, rel.source_attribute),
             %Ash.Resource.Attribute{} = attr,
             {:ok, result} <- add_expression_part({attr.name, nested_statement}, context, expression) do
	  {:ok, result}
	else
	  _ ->
	    context =
	      context
	      |> Map.update!(:relationship_path, fn path -> path ++ [rel.name] end)
	      |> Map.put(:resource, rel.destination)
		
	    if is_list(nested_statement) || is_map(nested_statement) do
	      case parse_expression(nested_statement, context) do
		{:ok, nested_expression} ->
                  {:ok, BooleanExpression.optimized_new(:and, expression, nested_expression)}
		    
		{:error, error} ->
                  {:error, error}
	      end
	    else
	      with [field] <- Ash.Resource.Info.primary_key(context.resource),
                   attribute when not is_nil(attribute) <- attribute(context, field),
                   {:ok, casted} <-
		     Ash.Type.cast_input(attribute.type, nested_statement, attribute.constraints) do
		add_expression_part({field, casted}, context, expression)
	      else
		_other ->
                  {:error,
                   InvalidFilterValue.exception(
		     value: inspect(nested_statement),
		     message:
		       "a single value must be castable to the primary key of the resource: #{inspect(context.resource)}"
                   )}
	      end
	    end
	end
```

zachdaniel:

PR #1266, opened 2024-06-25
test: assert that notifications are sent for specific cases

ryanrborn:
Asserts that notifications are sent for bulk updates and destroys using the `:stream` strategy with a data layer that supports transactions.

# Contributor checklist

Asserts and fixes #1262

### DIFF
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -1269,7 +1269,7 @@ defmodule Ash.Actions.Destroy.Bulk do
 
             store_error(ref, new_errors, opts, new_error_count)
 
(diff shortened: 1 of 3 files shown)


zachdaniel:

lib/ash/actions/destroy/bulk.ex
```diff
@@ -1269,7 +1269,7 @@ defmodule Ash.Actions.Destroy.Bulk do
 
             store_error(ref, new_errors, opts, new_error_count)
 
-            notifications = Process.get({:bulk_update_notifications, tmp_ref}) || []
+            notifications = Process.get({:bulk_destroy_notifications, tmp_ref}) || []
```
zachdaniel, line 1272: 🙇

PR #1270, opened 2024-07-01
fix manual relationships when using 'complex' types

kernel-io:
manual relationships don't seem to work when using primary key types that don't have '==' equality i.e: uuid_v7 or AshUUID. not many tests in this manual relationship stuff so not sure if I've broken other cardinality manual relationships by doing this...

### DIFF
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -707,7 +707,7 @@ defmodule Ash.Actions.Read.Relationships do
 
         value =
           Enum.find_value(map, fn {key, value} ->
-            if resource.primary_key_matches?(key, pkey_values) do
+            if resource.primary_key_matches?(Map.from_keys(pkey, key), pkey_values) do
               {:ok, value}
             end
           end) || :error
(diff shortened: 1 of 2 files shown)


lib/ash/actions/read/relationships.ex
```diff
@@ -707,7 +707,7 @@ defmodule Ash.Actions.Read.Relationships do
 
         value =
           Enum.find_value(map, fn {key, value} ->
-            if resource.primary_key_matches?(key, pkey_values) do
+            if resource.primary_key_matches?(Map.from_keys(pkey, key), pkey_values) do
```
zachdaniel, line 710: This shouldn't be necessary, we extract the keys that we need.

zachdaniel:
oh, interesting... 🤔  AshUUID is not our package so you'll have to take that up with them. The solution isn't necessarily to remove the `equal?` callback, I just noticed in passing based on this that we had one that was not necessary.

zachdaniel:
I will look at this shortly

zachdaniel:

PR #1273, opened 2024-07-02
fix: use action from changeset

barnabasJ:
the passed action seems to only be the name (atom) of the action and not the action definition


zachdaniel (approved):

zachdaniel:

PR #1275, opened 2024-07-03
fix: invert logic for lazy-loading calculations/aggregates

hwuethrich:
Currently the logic for `lazy?`-loading calculations and aggregates is wrong, which results in already loaded fields to be loaded again and unloaded fields not to be loaded at all.

The following example will only load the calculation or aggregate if it was already loaded:

```elixir
Ash.load(resource, [:some_agg, :another_calc], lazy?: true)
```

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel (approved):
🤦

zachdaniel:

PR #1280, opened 2024-07-04
Another change to the way add_expression_part/2 handles relations

adonig:
I had to change the way `add_expression_part/2` handles relations. Instead of using the source attribute and a recursive call it now uses the destination attribute and needs no recursive call anymore.

## Changes
- Old Method: `[author: [eq: 1]]` becomes `[author_id: [eq: 1]]`
- New Method: `[author: [eq: 1]]` becomes `[author: [id: [eq: 1]]]`

## Benefits
- The source attribute can now be private.
- Eliminates recursion by modifying the nested_path when all conditions are met.

## How it works
```elixir
  defp add_expression_part({field, nested_statement}, context, expression)
       when is_atom(field) or is_binary(field) do
    cond do
      rel = relationship(context, field) ->

        # Instead of trying to recursively call add_expression_part/2 with {source_attr.name, nested_expression}
        # we now change the nested_statement to [{pk, nested_statement}] where pk is the destination attribute
        # of the relationship. We can then use this new nested_statement with the old code. We make sure that
        # the nested_statement only gets changed when it's first key isn't an attribute, because otherwise we
        # would turn [author: [name: [eq: "Zach"]]] into [author: [id: [name: [eq: "Zach"]]]].
        nested_statement =
          with relation_type when relation_type != :many_to_many <- rel.type,
               destination <- rel.destination,
               dest_attr <- rel.destination_attribute,
               [pk] when pk == dest_attr <- Ash.Resource.Info.primary_key(destination),
               attr <- attribute(%{public?: true, resource: destination}, pk),
               %Ash.Resource.Attribute{} = attr,
               true <- is_list(nested_statement) || is_map(nested_statement),
               {kw, _} <- Enum.at(nested_statement, 0),
               nil <- attribute(%{public?: true, resource: destination}, kw) do
            if is_list(nested_statement) do
              [{dest_attr, nested_statement}]
            else
              Map.put(%{}, dest_attr, nested_statement)
            end
          else
            _ -> nested_statement
          end

          # here comes the old code
```

## Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [x] Update dependencies

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -2708,46 +2708,53 @@ defmodule Ash.Filter do
        when is_atom(field) or is_binary(field) do
     cond do
       rel = relationship(context, field) ->
-        with relation_type when relation_type != :many_to_many <- rel.type,
-             true <- !Map.get(rel, :no_attributes),
-             [pk] when pk == rel.destination_attribute <-
-               Ash.Resource.Info.primary_key(rel.destination),
-             attr <- attribute(%{public?: true, resource: rel.source}, rel.source_attribute),
-             %Ash.Resource.Attribute{} = attr,
-             {:ok, result} <-
-               add_expression_part({attr.name, nested_statement}, context, expression) do
-          {:ok, result}
-        else
-          _ ->
-            context =
-              context
-              |> Map.update!(:relationship_path, fn path -> path ++ [rel.name] end)
-              |> Map.put(:resource, rel.destination)
-
-            if is_list(nested_statement) || is_map(nested_statement) do
-              case parse_expression(nested_statement, context) do
-                {:ok, nested_expression} ->
-                  {:ok, BooleanExpression.optimized_new(:and, expression, nested_expression)}
-
-                {:error, error} ->
-                  {:error, error}
-              end
+        nested_statement =
+          with relation_type when relation_type != :many_to_many <- rel.type,
+               destination <- rel.destination,
+               dest_attr <- rel.destination_attribute,
+               [pk] when pk == dest_attr <- Ash.Resource.Info.primary_key(destination),
+               attr <- attribute(%{public?: true, resource: destination}, pk),
+               %Ash.Resource.Attribute{} = attr,
+               true <- is_list(nested_statement) || is_map(nested_statement),
+               {kw, _} <- Enum.at(nested_statement, 0),
+               nil <- attribute(%{public?: true, resource: destination}, kw) do
+            if is_list(nested_statement) do
+              [{dest_attr, nested_statement}]
             else
-              with [field] <- Ash.Resource.Info.primary_key(context.resource),
-                   attribute when not is_nil(attribute) <- attribute(context, field),
-                   {:ok, casted} <-
-                     Ash.Type.cast_input(attribute.type, nested_statement, attribute.constraints) do
-                add_expression_part({field, casted}, context, expression)
-              else
-                _other ->
-                  {:error,
-                   InvalidFilterValue.exception(
-                     value: inspect(nested_statement),
-                     message:
-                       "a single value must be castable to the primary key of the resource: #{inspect(context.resource)}"
-                   )}
-              end
+              Map.put(%{}, dest_attr, nested_statement)
             end
+          else
+            _ -> nested_statement
+          end
+
+        context =
+          context
+          |> Map.update!(:relationship_path, fn path -> path ++ [rel.name] end)
+          |> Map.put(:resource, rel.destination)
+
+        if is_list(nested_statement) || is_map(nested_statement) do
+          case parse_expression(nested_statement, context) do
+            {:ok, nested_expression} ->
+              {:ok, BooleanExpression.optimized_new(:and, expression, nested_expression)}
+
+            {:error, error} ->
+              {:error, error}
+          end
+        else
+          with [field] <- Ash.Resource.Info.primary_key(context.resource),
+               attribute when not is_nil(attribute) <- attribute(context, field),
+               {:ok, casted} <-
+                 Ash.Type.cast_input(attribute.type, nested_statement, attribute.constraints) do
+            add_expression_part({field, casted}, context, expression)
+          else
+            _other ->
+              {:error,
+               InvalidFilterValue.exception(
+                 value: inspect(nested_statement),
+                 message:
+                   "a single value must be castable to the primary key of the resource: #{inspect(context.resource)}"
+               )}
+          end
         end
 
       attr = attribute(context, field) ->


zachdaniel:
I'm thinking that there are issues both with the original code and this code. You can do things like reference calculations and relationships in a nested filter, so the "is attribute" check is insufficient. 

We need to add some tests for this specific case with filter parsing. Can you add some tests for this implementation? I'll then suggest some example tests that would break this behavior. I think what we may want to do is then add a test for the original issue you saw, revert the code back to its original state(before the previous PR), and explore how to fix that original issue.

adonig:
This is the original implementation:

```elixir
      rel = relationship(context, field) ->
        context =
          context
          |> Map.update!(:relationship_path, fn path -> path ++ [rel.name] end)
          |> Map.put(:resource, rel.destination)

        if is_list(nested_statement) || is_map(nested_statement) do
          case parse_expression(nested_statement, context) do
            {:ok, nested_expression} ->
              {:ok, BooleanExpression.optimized_new(:and, expression, nested_expression)}

            {:error, error} ->
              {:error, error}
          end
        else
          with [field] <- Ash.Resource.Info.primary_key(context.resource),
               attribute <- attribute(context, field),
               {:ok, casted} <-
                 Ash.Type.cast_input(attribute.type, nested_statement, attribute.constraints) do
            add_expression_part({field, casted}, context, expression)
          else
            _other ->
              {:error,
               InvalidFilterValue.exception(
                 value: inspect(nested_statement),
                 message:
                   "a single value must be castable to the primary key of the resource: #{inspect(context.resource)}"
               )}
          end
        end
```

I believe it suffers from at least two issues:
1. It crashes when the destination primary key is private because we access attribute.type despite attribute is nil.
2. It supports filter statements like `[author: uuid]` but not filter statements like `[author: [in: [uuid1, uuid2]]]`. Instead it turns the latter into `[in: [uuid1, uuid2]]`.

I believe the implementation I suggested does better because it turns `[author: uuid]`  into `[author: [id: uuid]]` and  `[author: [in: [uuid1, uuid2]]]` into `[author: [id: [in: [uuid1, uuid2]]]]`.

In the current version it turns `author` into `author_id`. That works but it starts to fail when the source attribute is private. I realized this when I made my source attributes private because I didn't want them to show up in the JSON.

I believe those two tests already check that it works:

```elixir
    test "understands relationship filter subsets when a value coincides with the join field" do
      id1 = Ash.UUID.generate()
      id2 = Ash.UUID.generate()
      filter = Filter.parse!(Post, author1: [id: [in: [id1, id2]]])

      candidate = Filter.parse!(Post, author1_id: id1)

      assert Filter.strict_subset_of?(filter, candidate)
    end

    test "allows to omit the join field for belongs_to relationships" do
      id1 = Ash.UUID.generate()
      id2 = Ash.UUID.generate()
      filter = Filter.parse!(Post, author1: [in: [id1, id2]])

      candidate = Filter.parse!(Post, author1: id1)

      assert Filter.strict_subset_of?(filter, candidate)
    end
```

I made `author1_id` private to make sure this case also gets checked.

zachdaniel:
An example of something that would fail this is that you can use nested relationships or calculations in the expression syntax.

```elixir
[author: [org: [name: "fred"]]]
```

In that case it would be expanded by the current code to be `[author_id: [org: [name: "fred"]]]`

Or if `author` has a calculation on it called `has_friend`

```elixir
[author: [has_friend: true]]
```

And keep in mind you can mix and match these, i.e

```elixir
[author: [eq: 1, has_friend: true]]
```

So we need to test those cases and confirm that this transformation has the desired effects and does not break those formats.

adonig:
`[author: [eq: 1, has_friend: true]]`

This one is a bit harder. I believe I have to split the nested statement into multiple statements and then use recursion to handle them separately and combine the result using a boolean and.

I also realized that there is an issue with embedded fields but I have to check whether it already existed in the original implementation.

adonig:
I believe I found a bug. The code either doesn't get executed or `op_module` becomes `true`: 

https://github.com/ash-project/ash/blob/e1dffc0c0c4fe386dcbffc2038e9f1ad90fcb3aa/lib/ash/filter/filter.ex#L2857-L2872

I think it should be something like this instead:

```elixir
 op_module = get_operator(field) when match?([_, _ | _], nested_statement) -> 
 ```

**Comment by adonig** [comment]
I implemented something terrible that seems to work but I believe the whole filter module needs some love. Maybe we can add lots of tests to make sure everything works as expected and then start to clean it up piece by piece.

File: lib/ash/filter/filter.ex [hunk 1]
```diff
@@ -2854,11 +2851,13 @@ defmodule Ash.Filter do
...
-        with {:ok, [left, right]} <- hydrate_refs(nested_statement, context),
-             refs <- list_refs([left, right]),
+      (op_module = get_operator(field)) && match?([_, _ | _], nested_statement) ->
+        with {:ok, refs} <-
+               hydrate_refs(nested_statement, context),
+             refs <- list_refs(refs),
              :ok <- validate_refs(refs, context.root_resource, {field, nested_statement}),
-             {:ok, operator} <- Operator.new(op_module, left, right) do
+             {:ok, operator} <-
+               Operator.new(op_module, context.root_resource, refs) do
```
zachdaniel, line 2860: This doesn't seem right to me. The left side of the operator shouldn't be the root resource. It should be the left/right ref

lib/ash/filter/filter.ex
```diff
@@ -2708,46 +2708,43 @@ defmodule Ash.Filter do
        when is_atom(field) or is_binary(field) do
     cond do
       rel = relationship(context, field) ->
-        with relation_type when relation_type != :many_to_many <- rel.type,
-             true <- !Map.get(rel, :no_attributes),
-             [pk] when pk == rel.destination_attribute <-
-               Ash.Resource.Info.primary_key(rel.destination),
-             attr <- attribute(%{public?: true, resource: rel.source}, rel.source_attribute),
+        with rel_type when rel_type != :many_to_many <- rel.type,
+             dest <- rel.destination,
```
zachdaniel, line 2712: We still need the `true <- !Map.get(real, :no_attributes)`.

lib/ash/filter/filter.ex
```diff
@@ -2708,46 +2708,43 @@ defmodule Ash.Filter do
...
+             attr <- attribute(%{public?: true, resource: dest}, pk),
              %Ash.Resource.Attribute{} = attr,
-             {:ok, result} <-
-               add_expression_part({attr.name, nested_statement}, context, expression) do
-          {:ok, result}
+             true <- is_list(nested_statement) or is_map(nested_statement) do
+          {:ok,
+           Enum.reduce(nested_statement, true, fn {key, val}, acc ->
+             {:ok, expr_part} =
+               if is_nil(aggregate(%{public?: true, resource: dest}, key)) and
```
zachdaniel, line 2721: I think we actually want two branches here:

1. if the source_attribute is private, then we want to use the dest attribute.
2. otherwise, use this logic. But for now this should be fine 👍

lib/ash/filter/filter.ex
adonig: I tested with the in operator: `[author1: [in: [:id1, :id2]]]`

The existing code turned `[:id1, :id2]` into `[left, right]`. I believe left should be `author1.id` and right `[:id1, :id2]`.

lib/ash/filter/filter.ex
adonig: Right, I must have lost it somewhere on the way.

lib/ash/filter/filter.ex
zachdaniel: Even if that is the case, the `left` side of the `in` operator wouldn't be a resource, it would be a field or some other expression. In this case, I'm pretty sure this is to handle something in the format of `[in: [:id, [1, 2, 3]]]`.

adonig: Do we support both?

1. `[id: [in: [1, 2, 3]]]`
2. `[in: [id: [1, 2, 3]]]`

zachdaniel: 🤔 I *think* so? If we add a test for both, it should highlight the issue, I think. I believe the second one will fail now because of the change to `Operator.new` and that the first one will work because it doesn't go through this code path.

adonig: Yes, this fails now:

```elixir
      # prefix operator
      assert Filter.parse!(Post, author1: [in: [id: [id1, id2]]])
```

That makes it even harder to get this change working 😭

zachdaniel: The fix to `op_module` was required though, I can see how that would evaluate `op_module` to `true`.

zachdaniel: Alternatively, since this code path was broken already, we could just remove this match entirely :)

adonig:
Not sure why we get those dialyzer errors. Locally my dialyzer run succeeds:

```
Starting Dialyzer
[
  check_plt: false,
  init_plt: ~c"/home/asd/Code/github.com/adonig/ash/_build/dev/dialyxir_erlang-26.0.2_elixir-1.17.0_deps-dev.plt",
  files: [~c"/home/asd/Code/github.com/adonig/ash/_build/dev/lib/ash/ebin/Elixir.Ash.ProcessHelpers.beam",
   ~c"/home/asd/Code/github.com/adonig/ash/_build/dev/lib/ash/ebin/Elixir.Ash.Resource.Actions.Update.beam",
   ~c"/home/asd/Code/github.com/adonig/ash/_build/dev/lib/ash/ebin/Elixir.Ash.Resource.Change.GetAndLockForUpdate.beam",
   ~c"/home/asd/Code/github.com/adonig/ash/_build/dev/lib/ash/ebin/Elixir.Inspect.Ash.Vector.beam",
   ~c"/home/asd/Code/github.com/adonig/ash/_build/dev/lib/ash/ebin/Elixir.Ash.Resource.Dsl.Aggregates.Sum.beam",
   ...],
  warnings: [:unknown]
]
Total errors: 0, Skipped: 0, Unnecessary Skips: 0
done in 0m18.48s
done (passed successfully)
```

zachdaniel:

Issue #1287, opened 2024-07-05
Hide private fields with field policies.


torepettersen:
**Is your feature request related to a problem? Please describe.**
Currently if you query a resrouce with an interal function, e.g. `Ash.read` you would be able to see private fields, even if you have added field policies.

```elixir
attributes do
  attribute :private_field, :string
  attribute :admin_only_field, :string
end

field_policies do
  field_policy :admin_only_field do
    authorize_if actor_attribute_equals(:is_admin, true)
  end
end
```

```elixir
iex> Ash.read!(Resource, user: regular_user)
%{
  private_field: "Top secret",
  admin_only_field: #Ash.ForbiddenField<>,
}
```

**Describe the solution you'd like**
There should therefore be a way to hide private fields, if field policies apply:
```elixir
attributes do
  attribute :private_field, :string
  attribute :admin_only_field, :string
end

field_policies do
  hide_private? true
  field_policy :admin_only_field do
    authorize_if actor_attribute_equals(:is_admin, true)
  end
end
```

```elixir
iex> Ash.read!(Resource, user: regular_user)
%{
  private_field: #Ash.ForbiddenField<>,
  admin_only_field: #Ash.ForbiddenField<>,
}
```

A global flag, would also be good, so you don't forget to add the field to each field policy:
```elixir
config :ash, field_policies_hide_private: true
```

**Additional context**
Helpful conversation on the topic on [Elixir forum](https://elixirforum.com/t/trying-to-understand-ash-field-policy-behaviour/64679/6)

zachdaniel:
Thanks for the proposal! This should be *relatively* straightforward, as we apply field policies essentially in one place per action type. The global flag also sounds reasonable. PRs welcome!

PR #1289, opened 2024-07-07
feat: Allow field policies to hide private fields

torepettersen:
Closes https://github.com/ash-project/ash/issues/1287

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
So, looking at the implementation here, this is basically spitting distance from what you actually want, which is something like:

```elixir
field_policies do
  private_fields :include | :hide | :prevent
end
```

So `:include` would allow you to write field policies on private fields, `:hide` would hide them automatically and prevent you from writing field policies about them, and `:prevent` would leave them shown and prevent you from writing policies about them at all.

I'm still personally thinking that private fields ought not to have field policies, but it seems like something that we don't necessarily *need* to stop people from doing.

Would you want to take a stab at enhancing this PR to have those three options instead of a boolean flag?

torepettersen:
Thank you, that is a nice suggestion :blush: I will give it a try.

zachdaniel:
Maybe `:include | :hide | :show` would be better naming.

lib/ash/policy/authorizer/authorizer.ex
```diff
@@ -356,6 +356,20 @@ defmodule Ash.Policy.Authorizer do
...
+    schema: [
+      private_fields: [
+        type: {:one_of, [:show, :hide, :include]},
+        default: Application.compile_env(:ash, :policies)[:private_fields] || :show,
+        doc: """
+        Option for how private fields should be handeled by field policies in internal functions.
+
+        - :show will always show private fields
+        - :hide will always hide private fields
+        - :include will let you to write field policies for private fields and private fields
```
zachdaniel, line 369: We will want to have this as a single line, and tell users to consult the policies guide for more. Then, put this information there. Large option documentation blocks doesn't look good in the generated docs.

Other than that, this LGTM 🥳

zachdaniel (approved):

zachdaniel:
This looks great! I will merge this after the next release is cut.

zachdaniel:

PR #1309, opened 2024-07-12
feat: add relating_to_actor_via policy check

stocks29:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Adds a `relating_to_actor_via()` built-in policy check similar to `relating_to_actor` but supporting a relationship path to find where the entity relates to the actor.

This is PR is the result of this thread: https://elixir-lang.slack.com/archives/C0193AHFLPN/p1719753168326359 

I couldn't figure out how to make this work with `manage_relationship`. Any pointers here would be greatly appreciated! I added a test for this but marked it as skipped since it does not work.

Also, if there are better option names that better align with ash conventions please let me know. I tried to align them to the existing builtin option names but feel like the ones in the PR are not quite on point.

### DIFF
--- a/test/policy/relating_to_actor_via_test.exs
+++ b/test/policy/relating_to_actor_via_test.exs
@@ -1,0 +1,438 @@
+defmodule Ash.Test.Policy.RelatingToActorViaTest do
+  @doc false
+  use ExUnit.Case
+
+  alias Ash.Test.Domain, as: Domain
+
+  defmodule User do
+    use Ash.Resource, domain: Domain, data_layer: Ash.DataLayer.Ets
+
+    actions do
+      default_accept(:*)
+      defaults([:read, :destroy, create: :*, update: :*])
+    end
+
+    ets do
+      private?(true)
+    end
+
+    attributes do
+      uuid_primary_key(:id)
+
+      attribute(:c, :string, allow_nil?: false, public?: true)
+      attribute(:d, :string, allow_nil?: false, public?: true)
+    end
+  end
+
+  defmodule Maintainer do
+    use Ash.Resource, domain: Domain, data_layer: Ash.DataLayer.Ets
+
+    actions do
+      default_accept(:*)
+      defaults([:read, :destroy, create: :*, update: :*])
+    end
+
+    ets do
+      private?(true)
+    end
+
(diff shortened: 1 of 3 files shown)


zachdaniel:
Awesome! I will review this work early next week 🙇

test/policy/relating_to_actor_via_test.exs
```diff
@@ -0,0 +1,438 @@
...
+
+      assert {:error, %Ash.Error.Forbidden{}} =
+               Reaction
+               |> Ash.Changeset.for_create(:create, %{comment_id: comment.id})
+               |> Ash.create(actor: user)
+    end
+
+    # TODO: how can we make this work?
+    @tag :skip
+    test "relating_to_actor_via with manage_relationship", %{comment: comment, user: user} do
```
zachdaniel, line 381: I don't think there is any good way to make this work past one level. At least not one that isn't super complicated. This is how we do it in the other `relating_to_actor` check:

```elixir
      case changeset.relationships[relationship.name] do
        [{[input], opts}] ->
          primary_key = Ash.Resource.Info.primary_key(relationship.destination)

          actor_keys = take_keys(actor, primary_key)
          input_keys = take_keys(input, primary_key)

          opts[:on_lookup] == :relate and Enum.all?(actor_keys, & &1) &&
            Enum.all?(input_keys, & &1) and input_keys == actor_keys

        _ ->
          false
      end
```

Actually...you could probably do this...it's still really complex though. 

You'd need to:
1. find the input that corresponds to either the attribute backing the relationship
2. *or* find any argument that has a `manage_relationship` change for the given relationship w/ a `on_lookup: :relate`.

You'd have to check this recursively. It's not pretty and also probably not worth our time for now. I'd suggest just adding a check that explicitly shows an error if the first relationship is being managed.

zachdaniel:
I'm interested in this PR, but I think addressing multiple levels deep may be problematic, and there hasn't been recent movement on it. Closing for now, but happy to revisit 🙇

stocks29:
Sorry, was away on vacation for a bit. If I have some time soon I'll try to address the comments.

PR #1311, opened 2024-07-13
Add ash_step wrapper.

Torkan:
Adds a wrapper that handles enqueueing of any returned ash-notifications when running a regular step-module or an anonymous run-function.

### DIFF
--- a/lib/ash/reactor/builders/ash_step.ex
+++ b/lib/ash/reactor/builders/ash_step.ex
@@ -1,0 +1,87 @@
+defimpl Reactor.Dsl.Build, for: Ash.Reactor.Dsl.AshStep do
+  @moduledoc false
+
+  alias Ash.Reactor.AshStep
+  alias Reactor.Builder
+  alias Spark.Error.DslError
+  import Ash.Reactor.BuilderUtils
+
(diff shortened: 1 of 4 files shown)


jimsynz:
Hi @Torkan - just so that I understand - is there a reason that the existing `Ash.Reactor` action steps don't work for you?  Perhaps we should look for a way of retaining notifications as a step output that can be used in subsequent steps?

Torkan:
Hi there, @jimsynz!

In terms of the existing action steps, they work nicely when you just need to fire off a resource action.

Having the opportunity to run steps that are not tied to a specific action, enables the opportunity to for example define steps that contain logic, that can also fire off resource actions if needed.
When steps are inside a transaction, notifications need to be enqueued and published in the reactor's complete phase.

Regarding notifications, I at least haven't found any need to pass them on to later steps, I think the current approach with just having Ash.Reactor handle them automatically for you is great 😅 

So basically this just gives you the opportunity to define a custom Step-module, or an anon run-function, with the opportunity to also return any notifications from any resource actions that was run during the step.

jimsynz (approved):
blowing up

PR #1315, opened 2024-07-17
docs: Set default value for `include_nil?` aggregate option

sevenseacat:
Also updates the docs for the `kind` option to note that it is pre-set for you when using the DSL.

Ideally we would remove the `kind` option from the schema, to hide it from the docs entirely (ie. drop it places like [here](https://github.com/ash-project/ash/blob/main/lib/ash/resource/dsl.ex#L1037). Making those changes works (none of the tests fail), but I don't know what knock-on effect that might have, so I've not put it in this PR.

Screenshot 2024-07-17 at 3 38 52 PM

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Yes, you are totally right. We should hide the `kind` option. I'll merge this and make that change as well.

zachdaniel:

PR #1316, opened 2024-07-17
fix: handle casting nil vectors

hwuethrich:
`Ash.Type.Vector.cast_input/2` does not handle `nil`, which causes the following error when changing a vector to `nil`:

```elixir
change set_attribute(:embedding, nil)
```

Error:

```

* ** (FunctionClauseError) no function clause matching in Ash.Vector.new/1
  (ash 3.1.3) lib/ash/vector.ex:18: Ash.Vector.new(nil)
  (elixir 1.16.2) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
  (elixir 1.16.2) lib/enum.ex:1826: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
  (elixir 1.16.2) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
  (ecto 3.11.2) lib/ecto/repo/queryable.ex:214: Ecto.Repo.Queryable.execute/4
  (ash_postgres 2.1.1) lib/data_layer.ex:1344: AshPostgres.DataLayer.update_query/4
  (ash 3.1.3) lib/ash/actions/update/bulk.ex:505: Ash.Actions.Update.Bulk.do_atomic_update/5
  (ash 3.1.3) lib/ash/actions/update/bulk.ex:260: Ash.Actions.Update.Bulk.run/6
``` 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


hwuethrich:
By the way, the follow test would pass even before the fix because `nil` is handled by `Ash.Type`:

```elixir
assert {:ok, nil} = Ash.Type.cast_input(Ash.Type.Vector, nil)
```

I left the current test as is, but maybe it is better to use `Ash.Type.Vector` directly in all tests?

zachdaniel (approved):

zachdaniel:
`Ash.Type` was incorrectly handling `nil` which should be left to the type. I've removed that behavior, could you update your tests to use `Ash.Type.cast_input`?

zachdaniel:

PR #1328, opened 2024-07-22
fix: remove redundant overload of `Ash.Expr.get_path`

vonagam:
Struct is a map, so no need for same but separate overloads for struct and map.

Also fixed doc for `Ash.Query.Parent`.


zachdaniel (approved):

zachdaniel:

PR #1331, opened 2024-07-23
test: atomic only validation

barnabasJ:
Right now it's not always checked if a validation only has an atomic implementation and it throws an error. 

According to the docs this should be possible? https://hexdocs.pm/ash/dsl-ash-resource.html#validations-validate-always_atomic?

image


zachdaniel:
Although its not possible to do atomic validations on create actions 🤔 But we should raise an error in that case, i.e "this validation only defines an atomic callback, and so it can't be used with create actions"

barnabasJ:
> Although its not possible to do atomic validations on create actions 🤔 But we should raise an error in that case, i.e "this validation only defines an atomic callback, and so it can't be used with create actions"

Alright, so probably just add one action with the validation in the create case and a test that checks if this error occurs.

zachdaniel:

PR #1337, opened 2024-07-24
Add Ash.Type.File

maennchen:
Implementation of ash-project/ash_phoenix#226

## Common Denominator

The file type should allow access to the underlying file with a standardized interface.

I propose that we'll have two functions:

* `path/1` (optional) - Get path of the file if it is on disk. Allows maximum freedom but might not be supported by all implementations.
* `open/2` (required) - Allows access to the file content via the `IO` module. Custom implementations for the [I/O Protocol](https://www.erlang.org/doc/apps/stdlib/io_protocol.html) should allow access to any implementation and is already standardized.

## Storage of Type

I intended this type mainly for action arguments. Those will never be stored.

## Result of this PR

* File type is usable for direct code action access
* `Plug.Upload` works through phoenix controllers
* `Plug.Upload` works through `ash_graphql` if `absinthe` is configured adequately
* Usage in live view is possible but complicated, see https://github.com/ash-project/ash_phoenix/pull/227#issuecomment-2251286221

## Missing right now

* Live View Magic
* `ash_admin` implementation
* `ash_json_api` custom `Plug.Parser` and OpenAPI schema
* There’s no constraints. (file type / size) implemented

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/file.ex
+++ b/lib/ash/type/file.ex
@@ -1,0 +1,123 @@
+defmodule Ash.Type.File do
+  @moduledoc """
+  A type that represents a file on the filesystem.
+
+  > #### Persistence {: .warning}
+  >
+  > This type does not support persisting via `Ash.DataLayer`.
+  >
+  > It is mainly intended to be used in
+  > [arguments](dsl-ash-resource.html#actions-action-argument).
+
+  ## Valid values to cast
+
+  This type can cast multiple types of values:
+
+  * itself
+  * `Plug.Upload`
+  * Any value that implements the `Ash.Type.File.Source` protocol.
+  """
+
+  use Ash.Type
+
+  alias Ash.Type.File.Implementation
+  alias Ash.Type.File.IO, as: IOImplementation
+  alias Ash.Type.File.Path, as: PathImplementation
+  alias Ash.Type.File.Source
+
+  @enforce_keys [:source, :implementation]
+  defstruct [:source, :implementation]
+  @type t() :: %__MODULE__{source: Implementation.source(), implementation: Implementation.t()}
+
+  @impl Ash.Type
+  def storage_type(_constraints), do: :error
+
+  @impl Ash.Type
+  def cast_input(nil, _constraints), do: {:ok, nil}
+  def cast_input(%__MODULE__{} = file, _constraints), do: {:ok, file}
+
+  def cast_input(source, _constraints) do
+    with {:ok, implementation} <- Source.implementation(source) do
+      {:ok, %__MODULE__{source: source, implementation: implementation}}
+    end
+  end
+
+  @impl Ash.Type
+  def cast_stored(_file, _constraints), do: :error
+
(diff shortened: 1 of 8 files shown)


zachdaniel:
The general pattern is to do what you've done and make it always return an error for the storage callbacks. Allowing a file to be stored in a data layer definitely has more complexity than I think we can reasonably encapsulate in this one type. Like "stored how? storing the path? Storing the contents? etc.". I think we can leave that to users to implement in their actions.

The only thing that is up in the air for me is if we should try to make this live in core or if we should put it in a separate package. The benefit of that being we can put it at a 0.x version, mark it as new/experimental and move relatively rapidly.

With that said, this seems like a pretty straightforward solution as well, and so not sure its worth all the rigamarole?

I'm interested also to see how it plays with phoenix & liveview uploads, as that will be the proof in the proverbial pudding.

maennchen:
@zachdaniel If we're not providing a storage implementation, the feature is not very complex. A separate library seems a bit overkill to me in this situation. (However always happy to follow your guidance if you see this differently.)

Since you like the general direction, I'll explore how to integrate this into `ash_phoenix`, `ash_json_api`, `ash_graphql` and `ash_admin`.

maennchen:
Doing this with `ash_json_api` is a bit more involved than what I initially assumed. THis is because we want to send everything besides the file as JSON and not as multipart fields.

We should probably consider doing it similar to [how absinthe handles it](https://hexdocs.pm/absinthe/file-uploads.html). They just put a name of a multipart section into the query.

**Example:**

```http
POST /action
Host: example.com
Content-Length: 2740
Content-Type: multipart/form-data; boundary=abcde12345
--abcde12345
Content-Disposition: form-data; name="request"
Content-Type: application/vnd.api+json

{"users": "users_csv", "meatdata": "metadata_json"}
--abcde12345
Content-Disposition: form-data; name="users_csv"; filename="users.csv"
Content-Type: text/csv

[file content goes here]
--abcde12345
Content-Disposition: form-data; name="metadata_json"; filename="metadata.json"
Content-Type: application/json

[file content goes there]
--abcde12345--
```

zachdaniel:
Yep, I completely agree on the JSON:API approach.

zachdaniel:
@maennchen I'm down to get this merged. This strategy looks good, and the ash_phoenix implementation looks great too. Would be great to get this worked into `AshAdmin` also automatically. Would be very useful.

maennchen:
@zachdaniel I'll explore the live view phoenix part before merging. Just want to make sure it actually works as intended.

zachdaniel:
I think we may eventually want to support constraints on the core file type as well, like allowed file types/max file sizes, even if it's on the caller to enforce those constraints. We don't need to start there though :)

zachdaniel (approved):

zachdaniel:

PR #1346, opened 2024-07-25
fix: pass reuse_values? true when loading in Ash.Query.apply_to/2

rbino:
The function is meant to be used to act on data that is in-memory, so it makes sense to not read data again. Moreover this caused an infinite loop if using this at the end of a manual read.
Add a regression test for the manual read case.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #1355, opened 2024-07-29
fix: set accessing from and join read action for many_to_many relationships correctly

barnabasJ:


zachdaniel (approved):

zachdaniel:

PR #1363, opened 2024-08-03
test: cannot set_attribute on attribute with writable? false

rapidfsub:
I think `writable? false` should stop `set_attribute`, but it doesn't.
Is this a bug?
If not, then how can I prevent `set_attribute`?

(+ What I wrote is a failing test case.)


zachdaniel:
This works as intended. `writable? false` ensures that it can never be provided "externally", but the action can use `set_attribute` to write to it still. `set_attribut` uses `force_change_attribute` under the hood.

Thank you for your PR 🙇

PR #1364, opened 2024-08-05
Add defaults and typespec to resulting Resource struct

m0rt3nlund:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/schema.ex
+++ b/lib/ash/resource/schema.ex
@@ -274,6 +274,8 @@ defmodule Ash.Schema do
             Enum.each(struct_fields, &Module.put_attribute(__MODULE__, struct_fields_name, &1))
           end
 
+          @type t :: %__MODULE__{}
+
           after_compile =
             @after_compile -- [{Ecto.Schema, :__after_compile__}]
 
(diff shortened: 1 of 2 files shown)


lib/ash/resource/schema.ex
```diff
@@ -272,6 +274,8 @@ defmodule Ash.Schema do
             Enum.each(struct_fields, &Module.put_attribute(__MODULE__, struct_fields_name, &1))
           end
 
+          @type t :: %__MODULE__{}
```
zachdaniel, line 277: This will be a breaking changes for users that have done this

zachdaniel: So we can't do this here, unfortunately.

zachdaniel (changes requested):

lib/ash/resource/schema.ex
zachdaniel, line 277:
```suggestion
      unless Kernel.Typespec.defines_type?(__MODULE__, {:t, 0}) do
        @type t :: %__MODULE__{}
      end
```

zachdaniel:

PR #1365, opened 2024-08-06
test: use other resource's calculation in inline filter policy

rapidfsub:
This is just an failing test case.
If you comment out `attribute :is_active, :boolean, default: true, public?: true`(filter_condition_test.exs:124) in Post,
then this test would pass.

Here's what I found. (This could be wrong...)

1. Author has `is_active` attribute.
2. Post has `is_active` attribute, too.
3. Author's `is_deactivated` calculation uses `is_active` in expr.
4. Post's policy uses `Author.is_deactivated`.
With these conditions, `Author.is_deactivated` should check `Author.is_active`,
but it actually checks `Post.is_active`, I think.

### DIFF
--- a/test/policy/filter_condition_test.exs
+++ b/test/policy/filter_condition_test.exs
@@ -60,4 +60,121 @@ defmodule Ash.Test.Policy.FilterConditionTest do
 
     assert visible_resource.visible == true
   end
+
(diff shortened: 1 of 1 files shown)


zachdaniel:

zachdaniel:
Fixed in `main`

PR #1386, opened 2024-08-12
Implement condition for upsert

### LINKED ISSUE #1385: Support a condition for upserts

I'm bulk creating or changing lots of records in an import. The import is run without a transaction and sometimes fails. If I recreate the import, I want to continue where I left of and not touch records that are already up to date.

This is also important so that no new follow up jobs are created which are triggered for every new / changed row.

Postgres allows to set a `WHERE` statement in the `DO UPDATE` block. I want to be able to configure that with an expression.

Example Postgres Query:

```sql
INSERT INTO
  test (name, meta)
  ON CONFLICT (name)
  DO UPDATE
    SET meta = EXCLUDED.meta
    WHERE meta <> EXCLUDED.meta
```

Example Ash Code:

```elixir
Ash.bulk_create(
  [%{name: "Test", meta: %{"foo" => "bar"}]
  Test,
  :create,
  upsert?: true,
  upsert_identity: :test_index,
  upsert_fields: [:meta],
  upsert_condition: expr(meta != excluded.meta)
)
```

I tried the following code which seems to almost work.

```elixir
create :create do
  change fn changeset, _ -> 
    Ash.Changeset.filter(changeset, expr(meta != fragment("EXCLUDED.meta")))
  end
end
```

This code works with `Ash.bulk_create`, but fails with the following error on a normal `Ash.create`:

```

     * %CaseClauseError{term: {:ok, []}}
       (ash_postgres 2.1.18) lib/data_layer.ex:2533: AshPostgres.DataLayer.upsert/4
       (ash 3.3.3) lib/ash/actions/create/create.ex:352: anonymous fn/6 in Ash.Actions.Create.commit/3
       (ash 3.3.3) lib/ash/changeset/changeset.ex:3621: Ash.Changeset.run_around_actions/2
       (ash 3.3.3) lib/ash/changeset/changeset.ex:3162: anonymous fn/3 in Ash.Changeset.with_hooks/3
       (ecto_sql 3.11.3) lib/ecto/adapters/sql.ex:1358: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
       (db_connection 2.7.0) lib/db_connection.ex:1756: DBConnection.run_transaction/4
       (ash 3.3.3) lib/ash/changeset/changeset.ex:3160: anonymous fn/3 in Ash.Changeset.with_hooks/3
       (ash 3.3.3) lib/ash/changeset/changeset.ex:3304: anonymous fn/2 in Ash.Changeset.transaction_hooks/2
       (ash 3.3.3) lib/ash/changeset/changeset.ex:3141: Ash.Changeset.with_hooks/3
       (ash 3.3.3) lib/ash/actions/create/create.ex:236: Ash.Actions.Create.commit/3
       (ash 3.3.3) lib/ash/actions/create/create.ex:108: Ash.Actions.Create.do_run/4
       (ash 3.3.3) lib/ash/actions/create/create.ex:46: Ash.Actions.Create.run/4
```

Usage via function parameters:

```elixir
Ash.bulk_create(
  [%{name: "Test", meta: %{"foo" => "bar"}]
  Test,
  :create,
  upsert?: true,
  upsert_identity: :test_index,
  upsert_fields: [:meta],
  upsert_condition: expr(meta != excluded.meta)
)
```

```elixir
actions do
  create :upsert do
    upsert? true
    upsert_fields [:meta]
    upsert_condition expr(meta != excluded.meta)
  end
end
```

**Additional context**

None

maennchen:
Resolves #1385

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -293,6 +293,11 @@ defmodule Ash do
                              ]},
                           doc:
                             "The fields to upsert. If not set, the action's upsert_fields is used, and if that is not set, then any fields not being set to defaults are written."
+                        ],
+                        upsert_condition: [
+                          type: :any,
+                          doc:
+                            "An expression to check if the record should be updated when there's a conflict."
                         ]
                       ]
                       |> Spark.Options.merge(@global_opts, "Global Options")
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -3093,6 +3093,13 @@ defmodule Ash.Filter do
     do_hydrate_refs(%Ash.Query.Parent{expr: arg}, context)
   end
 
+  defp resolve_call(
+         %Call{name: :upsert_conflict, args: [arg], relationship_path: []},
+         context
+       ) do
+    do_hydrate_refs(Ash.Query.UpsertConflict.new(arg), context)
+  end
+
   defp resolve_call(%Call{name: name, args: args} = call, context)
        when name in @inline_aggregates do
     resource = Ash.Resource.Info.related(context.resource, call.relationship_path)
@@ -3663,6 +3670,25 @@ defmodule Ash.Filter do
       end
     end
   end
+
+  def do_hydrate_refs(
+        %Ash.Query.UpsertConflict{attribute: attribute} = this,
+        %{conflicting_upsert_values: %{} = resource} = context
+      ) do
+    context
+    |> Map.merge(%{resource: resource, relationship_path: []})
+    |> attribute(attribute)
+    |> case do
+      %Ash.Resource.Attribute{name: ^attribute} ->
+        {:ok, Map.fetch!(resource, attribute)}
+
+      nil ->
+        {:error, "Invalid reference #{inspect(this)}"}
+    end
+  end
+
+  def do_hydrate_refs(%Ash.Query.UpsertConflict{} = this, _context),
+    do: {:ok, this}
 
   def do_hydrate_refs(
         %Ash.Query.Exists{expr: expr, at_path: at_path, path: path} = exists,
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -213,6 +213,7 @@ spark_locals_without_parens = [
   update_timestamp: 1,
   update_timestamp: 2,
   upsert?: 1,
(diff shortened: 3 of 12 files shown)


lib/ash.ex
```diff
@@ -293,6 +293,12 @@ defmodule Ash do
                              ]},
                           doc:
                             "The fields to upsert. If not set, the action's upsert_fields is used, and if that is not set, then any fields not being set to defaults are written."
+                        ],
+                        upsert_condition: [
+                          # TODO: Better type?
```
zachdaniel, line 298: Unfortunately, `:any` is all we can really do. At some point in the future we will refactor expressions to live in `%Ash.Expr{}` so we can expect that struct, and make pattern matching much better in general.

lib/ash/filter/filter.ex
```diff
@@ -3664,6 +3671,25 @@ defmodule Ash.Filter do
     end
   end
 
+  def do_hydrate_refs(
```
zachdaniel, line 3674: hydrating refs can happen before the context we need is present. I think here needs to leave the `%Ash.Query.UpsertConflict{}` in place, and instead this logic should go in `filter/runtime.ex`

maennchen: @zachdaniel Yes, it already hydrates without the context in `Ash.Changeset.filter/2`.

This is why there's two function clauses defined.

My idea was the following:

* If the context is not present, leave the expression as is
* If the context is present, the adapter is setting the relevant info and we can hydrate it directly
* If the filter is run when the adapter didn't override the expression (as seen in the postgres PR) and does not provide the context (as with `ETS`), the call will land in runtime
* When reaching runtime, it will error that the datasource does not support the expression.

maennchen: I'm taking this back. When we use `filter_matches`, it works differently.

maennchen: I'm taking back the taking back :laughing: I just did it wrong.

zachdaniel:

PR #1389, opened 2024-08-14
Implement `upsert_condition` for bulk create

maennchen:
Missed a few spots in #1385 

# Contributor checklist

- [X] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [X] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

PR #1390, opened 2024-08-15
Emit Batch Status for Bulk Actions

### LINKED ISSUE #1367: Possibility to emit indices in bulk action with `return_stream?: true`

I'm writing a Mix task, which imports a large file via a bulk insert. To display progress, I would like to be able to log a message every n entries so that I can see the progress. TO do this, I have to enable `return_records?` at the moment. This uses memory unnecessarily.

A new option `emit_indices?` which will return the current index instead of the whole record. Only available when `return_stream?: true` and `return_records?: false`.

None

For example

```elixir
rows
|> Ash.bulk_create(Resource, :action, return_stream?: true, emit_indices?: true)
|> Enum.each(fn
  {:omitted_record, i} when mod(i, 100) -> IO.puts "imported #{i} records"
  _ -> :discard
end)
```

**Additional context**

None

maennchen:
Resolves #1367
Based on #1389 

# Scope

* Introduces the option `return_batch_status?` to all `Ash.bulk_*` operations.
* Adds `record_count` to `Ash.BulkResult`
* Adds `Ash.BulkResult.BatchStatus` and yields it in the result stream when enabled

# TODO

* [x] Support bulk_create
* [ ] Support bulk_update with list
* [ ] Support bulk_update with query
* [ ] Support bulk_destroy with list
* [ ] Support bulk_destroy with query

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -178,15 +178,18 @@ defmodule Ash.Actions.Create.Bulk do
       )
 
     if opts[:return_stream?] do
-      Stream.concat(changeset_stream)
+      changeset_stream
+      |> Stream.concat()
+      |> Ash.BulkResult.BatchStatus.accumulate()
     else
       try do
-        records =
+        {records, record_count} =
           if opts[:return_records?] do
-            Enum.to_list(Stream.concat(changeset_stream))
+            changeset_stream
+            |> Stream.concat()
+            |> Enum.map_reduce(0, fn element, count -> {element, count + 1} end)
           else
-            Stream.run(changeset_stream)
-            nil
+            {nil, changeset_stream |> Stream.concat() |> Enum.count()}
           end
 
         notifications =
@@ -204,6 +207,7 @@ defmodule Ash.Actions.Create.Bulk do
 
         bulk_result = %Ash.BulkResult{
           records: records,
+          record_count: record_count,
           errors: errors,
           notifications: notifications,
           error_count: error_count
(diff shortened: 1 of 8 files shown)


maennchen:
@zachdaniel Is this going in the direction you intended?

lib/ash/actions/create/bulk.ex
```diff
@@ -178,15 +178,17 @@ defmodule Ash.Actions.Create.Bulk do
...
+      |> Stream.concat()
+      |> Ash.BulkResult.BatchStatus.accumulate()
     else
       try do
-        records =
+        {records, record_count} =
           if opts[:return_records?] do
-            Enum.to_list(Stream.concat(changeset_stream))
+            records = Enum.to_list(Stream.concat(changeset_stream))
+            {records, length(records)}
```
zachdaniel, line 189: We should count and convert to  list in a single reduce here I think.

lib/ash/actions/create/bulk.ex
```diff
@@ -1216,12 +1275,38 @@ defmodule Ash.Actions.Create.Bulk do
...
 
             :ok ->
               Process.put({:any_success?, ref}, true)
+
+              cond do
+                opts[:return_records?] ->
+                  raise "`#{inspect(Ash.Resource.Info.data_layer(resource))}.bulk_create/3` returned :ok without a result when `return_records?` is true"
+
+                opts[:return_batch_status?] ->
+                  raise "`#{inspect(Ash.Resource.Info.data_layer(resource))}.bulk_create/3` returned :ok without affected row count when `return_batch_status?` is true"
```
zachdaniel, line 1304: We should synthesize this return type for calls to single creates.

zachdaniel (dismissed):

zachdaniel:
I know this is still WIP but did an initial review. This looks like it's going in the right direction to me 👌

lib/ash/actions/create/bulk.ex
maennchen: As I understand the current code, it is fully synthesized. The error is only raised if you implement `bulk_create` and return `:ok` from there.

Am I missing something?

That fn could use a refactor, it's quite hard to follow what it does :sweat_smile:

zachdaniel: Ah, right because that callback returns `{:ok, result}`.

maennchen:
Closing since I have no time soon to finish this and the attached refactorings required to make it happen.

PR #1400, opened 2024-08-21
fix: do not call tracer `set_metadata` with span type that it ignores

vonagam:
Adjacent note: I think `Ash.Tracer` methods `start_span`, `stop_span` and `set_error` should be marked with `@doc false` since they are not supposed to be part of public API and are public just so that they can be called from a macro.


zachdaniel:
Agreed on `@doc false`

zachdaniel (approved):

zachdaniel:

Issue #1407, opened 2024-08-26
`manage_relationship` should warn if the given options is empty


drtheuns:
**Is your feature request related to a problem? Please describe.**
If `manage_relationship` is called without opts it will do nothing. Ash could warn about this.

**Describe the solution you'd like**
Add a warning if `manage_relationship` is a no-op.

**Describe alternatives you've considered**
Remove the default value of the `opts` argument; likely not possible due to backwards compat.

**Express the feature either with a change to resource syntax, or with a change to the resource interface**
```
Ash.Changeset.manage_relationship(changeset, relationship, value)
# warning: this does nothing. Needs `type: :append` or some other configuration.
```

**Additional context**
-

PR #1408, opened 2024-08-26
Warn when `manage_relationship` is called without opts

drtheuns:
Should it perhaps be implemented as a separate clause to avoid doing any needless work? E.g.:

```
def manage_relationship(changeset, _, _) do
  IO.warn("...")
  changeset
end
```

Fixes:  #1407


zachdaniel:

PR #1416, opened 2024-08-28
add pattern for Ash.Query.Call in Filter.map

pinetops:
Adding specific handling for Ash.Query.Call in Filter.map to avoid unnecessary recursion into the internals of the struct.


zachdaniel:

PR #1417, opened 2024-08-29
Improve caching for relationships and add caching for calculations

pinetops:
Two performance optimizations:
- avoid falling back to a search on nil, which happens quite a lot because we often look up whether non relationships are relationships.
- add similar caching for calculations

But, it seems there's quite a lot of cases where we look up things other than string or atom keys. That seems like it's probably not the right thing to be doing.

In all these cases, in the current implementation we would be comparing relationship name with things like Attribute structs, calculation modules, Ash.Query. All of which would never match a string/atom name.

To get all the tests to pass without introducing a performance issue I add:
 def calculation(_resource, _relationship_name) do: nil
 def relationship(_resource, _relationship_name) do: nil

But I'm rather assuming that these are all examples of errors, and we should fix the root causes. I did have a go, but the question about where in the process it should be corrected was a bit too hard for me. It's not clear this is more broken than the pre-existing code though.

### DIFF
--- a/lib/ash/resource/dsl.ex
+++ b/lib/ash/resource/dsl.ex
@@ -1485,6 +1485,7 @@ defmodule Ash.Resource.Dsl do
 
   @persisters [
     Ash.Resource.Transformers.CacheRelationships,
+    Ash.Resource.Transformers.CacheCalculations,
     Ash.Resource.Transformers.AttributesByName,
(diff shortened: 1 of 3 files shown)


zachdaniel (approved):

zachdaniel:

PR #1419, opened 2024-08-29
fix: correctly handle return values of batch callbacks

hwuethrich:
Follow-up on #1415 that handles the following cases in the generated `change/3` function:

- `before_batch/3` returning list of `%Ash.Changeset{}`
- `after_batch/3` returning `:ok`

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/change/change.ex
+++ b/lib/ash/resource/change/change.ex
@@ -214,8 +214,15 @@ defmodule Ash.Resource.Change do
                 {[changeset], notifications} =
                   Enum.split_with(
                     apply(__MODULE__, :before_batch, [[changeset], opts, context]),
-                    fn %Ash.Notifier.Notification{} ->
-                      false
+                    fn
+                      %Ash.Notifier.Notification{} ->
+                        false
+
+                      %Ash.Changeset{} ->
+                        true
+
+                      other ->
+                        raise "Expected before_batch/3 to return a list of changesets and notifications, got: #{inspect(other)}"
                     end
                   )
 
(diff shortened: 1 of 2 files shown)


hwuethrich:
I get dialyzer warnings now in our project which seem to be caused by these changes:

```
________________________________________________________________________________
lib/myproject/records/record/changes/set_imported_after_action.ex:1:guard_fail
The guard clause:

when _ :: true === nil

can never succeed.
```

Will try to fix those before merging

hwuethrich:
@zachdaniel When I use `location: :keep` in the `__before_compile__` macro, I can see that the warnings are generated for the `has_before_batch?()` and `has_after_batch?()` calls:

```
/Users/hw/Projects/ash/lib/ash/resource/change/change.ex:212:guard_fail
The guard clause:

when _ :: true === nil

can never succeed.

________________________________________________________________________________
/Users/hw/Projects/ash/lib/ash/resource/change/change.ex:238:guard_fail
The guard clause:

when _ :: true === nil

can never succeed.

________________________________________________________________________________
done (warnings were emitted)
```

I don't unterstand why this warning is triggered (or how to suppress it), but currently this causes dialyzer warnings for everybody who is using the generated `change/3` function (also in the latest release, not just my PR).

Any ideas?

zachdaniel:
I've merged this commit, but in another PR w/ another commit: #1424 

For some reason my dialyzer isn't producing the failure you're mentioning locally, but I've seen it before so I know that it would arise in this situation. I've switched the code branching to be at compile time instead of runtime which will be a slight optimization and will also resolve this issue. You can try your dialyzer against `main` now and see if this has the desired effect.

hwuethrich:
@zachdaniel Thanks, the warnings are gone now! 👍

PR #1425, opened 2024-08-31
Remove cache fallback for Info.attribute

pinetops:
This catches a performance impact when looking up things that aren't attributes.

Also change a Transformer that runs before the attribute cache so it now doesn't depend on Info.attribute

### DIFF
--- a/lib/ash/resource/info.ex
+++ b/lib/ash/resource/info.ex
@@ -321,7 +321,15 @@ defmodule Ash.Resource.Info do
 
   def relationship(resource, relationship_name)
       when is_binary(relationship_name) or is_atom(relationship_name) do
-    Extension.get_persisted(resource, :relationships_by_name)[relationship_name]
+    case Extension.get_persisted(resource, :relationships_by_name) do
+      value when is_map(value) ->
+        value[relationship_name]
+
+      nil ->
+        resource
+        |> relationships()
+        |> Enum.find(&(to_string(&1.name) == to_string(relationship_name)))
+    end
   end
 
   def relationship(_, _), do: nil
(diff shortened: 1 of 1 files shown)


lib/ash/resource/info.ex
```diff
@@ -708,19 +708,13 @@ defmodule Ash.Resource.Info do
...
+    Extension.get_persisted(resource, :attributes_by_name)[name]
   end
 
-  def attribute(resource, name) do
-    Extension.get_persisted(resource, :attributes_by_name)[name] ||
-      resource
-      |> attributes()
-      |> Enum.find(&(&1.name == name))
-  end
+  # def attribute(resource, name) do
```
zachdaniel, line 715: 👻

pinetops: whoops

zachdaniel:
So, we can't be sure that others aren't using these functions before they are cached. However, I think what we can do is something like this:

```elixir
case Extension.get_persisted(resource, :attributes_by_name) do
  nil -> # old behaviour
  value when is_map(value) ->
    value[name]
end
```

that way if its been cached it will be used, and we never fall back to the traversal if we have the cached map.

pinetops:
Makes sense. Presumably we should do the same for relationships and calculations?

Also this was one a pretty clean change, but in doing the relationships and calculations it exposed that there's a bunch of cases in the test suite where we end up passing a bunch other things to these functions (Attribute structs, query structs, calculation structs) and wondered if they all represented mistakes. That's why there's the not an atom, not a binary version in those cases.

Is it desired behavior to return nill if we get something that isn't an atom or struct or should we update the callers to avoid doing that?

zachdaniel:
Yes, if we get something that is not an atom or a binary we should return `nil`, no need to actually search for a match 👍 And things passing structs to that list is very likely a bug.

pinetops:
Ok, here's a choice:
- first commit only, breaks zero tests
- second commit, breaks 66 :-)

zachdaniel:
lets stick with the fallbacks for now 😂 Please open an issue for resolving the remaining calls. In reality we probably want the fallbacks since those functions are public apis anyway.

pinetops:
Done and created bug:
https://github.com/ash-project/ash/issues/1426

zachdaniel (approved):

zachdaniel:

PR #1427, opened 2024-09-01
Add a case for handling mapsets in Filter.map

pinetops:
Filter.map is often called with data structures that include mapsets so we should handle those directly without unpacking the structs.

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -1165,6 +1165,9 @@ defmodule Ash.Filter do
       value when is_list(value) ->
         Enum.map(value, &map(&1, func))
 
+      %MapSet{} = value ->
+        MapSet.new(value, &map(&1, func))
+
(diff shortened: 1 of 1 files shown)


lib/ash/filter/filter.ex
```diff
@@ -1165,6 +1165,12 @@ defmodule Ash.Filter do
       value when is_list(value) ->
         Enum.map(value, &map(&1, func))
 
+      %MapSet{} = value ->
```
zachdaniel, line 1168: Lets use `MapSet.new(value, &map(&1, func))` to avoid many traversals.

pinetops:
Done.

But I'm wondering to what extent we apply func twice unecessarily. If any of our other structures contain a leaf node, we call func on it in map, but then we'll also apply map again in the other case. 

Does that make sense?

zachdaniel:

PR #1450, opened 2024-09-10
Add relationship editor to Resource Igniter

sinisijacob:
I found the need to add a many_to_many relationship to a Resource generated in an Igniter, and could not find a good way to do that. Rather than add some custom zipper work in the Igniter, I added a relationships editor to the Resource Igniter. I have tested this locally, but if there are tests that should be added please let me know.

If there is a better way that Ash already supports to do this I'd be happy to do it that way instead, but I couldn't find one.


zachdaniel (approved):

zachdaniel:

PR #1452, opened 2024-09-11
Handle Ash.Query.filter for array values

maennchen:
# Contributor checklist

- [x] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

## Background

I have a resource with a postgres ltree as primary key, which is represented as a list of strings.

The `InvalidFilterValue` bubbles up from a `manage_relationship` change.

PR #1455, opened 2024-09-11
Fix Ash.Changeset.manage_relationships/4 for list primary keys

maennchen:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Addresses same issue as #1452 

It's quite complicated to produce a good test for this. I would need two resources, one of which has an array primary key and a many to many relationship.

Should I build the test or is the PR ok like this?


zachdaniel:

PR #1460, opened 2024-09-13
Moved common block editing code into single function, repointed existing block editors to use that as well

sinisijacob:
There was some boilerplate code for the various section editors, so when writing the new add_resource_configuration I decided to move that logic into a common function and point the others to it with the block arg filled in.

This may also be applicable to add_bypass and add_policy, but since that seemed like a bit larger of a refactor I've left it for now.

Feedback appreciated.


zachdaniel:

PR #1476, opened 2024-09-25
fix: set right defaults for `action_select`

vonagam:
`action_select` on a action struct should default to nil instead of an empty list.

Because currently it is an empty list [this check](https://github.com/ash-project/ash/blob/e25f9c6d71a6d53711dc49ea9f080cc78648f903/lib/ash/resource/transformers/cache_action_inputs.ex#L49) in `Ash.Resource.Transformers.CacheActionInputs` always resolves to true and `action_select` ends up being set to `[:id]` for most actions.

Also tests revealed that there is a problem with [this](https://github.com/ash-project/ash/blob/e25f9c6d71a6d53711dc49ea9f080cc78648f903/lib/ash/resource/transformers/cache_action_inputs.ex#L61) in the false branch (that was never used before so was not detected). Codebase expects `Resource.Info.action_select()` to return a list and breaks when encounters nil that is set here when changeset has no changes. Instead of nil set it now to an empty list.


zachdaniel:

zachdaniel:

vonagam:
Two notes worth pointing out:

1) When attributes were not selected they were present as `nil`. Shouldn't they be `NotLoaded` in such cases?

2) Less important - documentation lists `action_select` for `read` and `action` actions even though it should not apply to them.

zachdaniel:
👍 will address both, thank you :)

zachdaniel:
Actually, the values that weren't selected will be `nil` during hooks only, then the select-hiding is applied with `%Ash.NotLoaded{}`. So I'm wondering if it is worth the expense to scrub the values. We won't be able to guarantee that the values you didn't specify in `action_select` won't be loaded, because the caller may have selected them. So you might see `some_field: 10` even though you didn't specify it in action select.

zachdaniel:
Point is, there is some inconsistency anyway...nah, nvm. We should set the fields to `%Ash.NotLoaded{}` :)

vonagam:
Hm... well part of the problem was that fields other than attributes were `NotLoaded`. Because of that I initially thought that I've set attributes wrong or they were wrote incorrectly.

Also `NotLoaded` should make itself known pretty fast if used while `nil` has `Access` protocol and is accepted as a valid thing in many places so you might not know that something is not selected and it appears to be working but then mysterious issues appear.

zachdaniel:
Fixed

PR #1478, opened 2024-09-25
docs: correct return type for `around_action`

vonagam:
Question: it is not mentioned but `around_action` is not compatible with atomic (as in, those hooks are not executed if changeset is atomic), correct?

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -5347,7 +5347,7 @@ defmodule Ash.Changeset do
   Adds an around_action hook to the changeset.
 
   Your function will get the changeset, and a callback that must be called with a changeset (that may be modified).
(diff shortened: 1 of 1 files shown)


zachdaniel:
The hooks shouldn't be ignored, they will just cause the action not to be able to be run atomically. So you'd get an error if you tried to run it with `require_atomic?: true`, or in some way that requires atomics, like `strategy: :atomic` with bulk actions.

zachdaniel:

vonagam:
Hm... I see so it is not supported. For record I've had a custom change that had `def atomic` that was returning `{:ok, change(changeset, ...)}` and in that `change()` I was setting `around_action` and `after_action`. And it was trying to run changeset skipping `around_action` and throwing in `after_action` (that depends on stuff from around action).

zachdaniel:
Ah, interesting...yeah, probably a last-minute check that no atomic callbacks have added those hooks is necessary.

PR #1515, opened 2024-10-14
improvement!: Require `transaction? false` in actions where the data layer doesn't support transactions.

jimsynz:
When using a data layer that doesn't support transactions, it's a good idea to make them add `transaction? false` to their actions.

Default actions are generated with the correct transaction option.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/data_layer/data_layer.ex
+++ b/lib/ash/data_layer/data_layer.ex
@@ -101,6 +101,7 @@ defmodule Ash.DataLayer do
           | {:sort, Ash.Type.t()}
           | :upsert
           | :composite_primary_key
+          | :inhibit_transaction_validation
 
   @type lateral_join_link ::
           {Ash.Resource.t(), atom, atom, Ash.Resource.Relationships.relationship()}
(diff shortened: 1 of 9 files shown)


zachdaniel:
This is going to require making changes to pretty much all tests on all projects. It's a lot of work. I'd rather make explicit carve outs for simple and ets data layers to avoid the massive churn.

zachdaniel:
I'm happy to make the change at some point, maybe on a major version change or at another time, but its just too much work to take on at this point.

zachdaniel:
Also, if we're doing this as an actual breaking change then its going to be ~a year before this gets merged. should probably make this opt in based on the data layer.

jimsynz:
@zachdaniel in the end I decided to add a capability to allow datalayers to opt out of the validation. works just as well and allows it to be stubbed out in tests if needed.

zachdaniel:
hmm...it's still breaking unless the default is opt out, as other people have custom data layers.

zachdaniel:
i'd suggest just making it a data layer callback, since it's an option not a capability.

jimsynz:
So I've been thinking about this and I think what we should do is just put the verifier directly on the sqlite extension and leave it out of core until 4.0.

PR #1522, opened 2024-10-16
feat: ash seed upsert! function

Sangresz:
Ash.Seed.upsert! function to create or update resources based on the specified identity

PR opened regarding the problem listed on [this](https://elixirforum.com/t/how-would-i-call-directly-the-datalayer-bypassing-the-action/66738) post on elixirforum. 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/seed.ex
+++ b/lib/ash/seed.ex
@@ -151,6 +151,8 @@ defmodule Ash.Seed do
   Ash.Seed.upsert!(%User{email: 'test@gmail.com', name: 'Test'}, [:email])
   ```
   """
+  def upsert!(_, identity \\ nil)
+
   def upsert!(%{__meta__: %{state: :loaded}} = input, _identities) do
     input
   end
@@ -178,7 +180,7 @@ defmodule Ash.Seed do
           Map.put(acc, key, value)
       end)
 
-    upsert!(
+    upsert_with_opts!(
       resource,
       input,
       identities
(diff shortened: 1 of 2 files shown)


lib/ash/seed.ex
```diff
@@ -143,6 +143,134 @@ defmodule Ash.Seed do
...
+
+  ```elixir
+  Ash.Seed.upsert!(%User{email: 'test@gmail.com', name: 'Test'}, [:email])
+  ```
+  """
+  def upsert!(%{__meta__: %{state: :loaded}} = input, _identities) do
+    input
+  end
+
+  def upsert!(%resource{} = input, identities) do
```
zachdaniel, line 158: What should be put here is an identity name, or nil (meaning upsert via primary key). So like `upsert!(_, identity \\ nil)`

zachdaniel: Then we look up the identity fields prior to upsert.

Sangresz:
I added the defaults but i needed to change the name because it was producing an arity error based on the defaults. Maybe its better to have the two functions have the same name and people specify "nil" when they want to use the second method?

Sangresz:
Alright! i didn't know the existence of that module so i didn't understand in the previous attempt.
With the last merge you pass the identity as an option. Unfortunately the arity problem still persist because having
```elixir
def upsert!(_, opts \\ [])
``` 
goes in conflict with
```elixir
def upsert!(resource, input, opts \\ [])
```
because you can't know if the 2nd parameter that you pass is the _opts_ for the first function or the _input_ for the second one

Sangresz:
@zachdaniel Done!

zachdaniel (approved):

zachdaniel:

PR #1538, opened 2024-10-21
fix: check nil before calling Decimal.eq? in Type.Decimal.equal?

rapidfsub:
simple bug fix


zachdaniel:

PR #1548, opened 2024-10-22
testing: prove that calculations can return a list of embedded resources

lukasender:
# What it does

Prove that Ash can return a list of embedded resources via a calculation.

# Context

We thought we encountered a bug.

Turns out, it was just me doing something wrong.

This test serves as the result and prove that what I tried to do indeed works.


zachdaniel:

PR #1569, opened 2024-10-31
fix: add missing case clause for `Changeset.atomic_update` type cast

vonagam:


zachdaniel:

PR #1575, opened 2024-11-04
fix example of using string_length

allenwyma:
Simple fix to docs to reflect actual api, where if you want to check the length of a string is exactly X, you should use the keyword `exact:`.


zachdaniel (approved):

zachdaniel:

PR #1584, opened 2024-11-05
fix: properly handle `changing_attributes` check with `from` in create

vonagam:
This piece of code in `changing_attributes` check does not make sense to me right now:
```elixir
if Keyword.has_key?(opts, :from) && changeset.action_type == :create do
  if opts[:from] == nil do
    {:halt, false}
  else
    {:cont, expr}
  end
else
```

It says that `from: nil` is bad but `from: :whatever` is good.

First, I assume that the branches were supposed to be reversed - if `from: nil` then it matches create otherwise not.

Second, but that supposed reversal is also bad as it does not take into account attribute's default value (so it is questionable that the change is from nil) and also ignores potential presence of `to` option.

So instead just do return false for create changeset if `from` option is present.


zachdaniel (approved):

zachdaniel:

PR #1586, opened 2024-11-06
fix: use private context of query to populate context

barnabasJ:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


barnabasJ:
It didn't feel quite right using the private context like this. But the code is buried pretty deep in the action live cycle so I wasn't sure what the best way to get to the data was.

jimsynz (approved):
nope

zachdaniel:
What was the original issue this was added to solve?

PR #1601, opened 2024-11-15
fix: finish conversion from `append?` to `prepend?` option

vonagam:
Since Ash 4 changed order of execution for before hooks there is no more `append?` option - it was replaced with `prepend?`. But built-in helpers (`before_action` and `before_transaction`) and some other places are still referencing and specifying `append?`.


zachdaniel (approved):

zachdaniel:

PR #1613, opened 2024-11-24
don't document options helpers but add typespecs

allenwyma:
# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/options_helpers.ex
+++ b/lib/ash/options_helpers.ex
@@ -3,20 +3,29 @@ defmodule Ash.OptionsHelpers do
   Helpers for working with nimble options
   """
 
-  @doc false
+  @doc """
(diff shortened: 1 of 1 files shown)


zachdaniel:
making them `@doc false` essentially declares them as private, so we should leave that change out.

allenwyma:
> making them `@doc false` essentially declares them as private, so we should leave that change out.

These seem quite specific for a reason and not really usable for most outside of those needing to use them. I was thinking for their specific use they don’t need to be documented.

if they aren’t doc false then I’d like to have some docs on them. I can try to add what I understand m them to be if you or someone else can touch them up after that’d be great.

zachdaniel (approved):

zachdaniel:

PR #1620, opened 2024-11-26
Add equal_to, not_equal_to to Validation.Compare

jechol:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/validation/compare.ex
+++ b/lib/ash/resource/validation/compare.ex
@@ -7,10 +7,20 @@ defmodule Ash.Resource.Validation.Compare do
   import Ash.Expr
 
   @opt_schema [
+    equal_to: [
+      type: {:or, [:integer, :atom, :float, {:struct, Decimal}, {:fun, 0}]},
+      required: false,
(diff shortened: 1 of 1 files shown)


zachdaniel:
Hmm...we currently have `validate attribute_equals(:foo, :bar)` and `validate attribute_does_not_equal`, I think it would make sense to use those instead?

jechol:
> Hmm...we currently have `validate attribute_equals(:foo, :bar)` and `validate attribute_does_not_equal`, I think it would make sense to use those instead?

1. `attribute_equals` doesn't work for Decimal when the precision differs.
2. I prefer to consistently use `compare` rather than mixing `attribute_equals` and `compare` for Decimal.

zachdaniel:
I think the fix is to use `Comp.equal?` in `attribute_equals` and `attribute_does_not_equal` in that case then.

jechol:
Oh, I see. Should I go ahead and create a PR for that?

PR #1624, opened 2024-11-27
fix: negate atomic validation expressions when used as `where` conditions

### LINKED ISSUE #1619: `where` doesn't seem to work properly on atomics
**Describe the bug**
It appears that `where` is not working in conjunction a custom atomic validator. In this particular situation I am checking on the value of an aggregate like so:

```elixir
defmodule Is100Percent do
  use Ash.Resource.Validation

  def atomic(_changeset, _opts, context) do
    {
      :atomic,
      [:total_percentage],
      expr(^atomic_ref(:total_percentage) != 100),
      expr(
        error(^Ash.Error.Changes.InvalidAttribute, %{
          field: :total_percentage,
          message: ^(context.message || "must equal 100%"),
          value: ^atomic_ref(:total_percentage)
        })
      )
    }
  end
end
```

And here's the aggregate (close enough anyway, I'm being a bit cagy with names):

```elixir
aggregates do
  sum :total_percentage, :items, :percentage do
    default Decimal.new(0)
    public? true
  end
end
```

The validation works fine like this:

```elixir
update :update do
  primary? true
  accept [:status]

  validate Is100Percent
end
```

However, not like this:

```elixir
update :update do
  primary? true
  accept [:status]

  validate Is100Percent do
    where attribute_equals(:status, "active")
  end
end
```

The idea is I only want to validate this when that status matches.

Here's hopefully helpful SQL:

```elixir
Ash.update(resource, %{status: "active"}, authorize?: false)
```

This _should_ return an error but instead, the status is updated.

```sql
UPDATE "resources" AS a0 SET "id" = s1."__new_id", "status" = s1."__new_status" FROM (SELECT $1::varchar::varchar AS "__new_status", (CASE WHEN ($2::varchar::varchar != $3::varchar) AND (coalesce(ss1."total_percentage", $4::decimal)::decimal != $5::decimal) THEN ash_raise_error(jsonb_build_object('exception', $6::text, 'input', jsonb_build_object($7::text, $8::jsonb, $9::text, coalesce(ss1."total_percentage", $10::decimal)::decimal, $11::text, $12::jsonb))::jsonb, NULL::uuid) ELSE sa0."id"::uuid END)::uuid AS "__new_id", sa0."id" AS "id" FROM "resources" AS sa0 LEFT OUTER JOIN LATERAL (SELECT ssa0."resource_id" AS "resource_id", coalesce(sum(ssa0."percentage"::decimal), $13::decimal)::decimal AS "total_percentage" FROM "public"."itms" AS ssa0 WHERE (sa0."id" = ssa0."resource_id") GROUP BY ssa0."resource_id") AS ss1 ON TRUE WHERE (sa0."id"::uuid = $14::uuid)) AS s1 WHERE (a0."id" = s1."id") RETURNING a0."id", a0."created_at", a0."status" ["active", "active", "active", Decimal.new("0"), Decimal.new("100"), "Ash.Error.Changes.InvalidAttribute", "message", "must equal 100%", "value", Decimal.new("0"), "field", "total_percentage", Decimal.new("0"), "01936690-1a73-7f84-9eca-b7292d788a42"]
```

If I remove `when`, it works:

```sql
UPDATE "resources" AS a0 SET "id" = s1."__new_id", "status" = s1."__new_status" FROM (SELECT $1::varchar::varchar AS "__new_status", (CASE WHEN coalesce(ss1."total_percentage", $2::decimal)::decimal != $3::decimal THEN ash_raise_error(jsonb_build_object('exception', $4::text, 'input', jsonb_build_object($5::text, $6::jsonb, $7::text, coalesce(ss1."total_percentage", $8::decimal)::decimal, $9::text, $10::jsonb))::jsonb, NULL::uuid) ELSE sa0."id"::uuid END)::uuid AS "__new_id", sa0."id" AS "id" FROM "resources" AS sa0 LEFT OUTER JOIN LATERAL (SELECT ssa0."resource_id" AS "resource_id", coalesce(sum(ssa0."percentage"::decimal), $11::decimal)::decimal AS "total_percentage" FROM "public"."itms" AS ssa0 WHERE (sa0."id" = ssa0."resource_id") GROUP BY ssa0."resource_id") AS ss1 ON TRUE WHERE (sa0."id"::uuid = $12::uuid)) AS s1 WHERE (a0."id" = s1."id") RETURNING a0."id", a0."created_at", a0."status" ["active", Decimal.new("0"), Decimal.new("100"), "Ash.Error.Changes.InvalidAttribute", "message", "must equal 100%", "value", Decimal.new("0"), "field", "total_percentage", Decimal.new("0"), "01936694-4b0d-74cc-93f0-b7edbf883b3e"]
```

It looks like `where` is having the opposite effect?

```sql
($2::varchar::varchar != $3::varchar)
```

Or:

```
('active' != 'active')
```

It seems to me like the `where` clauses aren't being negated as expected.

For instance,`where present(:status)` yields `("status"::varchar::varchar IS NULL) AND`

Sure enough, if I do this:

```elixir
update :update do
  primary? true
  accept [:status]

  validate Is100Percent do
    where attribute_does_not_equal(:status, "active")
  end
end
```

Things trigger where they should (or shouldn't, lol). I get: `($2::varchar::varchar = $3::varchar) AND`

**To Reproduce**
It looks like where is just all backwards. If this hasn't been captured by normal testing, something in my specific validation is causing confusion. My guess would be it's related to the aggregate?

**Expected behavior**
The opposite of what is happening.

**Runtime**
- Elixir version 1.17.2
- Erlang version 27.0.1
- OS macOS 15.0.1
- Ash version 3.4.43

**Additional context**
[We discussed this in the Discord.](https://discord.com/channels/711271361523351632/711271361523351636/1310775099275874408)

binarypaladin:
Conditions didn't work in fully atomic scenarios because the code was handling the logic per validation in a backward manner. This negates the result of each validation expression do atomic conditions work as expected.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

Fixes #1619.

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1094,48 +1094,56 @@ defmodule Ash.Changeset do
   end
 
   @doc false
+  # Returns either an appropriate expression for an atomic condition or a value
+  # indicated that the condition cannot be handled atomically.
+  #
+  # Validation logic matches on failure. So, for example, `present(:field)` is
+  # going to _match_ when `:field` is `nil`. However, when applying this logic
+  # to a `where` condition, the opposite is desired. The end result is kinda
+  # ugly because it can end up reading like "not is not equal to" but
+  # ultimately produces the correct results.
+  @spec atomic_condition([{module(), keyword()}], Ash.Changeset.t(), map()) ::
+          {:atomic, Ash.Expr.t() | boolean()} | {:not_atomic, String.t()}
   def atomic_condition(where, changeset, context) do
     Enum.reduce_while(where, {:atomic, true}, fn {module, validation_opts},
-                                                 {:atomic, condition} ->
+                                                 {:atomic, condition_expr} ->
       case module.atomic(
              changeset,
              validation_opts,
              struct(Ash.Resource.Validation.Context, context)
            ) do
         :ok ->
-          {:cont, {:atomic, condition}}
-
-        {:error, _} ->
-          {:cont, {:atomic, false}}
-
-        [{:atomic, _, expr, _as_error} | rest] ->
-          exprs = [expr | Enum.map(rest, &elem(&1, 2))]
-
-          new_expr =
-            Enum.reduce(exprs, condition, fn expr, condition ->
-              if condition == true do
-                expr
-              else
-                expr(^condition and ^expr)
-              end
-            end)
-
-          {:cont, {:atomic, new_expr}}
+          {:cont, {:atomic, condition_expr}}
 
         {:atomic, _, expr, _as_error} ->
-          new_expr =
-            if condition == true do
-              expr
-            else
-              expr(^condition and ^expr)
-            end
+          {:cont, {:atomic, atomic_condition_expr(condition_expr, expr)}}
 
-          {:cont, {:atomic, new_expr}}
+        {:error, _} ->
+          # Error from the validator, so the validations should just fail with
+          # a `false` expression.
+          {:halt, {:atomic, false}}
 
-        {:not_atomic, reason} ->
-          {:halt, {:not_atomic, reason}}
+        {:not_atomic, _reason} = not_atomic ->
+          {:halt, not_atomic}
+
+        atomic_conditions when is_list(atomic_conditions) ->
+          atomic_conditions
+          |> Stream.map(fn {:atomic, _, expr, _as_error} -> expr end)
+          |> Enum.reduce(condition_expr, &atomic_condition_expr(&2, &1))
+          |> then(&{:cont, {:atomic, &1}})
       end
     end)
+  end
+
+  # This is not expressly necessary as `expr(true and not ^new_expr)` would also
+  # work just fine, but the final output from omitting `true` is much easier to
+  # read if debugging.
+  defp atomic_condition_expr(true, expr) do
+    expr(not (^expr))
+  end
+
+  defp atomic_condition_expr(condition_expr, expr) do
+    expr(^condition_expr and not (^expr))
   end
 
   defp atomic_params(changeset, action, params, opts) do
--- a/test/actions/atomic_update_test.exs
+++ b/test/actions/atomic_update_test.exs
@@ -15,6 +15,15 @@ defmodule Ash.Test.Actions.AtomicUpdateTest do
     end
   end
 
+  # Forces `where` conditions to be evaluated atomically.
+  defmodule AtomicOnlyValidation do
+    use Ash.Resource.Validation
+
+    def atomic(_, _, _) do
+      :ok
+    end
+  end
+
   defmodule NotAtomic do
     use Ash.Resource.Change
 
@@ -44,6 +53,21 @@ defmodule Ash.Test.Actions.AtomicUpdateTest do
 
         validate attribute_equals(:name, "fred")
         validate compare(:score, greater_than_or_equal_to: 0, less_than_or_equal_to: 10)
+      end
+
+      update :with_conditional_validation do
+        accept([:bio, :name, :score])
+
+        validate present(:bio) do
+          where [
+            AtomicOnlyValidation,
+            attribute_equals(:name, "Bill S. Preston, Esq."),
+            changing(:score),
+            compare(:score, greater_than_or_equal_to: 0, less_than_or_equal_to: 10)
+          ]
+
+          message "If your name is Bill S. Preston, Esq. and you're providing a new score between 0 and 10, you most provide a bio!"
+        end
       end
 
       update :with_around_action do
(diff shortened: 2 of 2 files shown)


lib/ash/changeset/changeset.ex
```diff
@@ -1094,50 +1094,58 @@ defmodule Ash.Changeset do
...
       end
     end)
   end
 
+  # This is not expressly necessary as `expr(true and not ^new_expr)` would also
+  # work just fine, but the final output from omitting `true` is much easier to
+  # read if debugging.
+  defp atomic_condition_expr(true, expr) do
+    expr(not (^expr))
+  end
```
binarypaladin, line 1143: @zachdaniel I know we had mentioned dropping this, but see my comments. Maybe no one ever sees it, but it made debugging the output easier on my eyes.

lib/ash/changeset/changeset.ex
```diff
@@ -1094,50 +1094,58 @@ defmodule Ash.Changeset do
...
-
-          {:cont, {:atomic, new_expr}}
+          {:cont, {:atomic, atomic_condition_expr(condition_expr, expr)}}
 
-        {:not_atomic, reason} ->
-          {:halt, {:not_atomic, reason}}
+        {:error, _} ->
+          # Error from the validator, so the validations should just fail with
+          # a `false` expression.
+          {:halt, {:atomic, false}}
```
binarypaladin, line 1124: This is more efficient than `{:cont, {:atomic, false}}` although both yield the same result.

lib/ash/changeset/changeset.ex
```diff
@@ -1094,50 +1094,58 @@ defmodule Ash.Changeset do
...
+  # going to _match_ when `:field` is `nil`. However, when applying this logic
+  # to a `where` condition, the opposite is desired. The end result is kinda
+  # ugly because it can end up reading like "not is not equal to" but
+  # ultimately produces the correct results.
+  @spec atomic_condition([{module(), keyword()}], Ash.Changeset.t(), map()) ::
+          {:atomic, Ash.Expr.t() | boolean()} | {:not_atomic, String.t()}
   def atomic_condition(where, changeset, context) do
     Enum.reduce_while(where, {:atomic, true}, fn {module, validation_opts},
-                                                 {:atomic, condition} ->
+                                                 {:atomic, condition_expr} ->
```
binarypaladin, line 1109: I renamed this variable because `condition` coupled with it starting as `true` was _really_ confusing at first glance. I get it now, but I think this slight change improves the clarity.

lib/ash/changeset/changeset.ex
```diff
@@ -1094,50 +1094,58 @@ defmodule Ash.Changeset do
...
+  # Returns either an appropriate expression for an atomic condition or a value
+  # indicated that the condition cannot be handled atomically.
+  #
+  # Validation logic matches on failure. So, for example, `present(:field)` is
+  # going to _match_ when `:field` is `nil`. However, when applying this logic
+  # to a `where` condition, the opposite is desired. The end result is kinda
+  # ugly because it can end up reading like "not is not equal to" but
+  # ultimately produces the correct results.
+  @spec atomic_condition([{module(), keyword()}], Ash.Changeset.t(), map()) ::
+          {:atomic, Ash.Expr.t() | boolean()} | {:not_atomic, String.t()}
```
binarypaladin, line 1106: Double check. I know this is a semi-private function but I like specs when just trying to figure out a function at a glance.

lib/ash/changeset/changeset.ex
```diff
@@ -1094,50 +1094,58 @@ defmodule Ash.Changeset do
...
+          {:halt, {:atomic, false}}
+
+        {:not_atomic, _reason} = not_atomic ->
+          {:halt, not_atomic}
+
+        atomic_conditions when is_list(atomic_conditions) ->
+          atomic_conditions
+          |> Stream.map(fn {:atomic, _, expr, _as_error} -> expr end)
+          |> Enum.reduce(condition_expr, &atomic_condition_expr(&2, &1))
+          |> then(&{:cont, {:atomic, &1}})
```
binarypaladin, line 1133: This was a combination of better DRY and what I felt was safer matching. This ensures every item in the list conforms to the correct (as much as it can) format and I think it reads better too.

I don't actually have a test to check this path. What's an example of a situation where this would be returned from `atomic/3` on a validation?

binarypaladin: I tried writing an explicit test for this but after tossing `IO.inspect/3` in the mix, it appears this is getting checked by the tests and everything is indeed hunky dory.

lib/ash/changeset/changeset.ex
```diff
@@ -1094,50 +1094,58 @@ defmodule Ash.Changeset do
...
+          # a `false` expression.
+          {:halt, {:atomic, false}}
+
+        {:not_atomic, _reason} = not_atomic ->
+          {:halt, not_atomic}
+
+        atomic_conditions when is_list(atomic_conditions) ->
+          atomic_conditions
+          |> Stream.map(fn {:atomic, _, expr, _as_error} -> expr end)
+          |> Enum.reduce(condition_expr, &atomic_condition_expr(&2, &1))
```
zachdaniel, line 1132: I'd probably just do the matching here, i.e `Enum.reduce(condition_expr, fn {:atomic, _, expr, _} -> ...` since streams have some overhead and aren't usually worth it on small collections.

lib/ash/changeset/changeset.ex
zachdaniel: You could ensure that you are testing this by either 
1: using `Ash.bulk_update(a_query, :the_action, %{...input}, strategy: :atomic)`
2: testing with an update action that has `require_atomic?: true` (the default for update actions)

zachdaniel: You're already doing #2 so yes, hunky dory indeed ❤️

lib/ash/changeset/changeset.ex
zachdaniel: actually its such a tiny change I'll make it myself. No need to hold up this awesome PR ❤️

zachdaniel: yknow what, it doesn't matter. I'll leave it 😄

binarypaladin: I have had it in my head to always Stream and then finish with an Enum. I wonder where I picked that up. Should I avoid enums on small collections? (Any chance you have a source handy?)

zachdaniel: https://elixirforum.com/t/understanding-streams-and-enum/18574/2

binarypaladin: Ah-ha. From Valim himself:

> I like to say that streams are about using less memory at the cost of CPU.

I _think_ I got the idea from [this book](https://www.manning.com/books/elixir-in-action-third-edition) which was the first Elixir book I read (and I think it's great too). I'm going to have to dig it up and see how well my memory matches reality—at my age, that's becoming a more frequent issue.

Oh, and look at that... the author of said book [literally weights in](https://elixirforum.com/t/understanding-streams-and-enum/18574/5?u=binarypaladin) just after José. Welp, I need to revise my position and link this to my team, lol. Thanks.

PR #1629, opened 2024-11-27
add docs and typespecs for most of the error modules

allenwyma:
# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/error/error.ex
```diff
@@ -27,6 +44,11 @@ defmodule Ash.Error do
     to_error(value, Keyword.put(opts, :stacktrace, stacktrace))
   end
 
+  @doc """
+  Converts a value to an Ash.Error type.
+  """
+  # @spec to_error_class(ash_error_fields() | term() | [ash_error_fields()] | [term()], Keyword.t()) :: ash_errors() | [ash_errors()]
```
zachdaniel, line 50: Hmm...it should be returning `t()` every time. Is it not? the `t()` is provided by splode itself.

allenwyma:
@zachdaniel looks like latest master dialyzer is broken? not sure how to fix that if you can handle that?

lib/ash/error/error.ex
```diff
@@ -11,6 +11,23 @@ defmodule Ash.Error do
     ],
     unknown_error: Ash.Error.Unknown.UnknownError
 
+  @type ash_errors ::
+          Ash.Error.Forbidden.t()
+          | Ash.Error.Invalid.t()
+          | Ash.Error.Framework.t()
+          | Ash.Error.Unknown.t()
+          | Ash.Error.Unknown.UnknownError.t()
```
zachdaniel, line 19: `ash_errors` here includes any of a huge list of errors, not just the classes. It also includes custom errors defined by end users that use one of our error classes. I think the best we can do here is `Exception.t()`.

lib/ash/error/error.ex
```diff
@@ -106,8 +133,17 @@ defmodule Ash.Error do
...
   defp header(:unknown), do: "Unknown Error"
 
+  @doc """
+  Returns whether or not a term is an Ash.Error type.
+  """
+  @spec ash_error?(term()) :: boolean()
   def ash_error?(value), do: splode_error?(value, __MODULE__)
 
+  @doc """
+  Sets the main path of the errors on the containing struct.
```
zachdaniel, line 143: This function prepends the provided path to any existing path on the errors, may be good to express that in the docs.

zachdaniel:
Fixed the latest dialyzer, thanks for letting me know!

zachdaniel:

PR #1644, opened 2024-12-04
docs(filters): clarify how to use string maps

alexslade:
I personally struggled with the current description. I hope this reduced wording is more explicit, while re-using explanation from `Ash.Query.filter_input/2`.

Following on from my post [here](https://elixirforum.com/t/how-to-filter-an-array-attribute-to-find-at-least-one-instance-using-map-syntax/67887) while the iron is hot. If I've misunderstood things I'll remove/correct this PR as needed. 

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -116,12 +116,10 @@ defmodule Ash.Filter do
 
   ## Security Concerns
 
-  If you are using a map with string keys, it is likely that you are parsing
-  input. It is important to note that, instead of passing a filter supplied from
-  an external source directly to `Ash.Query.filter/2`, you should call
-  `Ash.Filter.parse_input/2`.  This ensures that the filter only uses public
-  attributes, relationships, aggregates and calculations, honors field policies
-  and any policies on related resources.
+  Do not pass user input directly to `Ash.Query.filter/2`, it will not be sanitised. Instead use
+  `Ash.Filter.parse_input/2` or `Ash.Query.filter_input/2`.
+
+  Refer to those functions for more information on how to safely work with user input.
 
   ## Writing a filter
 


zachdaniel:
So it's a bit more nuanced than this. It does in fact confirm that any fields referenced are public fields, which is important. When using `filter_input`, it also annotates any field references as "input?: true", which is used to control various aspects of authorizing a read action. Lets use a simple example where `User` resource has a policy like 

```elixir
policy :read do
  authorize_if expr(profile_public)
  authorize_if expr(id == ^actor(:id))
end
```

Additionally, lets imagine they had a field policy like this:

```elixir
field_policy :id, always() do
  authorize_if expr(profile_public and show_email)
  authorize_if expr(id == ^actor(:id))
end
```

Then you had a resource called `Post` which `has_many` `User`s.

If they were to do something like this:

```elixir
User
|> Ash.Query.filter_input(%{user: %{email: %{eq: "foo@bar.com"}}})
```

The ultimate authorized query would actually look like this (pretty wacky, I know, but this is what it looks like to actually have securely filterable resource interfaces that honor access rules in an app 😄 )

```elixir
User
|> Ash.Query.filter_input(
  user.profile_public or user.id == ^actor(:id) and (
    if (profile_public and show_email) or user.id == ^actor(:id) do
      user.email
    else
      nil
    end
  )
```

If you were to just do
`Ash.Query.filter(...)`

with that same input, the reference would be considered an "internal" reference, and those rules would not be applied. This is how you can write actions that have internal filtering rules independent of user authentication status.

alexslade:
Thanks for the information @zachdaniel. After seeing `Ash.Query.filter_input` I figured this docstring could be slimmed down and mostly just refer to it. 

If you don't like this, feel free to just close the PR.

zachdaniel (approved):

zachdaniel:

PR #1652, opened 2024-12-11
`parse_input/2` fails when being passed relationships

StephanH90:
I assume it should be possible to filter by nested relationships when using strings as input to `parse_input/2`?


zachdaniel:

PR #1655, opened 2024-12-12
Calc examples

allenwyma:
# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
So I'm not necessarily opposed to having these as they are, but currently all of the how-to guides are done as live books so that users can try them out and play with them on their own. Would you be open to putting them in livebooks?

allenwyma:
hi @zachdaniel i'll get back to this, just been slammed with work. I'll try to use the sample style like you have setup.

PR #1669, opened 2024-12-20
improvement: add support for `strict?` in read options

vonagam:
`Ash.Query.load` has `strict?` option that limits loaded/selected fields to those that explicitly specified.

This PR allows passing that option to read methods - `Ash.get`/`Ash.read`/`Ash.read_one` and so on - that already do have `load`/`reuse_values?` options.

So now one will be able to write `Ash.get!(User, id, load: [:name], strict?: true)`.


zachdaniel:

PR #1680, opened 2024-12-28
allow exclusion of certain actions from publish_all

ademenev:
This patch adds `except` option to `pub_sub.publish_all` that allows to exclude certain actions from notifications.

My motivation for this is the following scenario:

I have a resource that has multiple update actions that require notifications, and only one action for which notifications are not required (and are pointless since it is only used during application shutdown). Without this option, I have to list all actions separately and remember to update the notifications list each time I add a new action.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel (approved):

zachdaniel:

PR #1681, opened 2024-12-29
Add default code interface options

binarypaladin:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

We had discussed [this feature in Discord](https://discord.com/channels/711271361523351632/711271361523351636/1303019179472977930) when I first started messing around with Ash. I figured I'd finally give it a crack.

I did more refactoring. It's probably easier to understand specifically what happened going through the changes commit by commit rather than just looking over the files changed. I left a pretty good trail of what I did.

I wonder if `Ash.CodeInterface.params_and_opts/2` should exist there specifically or perhaps elsewhere. That pattern seems like it's really common.

What is the purpose of allowing keyword parameters? It seems like such a weird feature since it _must_ be paired with non-empty options.

### DIFF
--- a/documentation/dsls/DSL-Ash.Domain.md
+++ b/documentation/dsls/DSL-Ash.Domain.md
@@ -126,7 +126,7 @@ define :get_user_by_id, action: :get_by_id, args: [:id], get?: true
 | [`get?`](#resources-resource-define-get?){: #resources-resource-define-get? } | `boolean` | `false` | Expects to only receive a single result from a read action or a bulk update/destroy, and returns a single result instead of a list. Sets `require_reference?` to false automatically. |
 | [`get_by`](#resources-resource-define-get_by){: #resources-resource-define-get_by } | `atom \| list(atom)` |  | Takes a list of fields and adds those fields as arguments, which will then be used to filter. Sets `get?` to true and `require_reference?` to false automatically. Adds filters for read, update and destroy actions, replacing the `record` first argument. |
 | [`get_by_identity`](#resources-resource-define-get_by_identity){: #resources-resource-define-get_by_identity } | `atom` |  | Takes an identity, gets its field list, and performs the same logic as `get_by` with those fields. Adds filters for read, update and destroy actions, replacing the `record` first argument. |
-| [`default_options`](#resources-resource-define-default_options){: #resources-resource-define-default_options } | `any` | `[]` | Default options used when none are supplied. These can override domain or action defaults. |
+| [`default_options`](#resources-resource-define-default_options){: #resources-resource-define-default_options } | `keyword` | `[]` | Default options used when none are supplied. These can override domain or action defaults. |
 
 
 
--- a/documentation/dsls/DSL-Ash.Domain.md
+++ b/documentation/dsls/DSL-Ash.Domain.md
@@ -126,7 +126,7 @@ define :get_user_by_id, action: :get_by_id, args: [:id], get?: true
 | [`get?`](#resources-resource-define-get?){: #resources-resource-define-get? } | `boolean` | `false` | Expects to only receive a single result from a read action or a bulk update/destroy, and returns a single result instead of a list. Sets `require_reference?` to false automatically. |
 | [`get_by`](#resources-resource-define-get_by){: #resources-resource-define-get_by } | `atom \| list(atom)` |  | Takes a list of fields and adds those fields as arguments, which will then be used to filter. Sets `get?` to true and `require_reference?` to false automatically. Adds filters for read, update and destroy actions, replacing the `record` first argument. |
 | [`get_by_identity`](#resources-resource-define-get_by_identity){: #resources-resource-define-get_by_identity } | `atom` |  | Takes an identity, gets its field list, and performs the same logic as `get_by` with those fields. Adds filters for read, update and destroy actions, replacing the `record` first argument. |
-| [`default_options`](#resources-resource-define-default_options){: #resources-resource-define-default_options } | `keyword` | `[]` | Default options used when none are supplied. These can override domain or action defaults. |
+| [`default_options`](#resources-resource-define-default_options){: #resources-resource-define-default_options } | `keyword` | `[]` | Default options to be merged with client-provided options. These can override domain or action defaults. |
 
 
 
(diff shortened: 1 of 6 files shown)


documentation/dsls/DSL-Ash.Domain.md
```diff
@@ -126,6 +126,7 @@ define :get_user_by_id, action: :get_by_id, args: [:id], get?: true
...
 | [`get_by`](#resources-resource-define-get_by){: #resources-resource-define-get_by } | `atom \| list(atom)` |  | Takes a list of fields and adds those fields as arguments, which will then be used to filter. Sets `get?` to true and `require_reference?` to false automatically. Adds filters for read, update and destroy actions, replacing the `record` first argument. |
 | [`get_by_identity`](#resources-resource-define-get_by_identity){: #resources-resource-define-get_by_identity } | `atom` |  | Takes an identity, gets its field list, and performs the same logic as `get_by` with those fields. Adds filters for read, update and destroy actions, replacing the `record` first argument. |
+| [`default_options`](#resources-resource-define-default_options){: #resources-resource-define-default_options } | `any` | `[]` | Default options used when none are supplied. These can override domain or action defaults. |
```
zachdaniel: "When none are supplied" isn't quite right here is it? It merges options, even when some are supplied right?

We should also likely merge at one level deeper for `bulk_options`.

zachdaniel:
Two small comments, but this looks great! Thank you both for the new capability as well as the cleanup 🙇

documentation/dsls/DSL-Ash.Domain.md
binarypaladin: I should have just slept on it, lol. The last commit where I cleaned up my own mistakes implied I was tired… but I was in the zone. 

I’ll fix the docs a little later. I don’t actually understand the comment about `bulk_options`.

zachdaniel: There is a nested set of options called `bulk_options` that is accepted by code interfaces. If the caller specified `bulk_options` as well as the defaults we should merge the two values instead of overwriting.

binarypaladin: I'm a bit lost on the `bulk_options` comment. So far as I can tell, `:default_options` merges _before_ `:bulk_options`. That seems like the correct behavior.

My next project might be a full on refactor of that module. It'd be a good project to improve my macro skills, but there's a lot of duplication. Searching for `bulk_options` in the code led me to a bit of a... "Hmmm, where do I even start?"

Do you want to merge this as is (I pushed to doc changes) and I can change the way the bulk options resolve in another PR?

lib/ash/code_interface.ex
```diff
@@ -449,60 +503,25 @@ defmodule Ash.CodeInterface do
...
-                  {%{},
-                   unquote(interface_options).validate!(params_or_opts)
-                   |> unquote(interface_options).to_options()},
-                else:
-                  {params_or_opts,
-                   unquote(interface_options).validate!(opts)
-                   |> unquote(interface_options).to_options()}
+              Ash.CodeInterface.params_and_opts(params_or_opts, opts, fn opts ->
+                unquote(Macro.escape(interface.default_options))
+                |> Keyword.merge(opts)
```
zachdaniel, line 511:
```suggestion
                |> Keyword.merge(opts, fn :bulk_options, v1, v2 -> 
                  Keyword.merge(List.wrap(v1), List.wrap(v2))
                  
                  _key, v1, v2 ->
                    v2
                end)
```

This is the change I'm suggesting.

This way, if the interface sets `default_options: [bulk_options: [strategy: :stream]]` and the user specifies `bulk_options: [batch_size: 500]`, the value of `bulk_options` will be `[strategy: :stream, batch_size: 500]`, not `[batch_size: 500]`

binarypaladin: Hmmm. Is there anything else that might cause a similar issue where it'd make sense to deep merge?

zachdaniel: Probably also the `page` parameter and `load` parameter. These are more complex though, as `page` can be `false`. So if the users value is `false` we should overwrite the default, otherwise we should merge.

for `load` we should just be able to do something like `List.wrap(default) ++ List.wrap(new)`

binarypaladin: Time to write a function, lol.

binarypaladin: Written. This is turning into a lot more than I anticipated. I added a bunch of doctests too which... caught a bug.

lib/ash/resource/interface.ex
```diff
@@ -245,6 +246,12 @@ defmodule Ash.Resource.Interface do
       doc: """
       Takes an identity, gets its field list, and performs the same logic as `get_by` with those fields. Adds filters for read, update and destroy actions, replacing the `record` first argument.
       """
+    ],
+    default_options: [
+      type: :keyword_list,
+      default: [],
+      doc:
+        "Default options to be merged with client-provided options. These can override domain or action defaults."
```
zachdaniel, line 254:
```suggestion
        "Default options to be merged with client-provided options. These can override domain or action defaults. `:load`, `:bulk_options`, and `:page` options will be deep merged."
```

zachdaniel:

mavuio:
since 3.4.50 this call now brings an error, when using an options-array:

```
MyApp.Agent.get_by_id("01907343-3045-721e-8822-be3266f4ee26",[])
** (BadMapError) expected a map, got: []
    (elixir 1.18.1) lib/map.ex:836: Map.split([], [:id])
    (my_app 1.2.9) deps/ash/lib/ash/code_interface.ex:683: MyApp.Agent.get_by_id/3
    iex:4: (file)
```

 seems like the params_or_opts only accept maps now.

binarypaladin:
I’ll add a fix. There also needs to be a test for this!

binarypaladin:
@mavuio I believe #1692 will fix this.

binarypaladin:
@zachdaniel what's weird is... I tried to replicate the above error in the current code interface tests and...

```elixir
User.get_by_id(user.id, [])
```

Results in:

```elixir
{:error,
 %Ash.Error.Invalid{
   errors: [
     %Ash.Error.Query.Required{
       field: :id,
       type: :argument,
       resource: Ash.Test.CodeInterfaceTest.User,
       splode: Ash.Error,
       bread_crumbs: [],
       vars: [],
       path: [],
       class: :invalid
     }
   ]
 }}
```

However, jumping back to before my changes were applied resulted in the same thing. I tested my function in the doctest but I didn't see how to test for this and get the same results @mavuio mentioned.

zachdaniel:
Yeah, I commented similarly in your linked PR.  not sure how to reach this particular case.

binarypaladin:
lol. I may have jumped the gun on a fix.

PR #1684, opened 2024-12-29
fix: properly handle aggregate defaults in ets/mnesia

vonagam:
Right now ETS data layer (and Mnesia because of that) use aggregate default only in case of `first` aggregate kind and only when there are no records - so if the first record has a value of nil then it will return nil. That is not how it is in other data layers or in Ash runtime filter logic - if the resulted aggregate value is nil (because there are no records or because they have nil values) then the default is used. This PR aligns ETS logic with that of sql/runtime ash.


zachdaniel:

PR #1692, opened 2025-01-02
fix: allow non-list and non-map param

binarypaladin:
Empty options combined with a non-list, non-map param broke in the prior code.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -210,10 +210,11 @@ defmodule Ash.CodeInterface do
   With those two defaults in mind, this function will decipher, from two inputs,
   which should be parameters and which should be options.
 
-  Parameters can take one of two primary forms:
+  Parameters can take one of three primary forms:
 
     1. A map.
     2. A list of maps for bulk operations.
+    3. A single value, such as a UUID or integer (used as an ID).
 
   Additionally, if options are set explicitly (i.e. at least one option has
   been set), a keyword list will be converted to a map.
@@ -234,16 +235,19 @@ defmodule Ash.CodeInterface do
 
       iex> params_and_opts([key: :param], [key: :opt])
       {%{key: :param}, [key: :opt]}
+
+      iex> params_and_opts("9aee6849-d52a-4bb9-818d-ff242ac6582f", [])
+      {"9aee6849-d52a-4bb9-818d-ff242ac6582f", []}
   """
-  @spec params_and_opts(params_or_opts :: map() | [map()] | keyword(), keyword()) ::
-          {params :: map() | [map()], opts :: keyword()}
+  @spec params_and_opts(params_or_opts :: term(), keyword()) ::
+          {params :: term(), opts :: keyword()}
   def params_and_opts(%{} = params, opts), do: {params, opts}
 
   def params_and_opts([], opts), do: {[], opts}
 
   def params_and_opts([%{} | _] = params_list, opts), do: {params_list, opts}
 
-  def params_and_opts(opts, []), do: {%{}, opts}
+  def params_and_opts(opts, []) when is_list(opts), do: {%{}, opts}
 
   def params_and_opts(params_or_list, opts) do
     params =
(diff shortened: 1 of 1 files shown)


zachdaniel:
I'm actually surprised that this is an issue 🤔 I don't think that params or opts are meant to be able to be a non list/non map... thats what `args` are for. @mavuio can you share your `code_interface` definition that produces that function that originally failed? We should reproduce the issue first.

mavuio:
well, my code interface is really simple:

`    
define :get_by_id, action: :read, get_by: [:id]
`

zachdaniel:
@binarypaladin perhaps that is enough to go off of for a repro?

binarypaladin:
> @binarypaladin perhaps that is enough to go off of for a repro?

Looking now. It differs from the one in the test case.

binarypaladin:
So, this might be beyond me right now but, here's what's defined in the test:

```elixir
      define :get_by_id, action: :by_id, get?: true, args: [:id]
```

I can change this to:

```elixir
      define :get_by_id, action: :read, get_by: [:id]
```

These two work differently, and this is an area I am unfamiliar with. With just one argument, both work the same. However, if I add an empty list like this:

```elixir
    assert User.get_by_id!(user.id, []).id == user.id
```

The original produces:

```
  1) test get_by! adds the proper arguments and filters (Ash.Test.CodeInterfaceTest)
     test/code_interface_test.exs:386
     ** (Ash.Error.Invalid)
     Invalid Error
```

The second version produces:

```
  1) test get_by! adds the proper arguments and filters (Ash.Test.CodeInterfaceTest)
     test/code_interface_test.exs:386
     ** (BadMapError) expected a map, got: []
```

So that's weird in and of itself. Additionally, the "fix" here doesn't actually have any effect at all.

The line causing the issue is actually this one:

```elixir
  def params_and_opts([], opts), do: {[], opts}
```

The params being passed in via `define :get_by_id, action: :read, get_by: [:id]` are `[]`. But the args are `%{id: uuid}` and they're not getting merged in because... there are no args. However, I'm not sure how to fix because we need to support an empty list as an acceptable input for bulks here:

```elixir
    test "bulk_create can be take an empty list" do
      assert %Ash.BulkResult{status: :success} = User.bulk_create!([])
    end
```

The broader question here, before I dive in to fix more is... is the fact that something that appears to be the same function can be defined two ways and have different functionality its own problem?

binarypaladin:
@zachdaniel 

Okay... so, here's the problem.

The divergence between the bang and non-bang version is where the trouble starts.

In 3.4.49 the non-bang version works as @mavuio expected. However, the bang version produces the `BadMapError` as far back as `3.0.0`. The non-bang version works fine. What breaks in 3.4.50 isn't anything in that function, it's the fact that bang and non-bang no longer treat parameters differently and I opted to follow the bang path.

For instance, the non-bang version _did not_ work with keyword parameters. (Something I think should be dropped anyway because it's super weird since it _requires_ non-empty options to work in the first place.) I "fixed" this error.

Since the tests followed only the bang path, I assumed that was the correct path.

There are a legion of inconsistencies prior to 3.4.50.

For example: `User.bulk_create!([])` results in `%{}` for params (the path I followed) and `User.bulk_create([])` results in `[]`. 

So far as I can tell, in 3.4.49 sending a list of maps also does not work at all.

Non-bang:

```elixir
params =
  unquote(args)
  |> Enum.zip([unquote_splicing(arg_vars)])
  |> Enum.reduce(params, fn {key, value}, params ->
    Map.put(params, key, value)
  end)
```

And bang:

```elixir
params =
  if is_list(params) do
    to_merge =
      unquote(args)
      |> Enum.zip([unquote_splicing(arg_vars)])
      |> Map.new()

    Enum.map(params, fn params ->
      Map.merge(params, to_merge)
    end)
  else
    unquote(args)
    |> Enum.zip([unquote_splicing(arg_vars)])
    |> Enum.reduce(params, fn {key, value}, params ->
      Map.put(params, key, value)
    end)
  end
```

The line causing the trouble is indeed:

```elixir
def params_and_opts([], opts), do: {[], opts}
```

That exists specifically to accommodate this:

```elixir
assert %Ash.BulkResult{status: :success} = User.destroy([])
```

I don't know how to fix this because I don't know what the "right" answer here is. Removing the above clause would fix the issue but break above example. 

I don't really understand the purpose of `User.destroy([])`. If it's possible, dropping the need to support that _and_ dropping the keyword list as params path would probably clear up a lot of confusion. Otherwise... there's probably going to be even more conditional logic or state that needs to be passed in upstream.

binarypaladin:
The divergence happened in 3.0. Since commits are squashed, it's hard to tell precisely where or what the rationale was. Whatever the case, a number of features were added to bang that didn't make it to non-bang version. If that was intentional, I need to refactor what I have a little to match that although... I'd also want to understand why they diverge.

PR #1702, opened 2025-01-08
fix: Fix typo in `initial_data`

sevenseacat:
I randomly came across this typo while looking at a diff for another issue I'm facing.

Now, I don't know what this code is for, and this condition would have never passed before, so this change might have unintended effects.

The tests pass both before and after making the change, so either it's not tested or the conditional has no impact here anyway.


zachdaniel:
Well, fun fact, none of that context is used anywhere else. So I'll just remove that whole block. Thanks for pointing this out :)

PR #1703, opened 2025-01-09
fix: Fix issues in `Ash.Generator.generate_many/2`

sevenseacat:
When testing out `generate_many/2` I found there were a couple of issues with it:

* It would error if `after_action: nil` was passed to `bulk_create`
* It would error if the full action struct was passed to `bulk_create` (instead of just the action name)
* It would error if the action passed to `bulk_create` required authorization (no actor being passed in)
* It would raise warnings if the action used in `bulk_create` generated any notifications (tbh I'm not sure *why* the action I'm testing with is generating notifications, but that's a different story)
* It would ignore partial success results, leading to random test failures due to intermittent erroneous generator data

With these changes I can successfully call `generate_many` like `generate_many(artist(), 2)` and get two Artist records back.

### DIFF
--- a/lib/ash/generator/generator.ex
+++ b/lib/ash/generator/generator.ex
@@ -12,7 +12,7 @@ defmodule Ash.Generator do
     use Ash.Generator
 
     # using `seed_generator`, bypasses the action and saves directly to the data layer
-    def blog_post(opts \\ []) do
+    def blog_post(opts \\\\ []) do
       seed_generator(
         %MyApp.Blog.Post{
           name: sequence(:title, &"My Blog Post \#{&1}")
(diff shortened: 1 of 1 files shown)


lib/ash/generator/generator.ex
```diff
@@ -505,12 +505,27 @@ defmodule Ash.Generator do
             fn _changeset, record ->
               {:ok, after_action.(record)}
             end
+          else
+            # Do nothing
+            fn _changeset, record -> {:ok, record} end
```
zachdaniel, line 510: I'll merge but I'm going to figure out how to make this not necessary, as adding hooks that aren't needed can prevent internal optimizations etc.

zachdaniel:

PR #1710, opened 2025-01-12
Fix nested load in calculation

jechol:
After commit 3dce3f63, a calculation with a nested load in my code failed to load, causing an error. When I reversed the if/else statement in the corresponding change, the error disappeared. Since I don’t fully understand all the code in that section, please review it.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/calculations.ex
+++ b/lib/ash/actions/read/calculations.ex
@@ -1742,9 +1742,9 @@ defmodule Ash.Actions.Read.Calculations do
 
           new_calculation =
             if should_be_in_expression?(new_calculation, query) do
-              new_calculation
-            else
               query.calculations[new_calculation.name]
+            else
+              new_calculation
             end
 
           domain


zachdaniel:
What error did you get? How did it fail?

zachdaniel:
If you could reproduce it in a test and include that in this PR that would be great.

zachdaniel:
Please open an

Torkan:
> If you could reproduce it in a test and include that in this PR that would be great.

This looks like more or less the same issue as I created here, which includes a repro:
https://github.com/ash-project/ash/issues/1712

jechol:
I am closing this PR in favor of [#1720](https://github.com/ash-project/ash/issues/1720), which better addresses the issue and includes a more comprehensive context. Thank you for reviewing this PR. Let’s continue the discussion and tracking in the new issue.

PR #1715, opened 2025-01-13
Add `Ash.Type.EctoType.autogenerate/1` for drop in replacement with existing Ecto Schemas

### LINKED ISSUE #1714: UUID and UUIDv7: Add Ecto.ParameterizedType.autogenerate/1 callback for backwards compatability with Ecto schemas
**Describe the bug**
`Ash.Type.UUID.EctoType` and `Ash.Type.UUIDv7.EctoType` are missing the `Ecto.Type.autogenerate` callback, which doesn't allow you to use either with Ecto schema's `@primary_key` module attribute for using custom types.

**To Reproduce**
1. The existing app has the `@primary_key` and `@foreign_key_type` module attributes set to `:binary_id` which uses `Ecto.UUID`

```elixir
defmodule MyApp.Schema
  defmacro __using__(_) do
      quote do
        import Ecto.Changeset
  
        @primary_key {:id, :binary_id, autogenerate: true}
        @foreign_key_type :binary_id
      end
    end
end
```

2. Convert to use Ash.UUID to make the existing Ecto schemas aware of the new type and able to autogenerate the UUIDs.

```elixir
defmodule MyApp.Schema
  defmacro __using__(_) do
      quote do
        import Ecto.Changeset
  
        @primary_key {:id, Ash.Type.UUID.EctoType, autogenerate: true}
        @foreign_key_type Ash.Type.UUID.EctoType
      end
    end
end
```

3. You'll get the following error
```elixir
* (ArgumentError) field :id does not support :autogenerate because it uses a parameterized type #Ash.Type.UUID.EctoType<[]> that does not define autogenerate/1
    (ecto 3.12.5) lib/ecto/schema.ex:2541: Ecto.Schema.store_type_autogenerate!/5
    (ecto 3.12.5) lib/ecto/schema.ex:2118: Ecto.Schema.define_field/4
    lib/petal_pro/posts/post.ex:9: (module)
```

**Expected behavior**
When putting `Ash.Type.UUID.EctoType`, you expect the existing Ecto schemas to be able to autogenerate the UUID just like it does when using `:binary_id` or any other UUID library.

**Runtime**
 - Elixir version 1.17.2-otp-27
 - Erlang version 27.2
 - OS
 - Ash version v3.4.55

**Additional context**

This is an easy fix by adding the `autogenerate` callback to both `Ash.Type.UUID.EctoType` and `Ash.Type.UUIDv7.EctoType` which I'll submit a PR for shortly.

chazwatkins:
# Contributor checklist

Closes #1714 

- Adds `autogenerate_enabled?` option to `Ash.Type`.  When true `Ash.Type.EctoType.autogenerate/1` is added to the module.
- Adds `Ash.Type.EctoType.autogenerate/1` which mirrors `Ecto.ParamaterizedType.autogenerate/1` so Ash types like `Ash.Type.UUID.EctoType` and `Ash.Type.UUIDv7.EctoType` can be drop ins for existing Ecto schemas `@primary_key` and `@foreign_key_type` schema module attributes.

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/type.ex
+++ b/lib/ash/type/type.ex
@@ -1638,6 +1638,15 @@ defmodule Ash.Type do
 
         @impl true
         def embed_as(_, _), do: :self
+
+        if Keyword.get(unquote(opts), :autogenerate_enabled?) do
+          @impl true
+          def autogenerate(constraints) do
+            constraints
+            |> @parent.generator()
+            |> Enum.at(0)
+          end
+        end
       end
 
       @impl true
--- a/lib/ash/type/uuid.ex
+++ b/lib/ash/type/uuid.ex
@@ -5,7 +5,8 @@ defmodule Ash.Type.UUID do
   A builtin type that can be referenced via `:uuid`
   """
 
-  use Ash.Type
+  use Ash.Type,
+    autogenerate_enabled?: true
 
   @impl true
   def storage_type(_), do: :uuid
--- a/lib/ash/type/uuid_v7.ex
+++ b/lib/ash/type/uuid_v7.ex
@@ -5,7 +5,8 @@ defmodule Ash.Type.UUIDv7 do
   A builtin type that can be referenced via `:uuid_v7`
   """
 
-  use Ash.Type
+  use Ash.Type,
+    autogenerate_enabled?: true
 
   @impl true
   def storage_type(_), do: :uuid


lib/ash/type/type.ex
```diff
@@ -1638,6 +1638,11 @@ defmodule Ash.Type do
 
         @impl true
         def embed_as(_, _), do: :self
+
+        @impl true
+        def autogenerate(_) do
+          Enum.at(@parent.generator([]), 0)
```
zachdaniel, line 1644: 🤔 this is an interesting implementation. For many types this wouldn't really work, like for strings it would return empty strings etc.

Is this *only* used in ecto when generating a type when explicitly configured with that module attribute?

If so, then I'm okay with this implementation.

However, the first argument is likely the constraints which need to be passed through to `generator/1`

zachdaniel: `Enum.at` is the right way, the main thing I'm wondering is if we want to make this an option in the type whether or not to define this callback. `use Ash.Type, autogeneratable?: true`

zachdaniel: LGTM!

zachdaniel:

PR #1716, opened 2025-01-14
fix: don’t require multitenancy attribute in `get`

adamtharani:
## Implements

- allows `get` to not require the multitenancy attribute

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -471,6 +471,18 @@ defmodule Ash.Filter do
         end
 
       {fields, value} ->
+        multitenancy_attribute = Ash.Resource.Info.multitenancy_attribute(resource)
+        fields = Enum.reject(fields, fn key -> key == multitenancy_attribute end)
+
+        {keyval?, value} =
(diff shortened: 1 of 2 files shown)


lib/ash/filter/filter.ex
```diff
@@ -448,7 +448,11 @@ defmodule Ash.Filter do
       :error # not enough information
   """
   def get_filter(resource, id) do
+    multitenancy_attribute = Ash.Resource.Info.multitenancy_attribute(resource)
     primary_key = Ash.Resource.Info.primary_key(resource)
+
+    primary_key = Enum.reject(primary_key, fn key -> key == multitenancy_attribute end)
```
zachdaniel, line 454: We should do this only in the third branch of the conditional below, because this could result in an empty primary key which would cause problems here.

zachdaniel (approved):

zachdaniel:

PR #1719, opened 2025-01-16
fix: matching in managed_relationships handle_update

stevebrambilla:
Quick fix for a small matching mistake in managed_relationships.ex


zachdaniel:

PR #1734, opened 2025-01-27
Fix/cascade destroy

barnabasJ:
fixes the test in https://github.com/ash-project/ash_postgres/pull/466

### DIFF
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -1729,17 +1729,35 @@ defmodule Ash.Actions.Destroy.Bulk do
     end)
     |> Enum.reduce(batch, fn {%{change: {module, change_opts}}, index}, batch ->
       if changes[index] == :all do
-        module.before_batch(
-          batch,
-          change_opts,
-          struct(Ash.Resource.Change.Context, %{
-            bulk?: true,
-            actor: opts[:actor],
-            tenant: opts[:tenant],
-            tracer: opts[:tracer],
-            authorize?: opts[:authorize?]
-          })
-        )
+        case change_opts do
+          {:templated, change_opts} ->
+            module.before_batch(
+              batch,
+              change_opts,
+              struct(Ash.Resource.Change.Context, %{
+                bulk?: true,
+                actor: opts[:actor],
+                tenant: opts[:tenant],
+                tracer: opts[:tracer],
+                authorize?: opts[:authorize?]
+              })
+            )
+
+          change_opts ->
+            Enum.flat_map(batch, fn record ->
+              module.before_batch(
+                [record],
+                change_opts,
+                struct(Ash.Resource.Change.Context, %{
+                  bulk?: true,
+                  actor: opts[:actor],
+                  tenant: opts[:tenant],
+                  tracer: opts[:tracer],
+                  authorize?: opts[:authorize?]
+                })
+              )
+            end)
+        end
       else
         {matches, non_matches} =
           batch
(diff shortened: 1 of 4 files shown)


lib/ash/actions/destroy/bulk.ex
```diff
@@ -1729,17 +1729,35 @@ defmodule Ash.Actions.Destroy.Bulk do
...
+                struct(Ash.Resource.Change.Context, %{
+                  bulk?: true,
+                  actor: opts[:actor],
+                  tenant: opts[:tenant],
+                  tracer: opts[:tracer],
+                  authorize?: opts[:authorize?]
+                })
+              )
+            end)
+        end
```
barnabasJ, line 1760: Similar problem to what was fixed in #1376

lib/ash/resource/change/cascade_update.ex
```diff
@@ -204,9 +204,21 @@ defmodule Ash.Resource.Change.CascadeUpdate do
...
-        Ash.bulk_update(query, action.name, inputs, context_opts)
+        Ash.bulk_update(
+          query,
+          action.name,
+          inputs,
+          Keyword.update(
+            context_opts,
+            :context,
+            relationship.context || %{},
+            &Map.merge(&1, relationship.context || %{})
```
barnabasJ, line 219: is there a reason that `cascade_destroy: true` is not passed into this bulk_destroy call?

test/resource/changes/cascade_destroy_test.exs
```diff
@@ -144,7 +144,10 @@ defmodule Ash.Test.Resource.Change.CascadeDestroy do
     1..Enum.random(3..5)
     |> Enum.map(fn _ -> Post.create!(%{author_id: author.id}) end)
 
-    Ash.bulk_destroy!([author], :no_notification_destroy, %{})
+    Ash.bulk_destroy([author], :no_notification_destroy, %{},
+      return_errors?: true,
+      strategy: [:stream, :atomic, :atomic_batches]
```
barnabasJ, line 149: Needed to add `:stream` because it now calls the before_batch hook and is therefore not atomic anymore

barnabasJ:
I'm not to happy with having to have the before_batch there. But I didn't now how to do it without it.

lib/ash/resource/change/cascade_update.ex
zachdaniel: No reason, makes sense to include it there (just using the `context`  you defined above)

zachdaniel:
Alright, so, I think we need to have both the old behavior and the new behavior as options. There are two reasons:

1. you can use `cascade_destroy` on an update action, which would be doable after the action
2. you can defer the constraints that would cause a foreign key error, as a performance improvement.

zachdaniel:
Ohhhhh right. So the right answer here is also a bit frustrating to implement probably.

If the action is a destroy action, we should try to do it without loading data actually.

What that looks like is:

- If the relationship filter has no parent references and the relationship is not a many_to_many (I think this doesn't support many to many anyway)
  - then we extract the source_attribute, and construct a query against the destination for the specific id matches that exist plus the relationship.filter
  - unless `no_attributes? true` in which case we just use the query as is

Otherwise for destroy actions they *must* use the new before_batch behavior.

barnabasJ:
> Ohhhhh right. So the right answer here is also a bit frustrating to implement probably.
> 
> If the action is a destroy action, we should try to do it without loading data actually.
> 
> What that looks like is:
> 
>     * If the relationship filter has no parent references and the relationship is not a many_to_many (I think this doesn't support many to many anyway)
>       
>       * then we extract the source_attribute, and construct a query against the destination for the specific id matches that exist plus the relationship.filter
>       * unless `no_attributes? true` in which case we just use the query as is
> 
> 
> Otherwise for destroy actions they _must_ use the new before_batch behavior.

I think that is pretty much what it is doing, isn't it? So in the [error](https://github.com/ash-project/ash/blob/86c51fb5f261bb92ed6b716372fb4c809d09594e/lib/ash/resource/change/cascade_destroy.ex#L251) case we just need to check if we are in the after_action part of a a destroy action and if so return an error?

zachdaniel:
lol. Yep, sorry. Was early when I was first responding, hadn't had my coffee yet.

So I think the only thing needed here is to make this behavior opt-out, but opting out would raise an error if it gets to that error case and is an after action as you mentioned.

test/resource/changes/cascade_destroy_test.exs
barnabasJ: Yeah, I forgot to change this back after adding in the old behaviour as default again.

zachdaniel:
Ready to go?

zachdaniel:

PR #1736, opened 2025-01-28
docs: fix reload/2 doc

Sangresz:
# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash.ex
```diff
@@ -2023,7 +2023,7 @@ defmodule Ash do
   end
 
   @doc """
-  Refetches a record by primary key or raises an error. See `reload/2` for more.
```
zachdaniel, line 2026: I think this one should be pointing to reload/2 as it is. So `reload!` -> `reload` -> `get`

PR #1739, opened 2025-01-28
improvement: Support guards in `Ash.Reactor` steps.

jimsynz:


zachdaniel:
The fact that we have to change all of our step types to leverage a new core step type feature of reactor is a bit concerning. Is there some way to improve this in the future?

jimsynz:
Yeah, I want to write some helpers in Reactor which will help generate entities for you.

zachdaniel (approved):

Issue #1551, opened 2024-10-22
misleading error if arguments are forgotten in Ash.Query.for_read


frankdugan3:
**Describe the bug**

When building a read query for an action with only optional arguments, a misleading error message occurs when no arguments are passed.

```
iex(1)> Ash.Query.for_read(Hsm.FFL.ADRecord, :bound_book_report, actor: nil)
** (Ash.Error.Forbidden) Forbidden Error

* The domain Hsm.FFL requires that an actor is provided at all times and none was provided.
  (elixir 1.17.3) lib/process.ex:864: Process.info/2
  (ash 3.4.35) 
```

**To Reproduce**

Build a read query on any action that has only optional arguments. Not sure if the same error also happens when the action has mandatory arguments.

**Expected behavior**

The error should indicate that it is an action that accepts arguments, but no arguments were passed. That or maybe it should automatically insert an empty map as the arguments? :thinking:

zachdaniel:
We should probably validate inputs first? Or require that inputs is a map? That second one would be a breaking change. Anyway, definitely confusing.

StephanH90:
I couldn't reproduce this. Simply running `Ash.Query.for_read(resource, :action, actor: nil)` returns an appropriate error:

```ex
iex(3)> Ash.Query.for_read(Ash.Test.QueryTest.User, :by_id)
#Ash.Query<
  resource: Ash.Test.QueryTest.User,
  filter: #Ash.Filter<id == nil>,
  errors: [
    %Ash.Error.Query.Required{
      field: :id,
      type: :argument,
      resource: Ash.Test.QueryTest.User,
      splode: Ash.Error,
      bread_crumbs: [],
      vars: [],
      path: [],
      stacktrace: #Splode.Stacktrace<>,
      class: :invalid
    }
  ]
>
```
Could it be that your domain force authorization, @frankdugan3? Because in that case the error looks exactly like you described (and it makes sense to me):

```ex
iex(5)> Ash.Query.for_read(Ash.Test.QueryTest.User, :by_id)
** (Ash.Error.Forbidden) Forbidden Error

* The domain Ash.Test.Domain requires that an actor is provided at all times and none was provided.
```

zachdaniel:
If a keyword list is passed then it should be treated as an opts list, so it should see `actor: nil`. the check to require an actor allows `nil` as a value, it just requires specifying it in some way.

ken-kost:
[...]

PR #1751, opened 2025-02-01
Improvement: handle for read without args

ken-kost:
#1551 
The
```
  def for_read(query, action_name, opts, _) when is_list(opts),
    do: for_read(query, action_name, %{}, opts)
```
kinda bugs me 🐞 but if it's better it's better. 🤷 
Also to note, some tests failed because caller was written as a keyword list and was considered as `args` and it worked. 🤔 
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/query/query.ex
+++ b/lib/ash/query/query.ex
@@ -523,7 +523,12 @@ defmodule Ash.Query do
   #{Spark.Options.docs(@for_read_opts)}
 
   """
-  def for_read(query, action_name, args \\ %{}, opts \\ []) do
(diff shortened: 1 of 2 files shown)


zachdaniel:
Well then. That actually means this is a breaking change 😭 I'll revisit on Monday but I think I that means we can't make this change. What we can do is deprecate (by warning on these calls) and then fix in the next major version.

zachdaniel:
Yeah, unfortunately not something we can do now.

Issue #1754, opened 2025-02-03
`WithClauseError` when an action argument constraint type does not match


sevenseacat:
**Describe the bug**

When defining an argument to an action like:

```elixir
      argument :google_place, :struct,
        constraints: [instance_of: Place],
        allow_nil?: false
```

And a struct is indeed passed in as the argument (but not the right type!) a WithClauseError is returned:

```
** (Ash.Error.Unknown)
Bread Crumbs:
  > building changeset for MyApp.Lead.create_lead_from_google_place

Unknown Error

* ** (WithClauseError) no with clause matching: {:constrained, :error, #MyApp.Place<...>}
  (ash 3.4.62) lib/ash/changeset/changeset.ex:5228: anonymous fn/5 in Ash.Changeset.do_set_argument/4
  (stdlib 6.2) maps.erl:860: :maps.fold_1/4
  (ash 3.4.62) lib/ash/changeset/changeset.ex:2087: Ash.Changeset.handle_params/4
  (ash 3.4.62) lib/ash/changeset/changeset.ex:1957: Ash.Changeset.do_for_action/4
  (my_app 2.2.0) deps/ash/lib/ash/code_interface.ex:785: MyApp.create_lead_from_google_place/2
  ...
```

After much head scratching, I did figure out that I was missing an alias for my Place resource. 

**Expected behavior**

A clearer error message, telling me that the constraint was violated because the struct types do not match.

**Runtime**
 - Elixir version  1.17.2-otp-27
 - Erlang version  27.2
 - OS macOS Sonoma
 - Ash version latest commit in main branch (a103d366d17b2deb7a89dcf33acc4fa6aed73950)

ken-kost:
Okay so I think what happens here is in the changeset part that fails:
```
             {:ok, casted} <-
               Ash.Type.cast_input(argument.type, value, constraints),
             {:constrained, {:ok, casted}, _last_val} when not is_nil(casted) <-
               {:constrained,
                Ash.Type.apply_constraints(argument.type, casted, constraints),
                casted}
```
`casted` is made from made up module so the call is `Ash.Type.Module.cast_input` which ends up I think here:
```
  def cast_input("Elixir." <> _ = value, _) do
    module = Module.concat([value])

    if Code.ensure_loaded?(module) do
      {:ok, module}
    else
      :error
    end
  end
```
and because it's not loaded `:error` is returned.

So one approach would be `:error` -> {:error, "Module #{module} not loaded}` or something like that, or
[...]

PR #1756, opened 2025-02-03
fix: type struct handle instance of return error tuple

ken-kost:
#1754 

Not sure about the error messages, took a guess. I also added `"is invalid"` for the last error. Although I couldn't figure out how to trigger that.
Also should `Elixir.` part be shown 🤔 
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/struct.ex
+++ b/lib/ash/type/struct.ex
@@ -271,7 +271,7 @@ defmodule Ash.Type.Struct do
             {:ok, value}
 
           is_struct(value) ->
-            {:error, "-> constraints: -> instance_of: #{struct}"}
+            {:error, "is invalid"}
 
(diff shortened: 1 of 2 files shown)


lib/ash/type/struct.ex
```diff
@@ -271,7 +271,7 @@ defmodule Ash.Type.Struct do
             {:ok, value}
 
           is_struct(value) ->
-            :error
+            {:error, "-> constraints: -> instance_of: #{struct}"}
```
zachdaniel, line 274: We don't have any errors in this format so for now this should also just be `is invalid`

Issue #72, opened 2020-06-16
Support Through Relationships


andrewcallahan:
**Is your feature request related to a problem? Please describe.**
We should be able to do all types of through relationships. So far I can think of `has_many though` and `has_one through`, but there are more too I think.

**Describe the solution you'd like**
Not sure how it will work, but I'd guess it would be similar to [Ecto](https://hexdocs.pm/ecto/Ecto.Association.HasThrough.html#content)

**Describe alternatives you've considered**
N/A

**Express the feature either with a change to resource syntax, or with a change to the resource interface**

```elixir
many_to_many :grades, MyApp.Grade, through: "school_grades"

has_one :school, MyApp.School, through: "students"
```

**Additional context**
N/A

zachdaniel:
I've had a thought on this, and I think that as long as these relationships are "read only" that it may actually be quite easy to work these out. They only really come into play for side loading. The other limitation may be that we may not be able to use them in ash policy authorizer

zachdaniel:
Currently there are "custom relationships", and "aggregates" to support some of these use cases. However, they aren't as fluid as the option to simply say things like `has_one :organization, Org, through: [:user, :owner]`. This would be a great "medium sized" contribution if anyone is interested in working on it!

ken-kost:
> Currently there are "custom relationships", and "aggregates" to support some of these use cases. However, they aren't as fluid as the option to simply say things like `has_one :organization, Org, through: [:user, :owner]`. This would be a great "medium sized" contribution if anyone is interested in working on it!

I have some interest; not promising anything, might be a bite too big as of yet. 🐛 
Yesterday I did a deep dive. 
One question for now.
`many_to_many` struct has `through` which in `define_schema`:
```
                %{type: :many_to_many} ->
                  many_to_many relationship.name, relationship.destination,
                    join_through: relationship.through,
                    join_keys: [
                      {relationship.source_attribute_on_join_resource,
                       relationship.source_attribute},
                      {relationship.destination_attribute_on_join_resource,
                       relationship.destination_attribute}
                    ]
```
has `join_through: relationship.through,`

I see in ecto that `join_through` can be a direct table or a schema, while `through` is a list of atoms. 
In ash  through has an ash resource as a value.
And for instance Ash Phoenix form auto relays on it. My question is should `through` be renamed to `join_through` so that key `through` would be reserved for list atom values?

Does this make sense?

Also any other ideas, tips and suggestions appreciated. 🙏 

^ @zachdaniel

[...]

PR #1780, opened 2025-02-08
feat: Support Through Relationships

ken-kost:
#72 
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/resource/relationships/belongs_to.ex
```diff
@@ -101,6 +102,10 @@ defmodule Ash.Resource.Relationships.BelongsTo do
                     type: :any,
                     default: Application.compile_env(:ash, :default_belongs_to_type, :uuid),
                     doc: "The type of the generated created attribute. See `Ash.Type` for more."
+                  ],
+                  through: [
+                    type: :any,
```
ken-kost, line 107: I'm thinking spark should support `{:array, :atom}` 🤔 right...?

zachdaniel: `{:list_of, :atom}`

ken-kost: 😮

ken-kost:
is this a feat or enhancement? 🤔😅

zachdaniel:
This would be a feature 😄 . Will have to review next week.

lib/ash/actions/read/relationships.ex
```diff
@@ -1191,6 +1251,9 @@ defmodule Ash.Actions.Read.Relationships do
       has_page? = query.page not in [nil, false]
 
       cond do
+        is_list(relationship.through) ->
+          false
```
zachdaniel, line 1255: So this will actually be the way that we enable data layers to join through relationships. So this should return `true` if the data layer can `:lateral_join_through_relationships` or something. For now, nothing will implement this, so we can leave it as a stub and talk about that implementation later. It is actually quite complex to do 😄

zachdaniel: In fact...yeah returning `false` makes sense here, never mind. When its time to support doing this in an optimized way in the data layer, *then* we'd return `true` here.

lib/ash/actions/read/relationships.ex
```diff
@@ -612,6 +623,50 @@ defmodule Ash.Actions.Read.Relationships do
     )
   end
 
+  defp load_related_records(records, through, cardinality) do
+    load_statement =
```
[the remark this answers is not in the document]
ken-kost: I overengineered it. 👷

zachdaniel:
I'm realizing an issue with our "do it in memory" solution here, which is that limits/offsets in Ash relationships apply "per-related-record". Our in-memory loader here won't be able to provide those same guarantees. I think what we may need to do is make it such that `ash_postgres` and `ash_sql` actually have first class support for `through` relationships 🥶. That will make this a bigger project, but it will also be a major win. 

What that would essentially look like in `ash_postgres` code base is 

- modifying the `run_query_from_lateral_join` code to support arbitrarily long paths. 
- modifying the joining code in `ash_sql` to expand the `through` path automatically whenever it needs to join for any reason (`aggregates.ex`, `join.ex`)

Then in this loader, we'd do the same thing that lateral joining currently does, but setting the full path to the relationship.

What I'm describing is a pretty intense change TBH. I'm happy to support you in the effort, but I imagine it would be quite some time before we can get it where we want it 😓

zachdaniel:
What I mean is that when you do `posts |> Ash.Query.load(comments: Ash.Query.limit(Foo, 10))` you get 10 records per post

ken-kost:
Should this PR be closed then?
So if I understand correctly, first on the list is  enabling arbitrary path length for `lateral_join_query` in `ash_postgres`.
Currently there is
```
  defp lateral_join_query(
         query,
         root_data,
         [{source_query, source_attribute, destination_attribute, relationship}] = path
       ) do
      ...
 end
 ```
and
```
  defp lateral_join_query(
         query,
         root_data,
         [
           {source_query, source_attribute, source_attribute_on_join_resource, relationship},
           {through_resource, destination_attribute_on_join_resource, destination_attribute,
            through_relationship}
         ] = path
       ) do
       ...
  end
```

zachdaniel:
Yep! There is also logic in `ash_sql` that does similar things that needs to be updated. As for closing this PR, my suggestion would instead be to pare back all of the features, leaving just the `through` syntax, and add a resource verifier that checks if the data layer supports a new capability, like `:lateral_join_through` that gets both the resource and the relationship structure.

Then we can merge this PR, despite no one being able to use this new relationship option.

ken-kost:
Man I got to get back to this. 😓 EOW 💯

PR #1799, opened 2025-02-18
Add generic action tests w/ and wo/ return type

jechol:
# Contributor checklist

- [X] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Did you have a bug that drove you to add these tests? They look like reasonable tests, so no reason not to add them, just curious.

zachdaniel:

jechol:
I had been defining a generic action without specifying a return type, using run fn _, _ -> {:ok, value} end, and the value was returned as expected. After a recent update, only :ok was returned. At first, I thought it was an Ash bug and wrote a test case to reproduce the issue.

Later, I found in the documentation that if the return type is not specified, the value is ignored. So, it wasn’t a bug.

Still, a warning message or a crash would have made it easier to catch the issue instead of silently ignoring the value.

jechol:
> Did you have a bug that drove you to add these tests? They look like reasonable tests, so no reason not to add them, just curious.

I had been defining a generic action without specifying a return type, using run fn _, _ -> {:ok, value} end, and the value was returned as expected. After a recent update, only :ok was returned. At first, I thought it was an Ash bug and wrote this test case to reproduce the issue.

Later, I found in the documentation that if the return type is not specified, the value is ignored. So, it wasn’t a bug.

Still, a warning message or a crash would have made it easier to catch the issue instead of silently ignoring the value.

zachdaniel:
I'd accept a PR to raise on an invalid return there 😄

Issue #1628, opened 2024-11-27
4.0: Return a single record from `Ash.read` when the action has `get?: true`


zachdaniel:
Make `get?: true` read actions cause `Ash.read` to return a single record, just like pagination causes it to return a page.

PR #1800, opened 2025-02-18
4.0: Return a single record from `Ash.read` when the action has `get?: true`

JohnDoneth:
Closes https://github.com/ash-project/ash/issues/1628

> Make get?: true read actions cause Ash.read to return a single record, just like pagination causes it to return a page.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [x] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -2191,7 +2191,7 @@ defmodule Ash do
   defp do_read_one(query, action, opts) do
     query
     |> Ash.Actions.Read.unpaginated_read(action, opts)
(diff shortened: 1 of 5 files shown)


zachdaniel:
If we want to introduce this now, we have to gate it behind a compilation flag. Additionally this will almost certainly break other packages calling `Ash.read`. So I think we would be best suited waiting until 4.0 (that's why the 4.0 prefix in the issue name).

Issue #1801, opened 2025-02-18
CaseClauseError when calling a code interface action with `get? true` that returns a BulkResult with many results


sevenseacat:
Wow that was a long title. 

**Describe the bug**

Doing some experimenting with code - if I define a code interface in a domain like this:

```elixir
resource Tunez.Music.ArtistFollower do
  define :unfollow_artist, action: :destroy, args: [:artist], get?: true
end
```

and that action doesn't apply a `filter` change, and affects multiple rows:

```elixir
    destroy :destroy do
      argument :artist, :struct do
        allow_nil? false
        constraints instance_of: Tunez.Music.Artist
      end
    end
```

It raises a big error - 

```
iex(39)> Tunez.Music.unfollow_artist!(artist, actor: user)
[debug] QUERY OK source="artist_followers" db=3.0ms idle=1508.2ms
DELETE FROM "artist_followers" AS a0 RETURNING a0."artist_id", a0."follower_id" []
↳ AshPostgres.DataLayer.destroy_query/4, at: lib/data_layer.ex:1776
** (CaseClauseError) no case clause matching: %Ash.BulkResult{status: :success, errors: [], 
records: [#Tunez.Music.ArtistFollower<...>, #Tunez.Music.ArtistFollower<...>], ...}
    (tunez 0.1.0) deps/ash/lib/ash/code_interface.ex:1315: Tunez.Music.unfollow_artist!/3
    iex:39: (file)
```

It *should* raise an error, but an actual error that rolls back the change, not a CaseClauseError :D

**Expected behavior**

Looking through the code I can see a `Ash.Error.Invalid.MultipleResults` eror type, which looks pretty appropriate for this case. An error of this type should be raised, and the transaction should roll back, and I should be told off because I wrote silly code and forgot to apply a filter (well I didn't, but I didn't realize that for a destroy action it has to be a `change` not a normal filter, but that's an entirely different issue)

**Runtime**
 - Elixir version 1.18.2-otp-27
 - Erlang version 27.2.1
 - OS macOS Sonoma
 - Ash version 3.4.64 (latest as of now)

ken-kost:
One place where I saw a potential fix is in `Ash.Actions.Destroy.Bulk` 
in `run(domain, %Ash.Query{} = query, action, input, opts, not_atomic_reason)` there is
```
          case Ash.Actions.Read.unpaginated_read(query, query.action, read_opts) do
            {:ok, results} ->
              run(
                domain,
                results,
                action,
[...]

PR #1806, opened 2025-02-19
fix: Missing case clause for bulk update/destroy with `get?: true` in interface

ken-kost:
#1801 
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
I think that two of these are meant to `raise Ash.Error.to_error_class(<the_error>)` as they are in the `act!` calls, which means they raise any errors.

zachdaniel (approved):

PR #1817, opened 2025-02-23
Generic transactional actions: Don't clear input.resource from resource list in case of rollback.

Torkan:
When running a generic action with `transaction? true`, this fix ensures the input.resource is not wiped from the list of resources to issue a rollback for if the transaction encounters an error.

Since the Ets-datalayer used in the test-suite doesn't have proper transaction/rollback-support, I've set up a repro which demonstrates the issue here instead: https://github.com/Torkan/ash-generic-tx-action-repro

Hopefully this can be merged as-is, and if there are any unintended consequences I'm not aware of, at least I shed some light on the issue 😄


zachdaniel:

PR #1843, opened 2025-03-07
fix: properly handle change with `where` validations in bulk

vonagam:
Was getting error about `nil` not implementing enumerable protocol. It was because `changes` map won't contain changes that did not pass their conditional `where` validations.

Also included a formatting fix, a credo fix, a typo fix and a cheat sheets fix.


zachdaniel (approved):

zachdaniel:

PR #1853, opened 2025-03-12
improvement: allow is_equal and is_not_equal for compare validation

### LINKED ISSUE #1846: add equal and not_equal to compare validator
```ex
compare(:user_id, not_equal: :connected_user_id)
compare(:user_id, equal: :connected_user_id)
```

the ability to add these two conditions for compairing

writing my own

For example

```elixir
    validate compare(:user_id, not_equal: :connected_user_id),
      where: [present(:user_id), present(:connected_user_id)],
      message: "cannot create activity for yourself"
```

**Additional context**

lifeiscontent:
closes #1846

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/validation/compare.ex
+++ b/lib/ash/resource/validation/compare.ex
@@ -8,24 +8,39 @@ defmodule Ash.Resource.Validation.Compare do
 
   @opt_schema [
     greater_than: [
-      type: {:or, [:integer, :atom, :float, {:struct, Decimal}, {:fun, 0}]},
+      type: {:or, [:any, {:tagged_tuple, :value, :any}, {:tagged_tuple, :ref, :atom}]},
       required: false,
       doc: "The value that the attribute should be greater than."
     ],
     greater_than_or_equal_to: [
-      type: {:or, [:integer, :atom, :float, {:struct, Decimal}, {:fun, 0}]},
+      type: {:or, [:any, {:tagged_tuple, :value, :any}, {:tagged_tuple, :ref, :atom}]},
       required: false,
       doc: "The value that the attribute should be greater than or equal to"
     ],
     less_than: [
-      type: {:or, [:integer, :atom, :float, {:struct, Decimal}, {:fun, 0}]},
+      type: {:or, [:any, {:tagged_tuple, :value, :any}, {:tagged_tuple, :ref, :atom}]},
       required: false,
       doc: "The value that the attribute should be less than"
     ],
     less_than_or_equal_to: [
-      type: {:or, [:integer, :atom, :float, {:struct, Decimal}, {:fun, 0}]},
+      type: {:or, [:any, {:tagged_tuple, :value, :any}, {:tagged_tuple, :ref, :atom}]},
       required: false,
       doc: "The value that the attribute should be less than or equal to"
+    ],
+    is_equal: [
+      type: {:or, [:any, {:tagged_tuple, :value, :any}, {:tagged_tuple, :ref, :atom}]},
+      required: false,
+      doc: "The value that the attribute should be equal to"
+    ],
+    is_not_equal: [
+      type: {:or, [:any, {:tagged_tuple, :value, :any}, {:tagged_tuple, :ref, :atom}]},
+      required: false,
+      doc: "The value that the attribute should not be equal to"
+    ],
+    is_nil: [
+      type: :boolean,
+      required: false,
+      doc: "Whether the attribute should be nil (true) or not nil (false)"
     ],
     attribute: [
       type: :atom,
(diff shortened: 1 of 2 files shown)


test/resource/validation/compare_test.exs
```diff
@@ -32,7 +32,396 @@ defmodule Ash.Test.Resource.Validation.CompareTest do
...
+    # test "atoms can be compared" do
+    #   {:ok, opts} = Compare.init(attribute: :atom_value, is_equal: :foo)
+    #   changeset = Post |> Ash.Changeset.for_create(:create, %{atom_value: :foo})
+
+    #   assert :ok = Compare.validate(changeset, opts, %{})
+
+    #   changeset = Post |> Ash.Changeset.for_create(:create, %{atom_value: :bar})
+
+    #   assert_error(changeset, opts, "must be equal to foo")
+    # end
```
lifeiscontent, line 209: @zachdaniel is this due to atoms being used to be able to lookup fields in a resource?

test/resource/validation/compare_test.exs
```diff
@@ -32,6 +32,396 @@ defmodule Ash.Test.Resource.Validation.CompareTest do
...
+    test "nil can be compared" do
+      # Use a tuple to represent a literal nil value
+      {:ok, opts} = Compare.init(attribute: :nil_value, is_equal: {:literal, nil})
+      changeset = Post |> Ash.Changeset.for_create(:create, %{nil_value: nil})
+
+      assert :ok = Compare.validate(changeset, opts, %{})
+
+      changeset = Post |> Ash.Changeset.for_create(:create, %{nil_value: "not nil"})
+
+      assert_error(changeset, opts, "must be equal to ")
```
lifeiscontent, line 243: should we do some extra special casing for this scenario? I guess you'd want to see "must be equal to nil"

lib/ash/resource/validation/compare.ex
```diff
@@ -169,18 +231,50 @@ defmodule Ash.Resource.Validation.Compare do
...
 
+  defp format_value(value) when is_list(value) do
+    items = Enum.map_join(value, ", ", &format_value/1)
+    "[#{items}]"
+  end
+
+  defp format_value(value) when is_map(value) do
+    items =
+      value
+      |> Enum.sort_by(fn {k, _v} -> to_string(k) end)
```
lifeiscontent, line 259: sorting here, so the keys are stable

test/resource/validation/compare_test.exs
zachdaniel: We probably should not allow `equal_to: nil`, but should instead require `is_nil: true/false`. Its unlikely in any other case you'd use `nil`, (because comparing to `nil` is weird in these contexts) but we should probably forbid it as a value in any of these options.

sevenseacat:
Is this for attribute comparison (ie. comparing two attributes) or comparing an attribute with a static value, or both? I'm a bit confused while reading over the code. It looks like it does both, but then you would have an issue if you wrote something like `compare(:foo, is_equal: :bar)` do you mean the static value `:bar` or the attribute `:bar`?

lifeiscontent:
@sevenseacat @zachdaniel and I discussed in discord, and came up with a convention of `{:value, :bar}` for comparing static values, and potentially in the future we'll do `{:ref, :thing_in_db}` and migrate away from `{:value, :bar}`

zachdaniel:
yes I think we should make `{:value, ...}` behave the same for all of the options 🙇

lib/ash/resource/validation/compare.ex
```diff
@@ -8,25 +8,40 @@ defmodule Ash.Resource.Validation.Compare do
 
   @opt_schema [
     greater_than: [
-      type: {:or, [:integer, :atom, :float, {:struct, Decimal}, {:fun, 0}]},
+      type: :any,
```
lifeiscontent, line 11: @zachdaniel I wasn't sure how to define the type correctly for this now that we want to support `{:value, number}` for this. if you have suggestions on how to update this, happy to add it

zachdaniel: Lets do something like this (for docs) `{:or, [:any, {:tagged_tuple, :value, :any}, {:tagged_tuple, :ref, :atom}]}`

lib/ash/resource/validation/compare.ex
```diff
@@ -63,34 +78,15 @@ defmodule Ash.Resource.Validation.Compare do
...
-              do: invalid_attribute_error(opts, value)
-
-          {:less_than_or_equal_to, attribute} ->
-            if !Comp.less_or_equal?(value, attribute_value(changeset, attribute)),
-              do: invalid_attribute_error(opts, value)
-        end) || :ok
+      {:ok, value} ->
+        validate_value(changeset, opts, value)
 
       _ ->
```
zachdaniel, line 84: This needs to be `{:ok, nil}`, w/ the fall through case still returning `:ok`.

zachdaniel (dismissed):

lib/ash/resource/validation/compare.ex
```diff
@@ -157,6 +159,52 @@ defmodule Ash.Resource.Validation.Compare do
...
+             error(^InvalidAttribute, %{
+               field: ^opts[:attribute],
+               value: ^atomic_ref(opts[:attribute]),
+               message: ^(context.message || "must be nil"),
+               vars: %{field: ^opts[:attribute]}
+             })
+           )}
+
+        {:is_nil, false} ->
+          {:atomic, [opts[:attribute]], expr(^atomic_ref(opts[:attribute]) == nil),
```
zachdaniel, line 199:
```suggestion
          {:atomic, [opts[:attribute]], expr(is_nil(^atomic_ref(opts[:attribute]))),
```

lib/ash/resource/validation/compare.ex
```diff
@@ -157,6 +159,52 @@ defmodule Ash.Resource.Validation.Compare do
...
+             error(^InvalidAttribute, %{
+               field: ^opts[:attribute],
+               value: ^atomic_ref(opts[:attribute]),
+               message: ^(context.message || "must not be equal to %{is_not_equal}"),
+               vars: %{field: ^opts[:attribute], is_not_equal: ^atomic_value(value)}
+             })
+           )}
+
+        {:is_nil, true} ->
+          {:atomic, [opts[:attribute]], expr(^atomic_ref(opts[:attribute]) != nil),
```
zachdaniel, line 188:
```suggestion
          {:atomic, [opts[:attribute]], expr(not(is_nil(^atomic_ref(opts[:attribute])))),
```

PR #1854, opened 2025-03-13
fix: properly handle bulk soft destroy

vonagam:
`Ash.Actions.Update.run` [can return `:ok`](https://github.com/ash-project/ash/blob/0761bbbaaab11e22ab33f7ee7a919473f7ebb33c/lib/ash/actions/update/update.ex#L391) in case of a soft destroy. So such return variant needs to be handled in processing of bulk update.

(Typespec for `run` does not list simple `:ok` as a potential result right now. Didn't know if I should have added it in this PR.)

### DIFF
--- a/lib/ash/actions/update/bulk.ex
+++ b/lib/ash/actions/update/bulk.ex
@@ -1677,6 +1677,11 @@ defmodule Ash.Actions.Update.Bulk do
                  action,
                  Keyword.put(opts, :atomic_upgrade?, false)
                ) do
+            :ok ->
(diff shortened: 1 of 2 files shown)


zachdaniel:
Would be great to have the type spec fixed here too 🙇

vonagam:
Ok, added `:ok` to update run typespec. I see that destroy run also has `{:ok, list}` in typespec but briefly looking at update action I do not see such possibility for update, so did not add it, but maybe I missed something.

zachdaniel:

PR #1866, opened 2025-03-16
Per process actor and tenant

ademenev:
In web applications, tenant and actor are typically  determined when authenticating a request, and each request runs in a separate process. This patch allows setting per-process default actor and tenant

# Contributor checklist

- [x] Documentation changes

### DIFF
--- a/documentation/topics/advanced/multitenancy.md
+++ b/documentation/topics/advanced/multitenancy.md
@@ -154,3 +154,16 @@ end
 ```
 
 This allows you to pass an `%Organization{}` or an organization_id around, and have that `organization_id` properly used with attribute and context-based multitenancy.
+
+## Per-process tenants
+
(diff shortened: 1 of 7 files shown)


zachdaniel:
I appreciate the PR, but we had this feature before and explicitly removed it. It comes with all kinds of difficult problems with how you handle the internals of actions when actions have hooks that call other actions, as well as makes it easy to make a mistake like using `Task.async` and now your actor is gone.

again, I really appreciate the effort and thought put into the PR, but we can't move forward with it for the reasons described above 🙇‍♂️🙇‍♂️🙇‍♂️

ademenev:
@zachdaniel   It's ok :) Even though setting tenant and actor each time  is annoying and tedious

Probably we could discuss alternative approaches. Ill give it a thought and reach out on Slack.

PR #1867, opened 2025-03-16
Fix erroneous policy group example

sodapopcan:
This fixes an error-throwing example in the policy docs (see [here](https://elixirforum.com/t/ensure-actor-can-only-create-resource-for-themselves/69988?u=sodapopcan)).

I can also add a note (either code comment above the `create` example or admonition block below) about not being able to use field references in create policies but I'm not sure this is the right place.

### DIFF
--- a/documentation/topics/security/policies.md
+++ b/documentation/topics/security/policies.md
@@ -163,7 +163,7 @@ policies do
     end
 
     policy action_type(:create) do
-      authorize_if always()
+      authorize_if relating_to_actor(:owner)
     end
   end
 end
--- a/documentation/topics/security/policies.md
+++ b/documentation/topics/security/policies.md
@@ -158,8 +158,12 @@ policies do
       authorize_if expr(owner_id == ^actor(:id))
     end
 
-    policy action_type([:create, :update, :destroy]) do
+    policy action_type([:update, :destroy]) do
       authorize_if expr(owner_id == ^actor(:id))
+    end
+
+    policy action_type(:create) do
+      authorize_if always()
     end
   end
 end


documentation/topics/security/policies.md
```diff
@@ -158,9 +158,13 @@ policies do
       authorize_if expr(owner_id == ^actor(:id))
     end
 
-    policy action_type([:create, :update, :destroy]) do
+    policy action_type([:update, :destroy]) do
       authorize_if expr(owner_id == ^actor(:id))
     end
+
+    policy action_type(:create) do
+      authorize_if always()
```
zachdaniel, line 166:
```suggestion
      authorize_if relating_to_actor(:owner)
```

zachdaniel:
I *think* we have somewhere in the guide describing that? Welcome addition if not. One suggestion but otherwise looks good 😄

sodapopcan:
Oh wow ok, so I actually misunderstood from EF.  So `relating_to_actor` actually works if `user_id` is given as a param!  I thought because you couldn't do a filter-type expression that a resource just didn't have access to param data, but I clearly don't have a good mental model of how filter expressions work.  Anyway, this is so cool!

It's likely in there somewhere.  If I find it I may add a note about here later but for now you can merge.  I have some other doc stuff I'm working on anyway.

zachdaniel:

zachdaniel:
If you're in the discord LMK so I can give you the contributor role (unless this isn't your first contribution)

sodapopcan:
I'm not but I'll join!  This is my first contribution, hopefully not my last.  I had a bit of a false start with Ash after not being able to sell my CTO on it (it was sorta dead in the water before I could even make a POC), but I have free hobby time now so finally building my first app in it that I aim to take to some semblance of polish.

Issue #2, opened 2019-10-05
Initial Thoughts


andrewcallahan:
**Feature Roadmap**
* Filtering
* Pagination
* Relationships
* Attributes
* JSONAPI
* GraphQL
* Optional Ecto
* Custom Errors
* Escape Hatches

**Niche Issues to Be Aware Of**
* Regular controllers for things such as Twilio callback or pdf generation
* Non-RESTful controller actions that completely break the CRUD pattern
* Custom callbacks before saving/creating such as setting the current user as a property on a model for whodunnit purposes
* Forced sideloading data (this could be more of a function of ember-data being strange)
* Limiting sorting/creating/updating/filtering/attribute serializing based on custom rules (ie: user type)
* Make sure we handle derived IDs (ie on database view) just as nicely as DB generated IDs
* Allow for default values (ie: students are always sorted by last name descending)
* Decorator methods/custom attributes (ie: full name)
* Filter by attribute or create a custom filter
* Multiple associations between the same two tables with different foreign key names (ie: created by user vs updated by user)
* Self associations (ie: users friending users)

**Assumptions:**
* almost all schemas and changesets currently defined within contexts will move to be defined in resources
* all boilerplate autogenerated CRUD code in context files such as School.get_list(id) will be removed
* almost all controllers and views will be removed
* almost all routes in the router file will be removed

**What about...**
* what happens to custom functions defined in a context (ie: import CSV logic)?
* what happens to autogenerated tests for a schema (and would we do this for a resource)?
* what happens to autogenerated tests for a controller (and would we do this for a resource)?
* where would CORS be handled?
* what about Phoenix Channels
* where would error handling/fallback controller happen?

**I'm not clear on**
* The interplay between Ash Resources and Phoenix Contexts - it can't replace 100% of what Contexts do, can it/should it?
* can ecto schemas access database views instead of tables?
* where to define/how to access current user?
* how to determine authenticated vs unauthenticated routes?

**A few specific and somewhat difficult examples we should have a story about**
* Full Text Search on either a db table or multiple db tables
* User registration (non 1-1 db backed endpoint)
* file downloads and uploads
* non CRUD endpoint such as /api/teachers/1/activate

**TODO**
* start building documentation
* start unit testing one isolated piece of the code
* make an integration test for one resource
* CLI
* build out a small demo app for learning purposes
* use Ash to build a substantial API
* get feedback from DY Elixir people, and people like Dan/Larry Gebhardt
[...]

PR #1869, opened 2025-03-17
Fix bulk_create for manual create actions with bulk_create/3

Torkan:
Fixes the handling of results from the bulk_create/3 for Ash.Resource.ManualCreate,

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -1206,9 +1206,6 @@ defmodule Ash.Actions.Create.Bulk do
                   })
                   |> Ash.Actions.Helpers.rollback_if_in_transaction(resource, nil)
                   |> case do
-                    {:ok, results} ->
-                      results
-
                     :ok ->
                       if opts[:return_records?] do
                         raise "`#{inspect(mod)}.bulk_create/3` returned :ok without a result when `return_records?` is true"
@@ -1220,9 +1217,22 @@ defmodule Ash.Actions.Create.Bulk do
                       store_error(ref, error, opts)
                       []
 
-                    {:notifications, notifications} ->
-                      store_notification(ref, notifications, opts)
-                      []
+                    results when is_list(results) ->
+                      ok_results =
+                        Enum.reduce(results, [], fn
+                          {:ok, result}, results ->
+                            [result | results]
+
+                          {:ok, result, notification}, results ->
+                            store_notification(ref, notification, opts)
+                            [result | results]
+
+                          {:error, error}, results ->
+                            store_error(ref, error, opts)
+                            results
+                        end)
+
+                      {:ok, ok_results}
                   end
                 else
                   [changeset] = batch
--- a/lib/ash/resource/manual_actions/manual_create.ex
+++ b/lib/ash/resource/manual_actions/manual_create.ex
@@ -54,6 +54,7 @@ defmodule Ash.Resource.ManualCreate do
               list(
                 :ok
                 | {:ok, Ash.Resource.record()}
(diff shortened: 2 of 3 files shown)


zachdaniel:
This looks good to me. Want to do the other action types in a separate PR?

Torkan:
> This looks good to me. Want to do the other action types in a separate PR?

Up to you! 😄 I won't be able to finish the others before tomorrow or the day after at the earliest though, but I'm eager to get this working so I'll try to do it as quickly as I can 😅

zachdaniel:

vonagam:
Hm... looking at the `Enum.reduce` cases in this PR it does not seem to match `bulk_create` callback typespec and other stuff:

1) Typespec says that list can contain `:ok` and reduce does not have a case for that.
2) Typespec says that third argument is a map with `notifications` key but the argument gets passed into `store_notification` but only notifications should be passed there, not the map.
3) Typescpec says that list can contain `{:notifications, notifications}` but reduce does not handle such case.

Am I missing something?

zachdaniel:
Nope, its on my list to add the other cases in

zachdaniel:
I didn't notice #2 though. @Torkan if you could work up some tests for this as well that would be much appreciated.

Torkan:
> I didn't notice #2 though. @Torkan if you could work up some tests for this as well that would be much appreciated.

Ah, I was a bit tired when updating that typespec guess, I meant to just define the "standard" `{:ok, record, notifications}` return value 😅

zachdaniel:
We do want it to return the `%{notifications: ...}`. I'll fix #2 while I add the other cases in.

zachdaniel:
@vonagam feel like that one would have been hard to find when someone reported it lol. Thanks.

vonagam:
I think you do want typespec for callback to be `{:ok, record, notifications}` not map. Like in manual destroy or after action hook. Variant with `%{notifications: ...}` in third element is used in internal representation only.

zachdaniel:
Hmm... we use %{notifications: []} style for after action hooks, as a public interface, so I'm not really sure. We could choose either here as a result of it already being inconsistent.

vonagam:
Typespec for after action hook:
```elixir
@type after_action_fun ::
  (t, Ash.Resource.record() ->
     {:ok, Ash.Resource.record()}
     | {:ok, Ash.Resource.record(), [Ash.Notifier.Notification.t()]}
     | {:error, any})
```

But I see that manual update unlike manual destroy does have a map with notifications key.

zachdaniel:
*sigh* this got pretty borked didn't it. It obviously never worked so we are free to change the signatures for the bulk callbacks for manual actions IMO. @Torkan has mentioned making the others, destroy/create, consistent with the new stuff.  I've fixed the inconsistencies w/ the type specs here. 

And yeah sorry I was thinking of `before_action`, sorry:

```elixir
  @type before_action_fun :: (t -> t | {t, %{notifications: [Ash.Notifier.Notification.t()]}})

```

zachdaniel:
I prefer `%{notifications: [...]}` personally because it allows us to expand the API over time w/o needing to add new structural returns.

Torkan:
> I prefer `%{notifications: [...]}` personally because it allows us to expand the API over time w/o needing to add new structural returns.

Makes sense! I'll just update all the manual bulk actions to adhere to that spec then, settled?

zachdaniel:
That's a very good idea 👌 can you open an issue for it?

vonagam:
Done #1877.

PR #1871, opened 2025-03-17
Reproduce decimal inequality comparison bug

jechol:
This pull request adds a test case to reproduce an issue where the decimal inequality check raises an error. No fix is included in this PR—only the reproduction of the bug.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/resource/validation/compare_test.exs
+++ b/test/resource/validation/compare_test.exs
@@ -227,6 +227,16 @@ defmodule Ash.Test.Resource.Validation.CompareTest do
       assert_validation_error(changeset, opts, "must not be equal to number_two")
     end
 
+    test "validates inequality with decimals" do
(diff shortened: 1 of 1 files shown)


lib/ash/resource/validation/compare.ex
```diff
@@ -225,6 +225,7 @@ defmodule Ash.Resource.Validation.Compare do
         |> Enum.map(fn {k, v} ->
           case v do
             fun when is_function(fun, 0) -> {k, fun.()}
+            %Decimal{} = v -> {k, v}
             v when is_list(v) -> {k, format_value(v)}
             v when is_map(v) -> {k, format_value(v)}
```
zachdaniel, line 230: The actual fix there should be `is_struct` to cover any struct since it's the `is_map` that's causing the bug.

zachdaniel:

PR #1873, opened 2025-03-17
test: check aggregation

rapidfsub:
This is just a failing test.
The last one should pass as other cases do.


zachdaniel:
Hello, could you provide a bit more context on what is going wrong here?

zachdaniel:
nvm I have a fix. Thanks for the test 🙇

PR #1878, opened 2025-03-18
Add failing test for ignored calculation arguments

jechol:
The issue has been present since commit `5665ec59` when `keep_read_action_loads_when_loading?` is true.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -34,7 +34,7 @@ if Mix.env() == :test do
 
   config :ash, :custom_expressions, [Ash.Test.Expressions.JaroDistance]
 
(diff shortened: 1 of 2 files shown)


zachdaniel:

zachdaniel:
Looking into this, it's not really a bug. Without the new option, the load statement from the read action is applied, and it load `:sum` with its default arguments. That's part of the reason we introduced this new behavior in the first place, because read action loads can supersede provided loads if the read action hasn't already been validated for the action with `Ash.Query.for_action`

PR #1881, opened 2025-03-18
fix: apply strict load in `Ash.get` properly

vonagam:
`load` option [will be applied](https://github.com/ash-project/ash/blob/d5f4c2a31ec04537107390e6ff2a404ba9191a28/lib/ash/actions/read/read.ex#L45) in `unpaginated_read` (it is in `read_opts`) and there it will take into account presence of `strict?` option.

Had a problem where `load: [relationship: [:field]]` was loading other attributes than `field` on `relationship`.


zachdaniel:

PR #1883, opened 2025-03-18
Improve processing of manual actions during bulk operations

Torkan:
Unifies the typespecs for bulk action in ManualCreate/Update/Destroy behaviours, and updates the bulk-modules to handle each of the possible return values.

I did notice some unexpected behavior while writing the tests, namely that the number of returned notifications were twice the expected amount.

I also noticed that BulkResult's records-field is set to `nil` when passing `return_records?: false`, but the notifications-field will be set to `[]` if `return_notifications?: false` is passed.

The notifications-field is however set to `nil` during the test that runs a manual action that only returns `:ok`

I'll add some comments below to point out  where these things happen.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/actions/create/bulk.ex
```diff
@@ -1177,7 +1177,7 @@ defmodule Ash.Actions.Create.Bulk do
         |> Enum.flat_map(fn {_atomics, batch} ->
           result =
             case action.manual do
-              {mod, opts} ->
+              {mod, _mod_opts} ->
```
Torkan, line 1180: This shadowed the original opts passed to the function.

test/actions/bulk/bulk_create_manual_test.exs
```diff
@@ -80,24 +230,173 @@ defmodule Ash.Test.Actions.BulkCreateManualTest do
...
+    result =
+      [%{name: "Author1"}, %{name: "Author2"}, %{name: "Author3"}]
+      |> Ash.bulk_create(Author, :create_manual_notification_list,
+        return_notifications?: true,
+        return_errors?: true,
+        return_records?: true
+      )
+
+    assert Enum.count(result.records) == 3
+    assert Enum.count(result.notifications) == 6
```
Torkan, line 243: Here is an example of 6 notifications being returned. I assume this is probably due notifications also being generated by the create-action being called inside bulk_create.

test/actions/bulk/bulk_create_manual_test.exs
```diff
@@ -80,24 +230,173 @@ defmodule Ash.Test.Actions.BulkCreateManualTest do
...
+    result =
+      [%{name: "Author1"}, %{name: "Author2"}, %{name: "Author3"}]
+      |> Ash.bulk_create(Author, :create_manual_tupled_notifications,
+        return_errors?: true,
+        return_records?: true,
+        return_notifications?: true
+      )
+
+    assert Enum.count(result.records) == 0
     assert Enum.count(result.notifications) == 3
```
Torkan, line 353: Here the amount of return notifications are 3 again.

lib/ash/actions/create/bulk.ex
zachdaniel: The `.bulk_create` function should be using the `mod_opts` there, not the opts passed into the function.

test/actions/bulk/bulk_create_manual_test.exs
```diff
@@ -80,24 +230,173 @@ defmodule Ash.Test.Actions.BulkCreateManualTest do
...
+    result =
+      [%{name: "Author1"}, %{name: "Author2"}, %{name: "Author3"}]
+      |> Ash.bulk_create(Author, :create_manual_ok,
+        return_errors?: true,
+        return_records?: false,
+        return_notifications?: true
+      )
+
+    assert result.records == nil
+    assert result.notifications == nil
```
Torkan, line 381: Notifications are set to nil here, even though return_notifications? is set to false. No biggie, just a little unexpected/erratic 😅

lib/ash/actions/create/bulk.ex
Torkan: Ah yes sorry, it should be the 2nd arg to bulk_create ofc 🙈

zachdaniel:
This is great 😄 some inconsistencies you've found may need to be addressed/looked into. Really appreciate the thorough testing as well 🙇

zachdaniel:
I will review this in the next couple of days.

test/actions/bulk/bulk_create_manual_test.exs
zachdaniel: Yeah, this does actually make sense, you get the notifications we generate for the manual create as well as for whatever you do in the callback.

test/actions/bulk/bulk_create_manual_test.exs
zachdaniel: However, can we investigate why there are only 3 notifications in this instance?

zachdaniel:
Alright, really only one question to be answered and then we can get it going. I see a test failure (you can ignore the cheat_sheets issue, I haven't figured that out yet), so make sure to pull `main` and resolve that failure if its part of this PR.

zachdaniel:
One other comment WRT why ther are only 3 other notifications.

Torkan:
> One other comment WRT why ther are only 3 other notifications.

Aaaah, gotcha! 😅
I'll look into it!

zachdaniel:

PR #1885, opened 2025-03-19
test: ignore global and action load

rapidfsub:
This is test case for recent updated `ignore action-defined loads when using Ash.load` feature.
(https://hexdocs.pm/ash/changelog.html#v3-4-66-2025-03-03)

### DIFF
--- a/test/ash/load_ignorance_test.exs
+++ b/test/ash/load_ignorance_test.exs
@@ -1,0 +1,85 @@
+alias Ash.Test.LoadIgnoranceTest, as: ThisTest
+alias Ash.Changeset
+
+defmodule ThisTest.Post do
+  use Ash.Resource,
+    domain: Ash.Test.Domain,
+    data_layer: Ash.DataLayer.Ets
+
+  actions do
+    defaults [:read, create: :*]
+
+    read :read_with_comment_count do
+      prepare build(load: [:comment_count])
+    end
+
+    update :append_comment do
+      require_atomic? false
+      argument :comment, :map, allow_nil?: false
+      change manage_relationship(:comment, :comments, type: :create)
+    end
+  end
+
+  preparations do
+    prepare build(load: [:has_comment])
+  end
+
+  attributes do
+    uuid_v7_primary_key :id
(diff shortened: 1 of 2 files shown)


zachdaniel:
I believe we do have a test for this already in `load_test.exs`. Could you double check?

rapidfsub:
@zachdaniel  You're right. I missed it.
But, I have a issue related to this test.
I copied this code from below.
https://github.com/rapidfsub/shule/blob/develop/emil/test/emil/load_ignorance_test.exs

This test just passes inside ash repository, but some deadlock-ish things happen in my repository.
You can run the test in emil directory after cloning my whole repository.

I reproduced it using elixir versions below.
 1.17.3-otp-27
 1.18.0-otp-27
 1.18.3-otp-27

rapidfsub:
It happens on line 76.

rapidfsub:
Sorry. I need to write more comment about this. Ignore until I write another one please.

rapidfsub:
> Using the new configuration solves this issue in your reproduction. My suggestion is to update to the new configuration:
> 
> ```elixir
> config :ash, keep_read_action_loads_when_loading?: false
> ```

I didn't know this page existed.
https://hexdocs.pm/ash/backwards-compatibility-config.html#keep_read_action_loads_when_loading

rapidfsub:
@zachdaniel Sorry for poor explanation, but I don't think I could give more information.
Because the test code just freezed until timeout, and I cannot figure out what was happening.
So, I gave you the freezed line number only.

Or, would you give me an advice about 'it would be helpful if you gave me this kind of information' thing?

zachdaniel:
Part of the new config is to solve recursive load problems from `Ash.load` etc. So ideally you would use the new config and that would be the "fix" here 😄

rapidfsub:
No, the bug is entirely fixed by that config.

rapidfsub:
> Please ensure to include details on the specific issues where possible. I understand you have a failure but I don't know what I'm looking for.

I was just asking your small advice, related to this comment.

zachdaniel:
Oh, no worries I understand what you mean that you don't have much to share.

PR #1891, opened 2025-03-19
test: CiString updates even if same casing with force_change_attribute

quartz55:
This was working a couple of versions ago, so adding a test to catch any future regressions.

### DIFF
--- a/test/type/ci_string_test.exs
+++ b/test/type/ci_string_test.exs
@@ -130,4 +130,26 @@ defmodule Ash.Test.Type.CiString do
              |> Ash.Query.filter(string_f == "FoObAr")
              |> Ash.read!()
   end
+
+  test "updates if force_change_attribute" do
+    p =
+      Post
+      |> Ash.Changeset.for_create(:create, %{string_f: "foobar"})
+      |> Ash.create!()
+
+    p =
+      p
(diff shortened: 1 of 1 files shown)


zachdaniel:
Are you saying that you want it to update even though it compares equal? We actually had a separate issue that caused us to revert to the original behavior. The issue is that `force_change_attribute` could have a few meanings, but was originally intended to mean "change it even if the action rules or current phase of the changeset would prevent you from doing so", not necessarily "change even if the value isn't changing". So to change a CiString regardless of comparing equal (I assume to change the underlying representation) you may need to do something else. An unfortunate an unintended breaking change. One thing you could do in the short term is just `Map.put` into `changeset.attributes`.

Some options moving forward could be to add `opts` to `change_attribute` and `force_change_attribute` so you can say something like `even_if_equal?: true` or something along those lines. Alternatively we could add a new callback to `Ash.Type` for comparing equality for changing vs comparison. That way `Ash.Type.CiString` would allow you to set new casings when changing but compare them as equal otherwise. The first one is the least invasive and least likely to lead to other surprising results, so I'm open to a proposal for it 🙇 
 
For now though we'll have to stick with the current behavior as it was actually a bug that it originally worked. Sorry about that and thank you for the PR 🙇

PR #1896, opened 2025-03-21
Add test: Read action loads of nested relationships are respected with `Ash.load`

jechol:
The test right above verifies that Ash.load ignores the read action's load when called on an already-loaded record.
This test, on the other hand, ensures that when loading a nested relationship, the read action's load is not ignored.

I had mistakenly assumed that the read action's load would also be ignored for nested relationships until I wrote this test.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #1897, opened 2025-03-21
Add failing test: calculation arguments should be respected for nested loads

jechol:
This test reproduces an issue where arguments are correctly applied when calling
`post |> Ash.load!(sum: [a: 3, b: 4])`,
but not applied when loading from a parent record like:
`author |> Ash.load!(posts: [sum: [a: 3, b: 4]])`.

You can reproduce the issue by running:
`mix test test/actions/load_test.exs:2024`

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/actions/load_test.exs
+++ b/test/actions/load_test.exs
@@ -2011,16 +2011,24 @@ defmodule Ash.Test.Actions.LoadTest do
     assert %Ash.NotLoaded{} = Ash.load!(post, :author).category_length
   end
 
-  test "calculation arguments should be respected" do
-    post =
-      Post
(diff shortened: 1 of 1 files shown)


zachdaniel:
I think we can potentially change it to new behavior but only if the new flag is set.

zachdaniel:
Yeah, okay in the context of the new flag, I would call this a bug. "working as intended" with the flag set to `false` though.

jechol:
Thanks, @zach — I’ve resolved the conflicts! ✅

zachdaniel:

PR #1903, opened 2025-03-23
Some improvements around bulk operations on manual actions.

Torkan:
This fixes a few things concerning bulk operations on manual actions:
- Ensures batch_size is correctly set, which earlier caused actions to raise if no bulk action was defined in the manual action module.
- Ensures error is raised if action doesn't return records when return_records?: true (both for :ok-returns, and {:notifications, notifications}-returns

Also it makes all bulk operations be a little more consistent with the returned BulkResult:
- When return_notifications?: false, notifications are always set to nil.
- When return_records?: false, records are always set to nil.
- When return_errors?: false, records are always set to nil.

Otherwise, these fields will always be a list.

# Contributor checklist

- [ x] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #1908, opened 2025-03-25
Failing test: update unloads already loaded relationships

jechol:
I've found that calling `Ash.update` causes already-loaded relationships to be replaced with `%Ash.NotLoaded{}`. I'm wondering if this is the intended behavior.

I noticed this because previously the loaded relationships were preserved after an update, but after a recent update, they are being set to `%Ash.NotLoaded{}`, which led to a raise in our code.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/actions/update_test.exs
+++ b/test/actions/update_test.exs
@@ -594,6 +594,29 @@ defmodule Ash.Test.Actions.UpdateTest do
   end
 
   describe "load" do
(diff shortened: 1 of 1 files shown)


zachdaniel:
Did you discover this with the postgres data layer? Or specifically the ETS data layer?

PR #1912, opened 2025-03-25
improvement: support `expr(exists(relation))`

vonagam:
Just allows omitting `true` as the last argument for `exists` in `expr`.

Any downsides that I don't see?


zachdaniel:
No downsides that I can see, makes perfect sense to me 👍

zachdaniel:

Issue #1899, opened 2025-03-22
Ash Custom Types Constraints Validations


djcarpe:
**Describe the bug**
There are no compiler errors to misconfiguring the constraints fields keyword lists.

**To Reproduce**
Create a new type with an incorrect type keyword list.

```elixir
defmodule MyApp.FooType do
  alias MyApp.BarResource
  use Ash.Type.NewType,
    subtype_of: :map,
    constraints: [
      fields: [
        bars: [type: :struct, instance_of: BarResource]
      ]
    ]
end
```

The correct keyword list would be as such:

```elixir
defmodule MyApp.FooType do
  alias MyApp.BarResource
  use Ash.Type.NewType,
    subtype_of: :map,
    constraints: [
      fields: [
        bars: [type: :struct, constraints: [ instance_of: BarResource]]
      ]
    ]
end
```

**Expected behavior**
An error at compile time for unknown keys or incorrect keys.

**Runtime**
 - Elixir version 1.18.1
 - Erlang version 27.2 
 - OS macOS Sonoma 14.7
 - Ash version 3.4.71
 - any related extension versions
  - ash_json_api 1.4.22

**Additional context**

Link to discord thread: https://discord.com/channels/711271361523351632/1254774851428814930/1352834543408775228

LukeFrandsen:
I was able to reproduce the issue on the current main and would like to work on this if nobody else is doing so.

zachdaniel:
I think this was intended to have been fixed here: https://github.com/ash-project/ash/pull/1913

But maybe it was incomplete.

PR #1913, opened 2025-03-26
improvement: new type recursive validation

ken-kost:
#1899 
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/new_type.ex
+++ b/lib/ash/type/new_type.ex
@@ -91,11 +91,12 @@ defmodule Ash.Type.NewType do
   end
 
   defmacro __using__(opts) do
-    with {:ok, type, constraints} <- validate_opts(opts),
-         :ok <- validate_constraints(type, constraints) do
-      :ok
-    else
-      {:error, error} -> raise ArgumentError, error
+    case Keyword.keys(opts) -- [:subtype_of, :constraints, :lazy_init?] do
+      [] ->
+        []
+
+      keys ->
+        raise ArgumentError, "Unknown options given to `use Ash.Type.NewType`: #{inspect(keys)}"
     end
 
     subtype_of =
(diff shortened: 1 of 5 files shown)


lib/ash/type/new_type.ex
```diff
@@ -428,4 +427,52 @@ defmodule Ash.Type.NewType do
...
+    case Keyword.keys(opts) -- [:subtype_of, :constraints, :lazy_init?] do
+      [] ->
+        {:ok, opts[:subtype_of], opts[:constraints]}
+
+      keys ->
+        {:error, "Unknown options given to `use Ash.Type.NewType`: #{inspect(keys)}"}
+    end
+  end
+
+  defp validate_constraints({:__aliases__, _, _}, _constraints) do
```
ken-kost, line 441: Had to add this case because 
```
defmodule Type.LazyInitTest.ExampleNested do
  @moduledoc false
  use Ash.Type.NewType, subtype_of: Type.LazyInitTest.Example, lazy_init?: true
end
```
was failing for this new validation. 👀 Not sure what to do.

zachdaniel: This logic has to happen as part of the `init` logic, otherwise recursive new types will find compiler locks.

ken-kost: > This logic has to happen as part of the `init` logic, otherwise recursive new types will find compiler locks.

So in here
```
  def init(type, constraints) do
    type = get_type(type)
    type.init(constraints)
  end
```
in `type:ex:898`?

ken-kost: Hmm, so when I moved `validate_constraints` into `init` the invalid new types I defined in my test do not go through it. 🤔 I only see `type #=> Ash.Type.DateTime`

zachdaniel: I'm not sure what the issue is, but ultimately we just cannot do this logic in the actual type definition because it will cause deadlocks. Did you add it to both of the init functions? We define them differently depending on the `lazy_init?` option.

ken-kost: yes, what happened was the custom new types I defined in test didn't appear in init. I will try again.
Do you want me to push what I have?

ken-kost: pushed WIP commit. The test I added is now failing. Maybe validation is also needed in type?

zachdaniel: Okay, so, figured some stuff out. Here is how we make it right:

```elixir
  def init(type, constraints) do
    type = get_type(type)

    case validate_constraints(type, constraints) do
      {:ok, constraints} ->
        type.init(constraints)

      {:error, error} ->
        Exception.format(:error, error)
    end
  end
```

In `Ash.Type`, we validate types on init. We also remove it in `actions/action/action.ex`:

```elixir
  def transform(%{returns: original_type, constraints: constraints} = thing) do
    type = Ash.Type.get_type!(original_type)

    with {:ok, constraints} <- Ash.Type.init(type, constraints) do
      {:ok, %{thing | returns: type, constraints: constraints}}
    else
      {:error, error} ->
        {:error, error}
    end
  end
```

Then, in your tests, you need to call `Ash.Type.init` to observe errors.

zachdaniel: We may discover a fair few errors as a result of this, not sure 😆

zachdaniel:

PR #1916, opened 2025-03-27
Failing test: Update unloads relationships on no-op update with tenant

jechol:
Calling Ash.update/2 with an empty attribute list and a tenant option causes previously loaded relationships to be unloaded unexpectedly.

The issue appears to stem from the following line:
https://github.com/ash-project/ash/blob/87340736e9d64bdeb6ead817c6b09db97d4df2d8/lib/ash/actions/update/update.ex#L553

At this point, changeset.data seems to be discarded and re-fetched via a new query, resulting in all previously loaded data being unloaded.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/filter/multitenancy_test.exs
+++ b/test/filter/multitenancy_test.exs
@@ -19,7 +19,7 @@ defmodule Ash.Test.MultitenancyTest do
 
     actions do
       default_accept :*
-      defaults [:read]
+      defaults [:read, update: :*]
 
       create :create do
         argument :multitenant_related, {:array, :map}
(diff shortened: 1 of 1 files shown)


zachdaniel:
Yes, we should merge the attributes from the newly fetched record into `changeset.data`. would you want to PR that fix?

jechol:
@zachdaniel I'm only understanding it in a pretty surface-level way, so it'd be great if you could take care of the fix.

zachdaniel:

zachdaniel:
Fixed in `main`.

PR #1920, opened 2025-03-27
Set doc false for internal functions

shankardevy:
Fixes a minor documentation issue which exposes `calculate_opts` function here in https://hexdocs.pm/ash/Ash.html#functions


zachdaniel:

PR #1925, opened 2025-03-30
feat(enum): Ash.Type.Enum - Add optional description and label value …

### LINKED ISSUE #1724: Add a function to get Enum descriptions and values.
The specific need I asked about in my [Elixir Forum Post](https://elixirforum.com/t/ash-type-enum-to-phoenix-html-form-options-for-select-options/68869/5) was for a function that supported using Enum descriptions and values for select input options on Phoenix Forms.  

Creating a custom Enum invites the opportunity to provide labels or descriptions for the values.  There's presently no function that returns the labels and descriptions.  It seems a pretty natural fit to leverage the descriptions as labels in Phoenix forms.  

Add a function to Ash.Type.Enum that returns the labels and descriptions.  

1. I have implemented this myself on my custom types.
2. Add options_for_select/1 to AshPhoenix.  The reason I bring this up is that options_for_select has the labels first where Ash.Type.Enum defines them second.  It might be more natural for the function on Ash.Type.Enum to return them the same way they're defined.  

On Ash.Type.Enum

```elixir
  @doc """
  Returns a list of the values and their descriptions.
  """
  @spec labeled_values() :: [{:atom, String.t()}]
  def labeled_values() do
    values()
    |> Enum.map(fn value -> {description(value), value} end)
  end
```

Or in AshPhoenix...

```elixir
  @doc """
  Returns a list of the descriptions, value pairs to pass to Phoenix.HTML.Form.options_for_select/2.
  """
  @spec options_for_select(term) :: [{String.t(), :atom}]
  def options_for_select(enum) do
    enum.values()
    |> Enum.map(fn value -> {enum.description(value), value} end)
  end
```

chazwatkins:
- Add keyword list for enum values to add `description` and/or `label`
- Add `label/1` to return the label for a given value
- Add `details/1` to return the label and description map
- Add validation to raise when unsupported options are provided for a value
- Change `description_map` to `values_map`.  When `{:key, description}` is provided per existing functionality, the description is put into the `values_map` with the `description` key.
- Change `description/1` to get the value from the `details_map` using the `description` key
- Update documentation to show the new functions and ways to add `description` and `label`
- Add `value_to_label/1` function that generates a label from the `:atom` key or uses the `:string` key as the label when a label is not provided.

closes #1724

This addition is completely backwards compatible, and all existing tests pass. Ash users who already have enums with descriptions won't be affected.

Example

```elixir
defmodule MyApp.MyEnum do
  use Ash.Type.Enum, values: [
    open: "An open ticket",
    closed: [description: "A closed ticket", label: "Closed"],
    escalated: [label: "Uh oh!"]
  ]
end

```

The next step is to add a PR to `ash_phoenix` to leverage the `label` to provide nice select options in forms.

I attempted to switch away from the `validate_values` function and leverage constraints, but I didn't have much luck. I'm open to guidance on how to do it, as I think it would be nice to do it instead of the manual validation function.

Let me know if you have any questions or feedback.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/enum.ex
+++ b/lib/ash/type/enum.ex
@@ -119,14 +119,14 @@ defmodule Ash.Type.Enum do
   @callback match(term) :: {:ok, atom} | :error
 
   defmacro __using__(opts) do
-    quote location: :keep, generated: true, bind_quoted: [opts: opts, behaviour: __MODULE__] do
+    quote location: :keep, generated: true, bind_quoted: [opts: opts] do
       use Ash.Type
 
       require Ash.Expr
 
-      @behaviour behaviour
+      @behaviour Ash.Type.Enum
 
-      @values_map behaviour.build_values_map(opts[:values])
+      @values_map Ash.Type.Enum.build_values_map(opts[:values])
       @values Map.keys(@values_map)
 
       atom_typespec =
@@ -158,24 +158,24 @@ defmodule Ash.Type.Enum do
 
       @any_not_downcase? Enum.any?(@string_values, fn value -> String.downcase(value) != value end)
 
-      @impl behaviour
+      @impl Ash.Type.Enum
       def values, do: @values
 
-      @impl behaviour
+      @impl Ash.Type.Enum
       def label(value) when value in @values do
         value
         |> details()
         |> Map.get(:label)
       end
 
-      @impl behaviour
+      @impl Ash.Type.Enum
       def description(value) when value in @values do
         value
         |> details()
         |> Map.get(:description)
       end
 
-      @impl behaviour
+      @impl Ash.Type.Enum
       def details(value) when value in @values do
         Map.get(@values_map, value)
       end
@@ -277,7 +277,7 @@ defmodule Ash.Type.Enum do
         end
       end
 
-      @impl behaviour
+      @impl Ash.Type.Enum
       @spec match?(term) :: boolean
       def match?(term) do
         case match(term) do
@@ -286,7 +286,7 @@ defmodule Ash.Type.Enum do
         end
       end
 
-      @impl behaviour
+      @impl Ash.Type.Enum
       @spec match(term) :: {:ok, atom} | :error
       def match(value) when value in @values, do: {:ok, value}
       def match(value) when value in @string_values, do: {:ok, String.to_existing_atom(value)}
--- a/lib/ash/type/enum.ex
+++ b/lib/ash/type/enum.ex
@@ -62,19 +62,22 @@ defmodule Ash.Type.Enum do
   `"HaLf_emPty"` would not be accepted by the code provided earlier. If case normalization is
   needed for additional values, it must be explicitly implemented.
 
-  ## Value descriptions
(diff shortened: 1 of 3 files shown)


lib/ash/type/enum.ex
```diff
@@ -102,7 +126,8 @@ defmodule Ash.Type.Enum do
 
       @behaviour behaviour
 
-      @values behaviour.build_values(opts[:values])
+      @values_map behaviour.build_values_map(opts[:values])
```
zachdaniel, line 129: I know this was here from before you, but there is no need for `behaviour` to be a variable like this. We should remove that variable and just use `Ash.Type.Enum` everywhere. It's an unnecessary indirection. Would you mind making that change while you're extending this?

zachdaniel (dismissed):

lib/ash/type/enum.ex
chazwatkins: Not a problem.  I'll make the update

chazwatkins: Done.  Any thoughts on replacing the manual opts validation functions with constraints?  I got close to it working, but kept running into an error that `values` wasn't provided.  Is there anything special you have to do with constraints when putting them in a macro?

zachdaniel:
So `Ash.Type.Enum` is a bit of an old school type. It existed before we had `NewType` and might have been implemented very differently had that existed. It's not actually using constraints, it's just defining a self contained type module. So I don't think we can reasonably use constraints, at least w/o some other more major changes.

zachdaniel:
Marked as `improvement` because `feat` triggers a minor release version bump 😄🚀 Thank you for your contribution! 🚀

chazwatkins:
> Marked as `improvement` because `feat` triggers a minor release version bump 😄🚀 Thank you for your contribution! 🚀

I'll remember that for the future.  Thanks!

I'm happy to give converting the Enum to a modern implementation a go.  Are you thinking it should lever `NewType` instead?

zachdaniel:
Probably yes. What it would look like is we'd have a concrete type called `Ash.Type.Enum`, that on its own does nothing until its given constraints, like the options you provide it here, so it has to be fully usable on its own

i.e

```elixir
attribute :state, :enum, constraints: [....]
```

Once that is up to parity with the current enum type, and we add support for it in `AshGraphql`, `AshJsonApi`, `AshAdmin` and `AshStateMachine` 😅 Then, moving forward, we would advise people to not `use Ash.Type.Enum`, but instead `use Ash.Type.NewType, subtype_of: :enum, constraints: [...]`. It's not a small task, but it is doable. I'm not entirely sure that it is actually worth the effort to do though.

chazwatkins:
Fair enough.  I might look into it for some practice after the complex actions guide.

sevenseacat:
Did the follow up ever get made to AshPhoenix to leverage these? I can't find if it did or not

zachdaniel:
AFAIK no

PR #1935, opened 2025-04-01
chore: remove repeat of the same type in Aggregate opts schema

vonagam:
Just same thing repeated twice. Maybe there was something else in mind for the second item.


zachdaniel:
Ah, I'm pretty sure that was meant to be `Ash.Query.Aggregate` :)

vonagam:
Right. Updated. Also ran `mix docs` to fix not updated pubsub doc.

PR #1936, opened 2025-04-01
tests: test propagation of context to calculation in policy expression

vonagam:
Have an issue with usage of a calculation in a policy expression. The calculation is an expression one and references actor through `^actor(:id)`. The problem that I see is that actor/context does not propagate to calculation expression.

Two reasons (solving either of them should fix the issue, but I don't know what would be appropriate):
a) Calculation context is not filled.
b) `Ash.Expr.fill_template` does not go into `Ash.Query.Ref`.

To demonstrate using the test example, this works:
```elixir
policy action(:erase) do
  authorize_if expr(id == ^context(:post_id) and author_id == ^actor(:id))
end
```

But this does not:
```elixir
policy action(:erase) do
  authorize_if expr(has_context)
end

calculate :has_context, :boolean, expr(id == ^context(:post_id) and author_id == ^actor(:id))
```

### DIFF
--- a/test/policy/complex_test.exs
+++ b/test/policy/complex_test.exs
@@ -279,4 +279,8 @@ defmodule Ash.Test.Policy.ComplexTest do
 
     assert_received {:runtime_check_executed, [_]}
   end
+
+  test "calculations get context and actor", %{me: me, post_by_me: post} do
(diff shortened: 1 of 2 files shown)


zachdaniel:

Will investigate.

PR #1938, opened 2025-04-02
chore(docs): Update custom_input with and transform examples to retur…

chazwatkins:
- Update examples for code interface resource `custom_input`
  - `transform`'s `using` function returns from `:ok`/`:error` tuples to the value or nil only.  The tuples per the example isn't unwrapped, so you have to return the value only.
  - Replaced references to `with` to `using` to match the `custom_input` `transform` schema

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/dsls/DSL-Ash.Domain.md
+++ b/documentation/dsls/DSL-Ash.Domain.md
@@ -147,7 +147,7 @@ custom_input name, type
 ```
 
 
-Define or customize an input to the action. 
+Define or customize an input to the action.
 
 See the [code interface guide](/documentation/topics/resources/code-interfaces.md) for more.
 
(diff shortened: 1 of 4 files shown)


zachdaniel:
There is another bug here which is that it's actually callled `using`, I just didn't update the docs 🤦‍♂️

chazwatkins:
I believe I changed all the references of `with` to `using`.  Any other docs changes needed for this feature?

chazwatkins:
Noticed the `spark.cheat_sheets --check` check failed.  Ran `spark.cheat_sheets` and pushed an update.

zachdaniel (approved):

PR #1940, opened 2025-04-02
feat: tasks for generating custom modules

ken-kost:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/mix/tasks/gen/ash.gen.expression.ex
```diff
@@ -0,0 +1,99 @@
...
+      @impl true
+      def expression(_data_layer, args) do
+        {:ok, expr(args)}
+      end
+      """)
+    end
+
+    defp name(_expression, name) when is_binary(name), do: name
+
+    defp name(expression, _name) do
```
ken-kost, line 58: This seems hackish, I'm sorry. 😅 I'm hoping you know a better way. 🐛

lib/mix/tasks/gen/ash.gen.expression.ex
```diff
@@ -0,0 +1,99 @@
+if Code.ensure_loaded?(Igniter) do
+  defmodule Mix.Tasks.Ash.Gen.Expression do
```
zachdaniel, line 2: We should call this `ash.gen.custom_expression` since that is what they are called.

lib/mix/tasks/gen/ash.gen.expression.ex
zachdaniel: Seems like the right approach, but you can do the camelization to snake with `Macro.underscore`.

lib/mix/tasks/gen/ash.gen.expression.ex
ken-kost: So should everything have that prefix? `ash.gen.custom_change`

ken-kost: Oh, it's only for expression. Alright.

lib/mix/tasks/gen/ash.gen.expression.ex
ken-kost: That was exactly what I was looking for, big diff! 😂

zachdaniel:
Can we add a reference to the module that each generator generates in that guide? Like `Ash.Resource.Change` etc. doesn't make sense for base resource but the others it would.

PR #1942, opened 2025-04-02
fix: maintain order of enum values

hwuethrich:
When using a custom `Ash.Type.Enum` type, make sure `values/0` returns the values in the same order as defined on the type.

I'm showing the values in a select input and noticed the order changed randomly. This seems to be caused by converting the values to a map (for which the order of keys is not predictable).

I decided to keep the map (instead of replacing it with a keyword list) to keep label/description lookups performant.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #1948, opened 2025-04-03
set access_from in cascade_<update/destroy>

barnabasJ:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/change/cascade_destroy.ex
+++ b/lib/ash/resource/change/cascade_destroy.ex
@@ -280,7 +280,11 @@ defmodule Ash.Resource.Change.CascadeDestroy do
         return_notifications?: opts.return_notifications?
       )
 
-    context = Map.merge(relationship.context || %{}, %{cascade_destroy: true})
+    context =
+      Map.merge(relationship.context || %{}, %{
+        cascade_destroy: true,
+        accessing_from: %{source: relationship.source, name: relationship.name}
+      })
 
     context_opts =
       Keyword.update(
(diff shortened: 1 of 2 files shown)


lib/ash/resource/change/cascade_update.ex
```diff
@@ -210,7 +210,11 @@ defmodule Ash.Resource.Change.CascadeUpdate do
         return_notifications?: opts.return_notifications?
       )
 
-    context = Map.merge(relationship.context || %{}, %{cascade_destroy: true})
+    context =
+      Map.merge(relationship.context || %{}, %{
+        cascade_update: true,
```
barnabasJ, line 215: I `grep`ed through the project but couldn't find a place that looked for this and the tests are still working. I assume this was probably copy/pasted from the `cascade_destroy` module?

lib/ash/resource/change/cascade_destroy.ex
```diff
@@ -309,9 +313,11 @@ defmodule Ash.Resource.Change.CascadeDestroy do
           |> Ash.load!(
             [
               {relationship.name,
-               Ash.Query.set_context(relationship.destination, %{cascade_destroy: true})}
+               Ash.Query.set_context(relationship.destination, %{
+                 cascade_destroy: true,
+                 accessing_from: %{source: relationship.source, name: relationship.name}
+               })}
             ],
-            authorize?: false,
```
barnabasJ, line 314: only the loads used `authorize?: false` the bulk_<update/destroy>s don't. In my mind it's better to just set the accessing_from everywhere and let the policies do their thing?

lib/ash/resource/change/cascade_update.ex
zachdaniel: Yep, sounds correct to me 😅

PR #1951, opened 2025-04-05
fix: use the notification logic from bulk update in create

barnabasJ:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -188,8 +188,8 @@ defmodule Ash.Actions.Create.Bulk do
           after
             if opts[:notify?] && !opts[:return_notifications?] do
               Process.put(
-                {:bulk_update_notifications, ref},
-                Ash.Notifier.notify(Process.delete({:bulk_update_notifications, ref}) || [])
+                {:bulk_create_notifications, ref},
+                Ash.Notifier.notify(Process.delete({:bulk_create_notifications, ref}) || [])
               )
             end
           end
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -187,7 +187,10 @@ defmodule Ash.Actions.Create.Bulk do
             )
           after
             if opts[:notify?] && !opts[:return_notifications?] do
-              Ash.Notifier.notify(Process.delete({:bulk_create_notifications, ref}))
+              Process.put(
+                {:bulk_update_notifications, ref},
+                Ash.Notifier.notify(Process.delete({:bulk_update_notifications, ref}) || [])
+              )
             end
           end
         end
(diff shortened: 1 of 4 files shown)


lib/ash/actions/create/bulk.ex
```diff
@@ -221,11 +215,18 @@ defmodule Ash.Actions.Create.Bulk do
...
+                Ash.Notifier.notify(notifications)
+              end
+
+              if Process.get(:ash_started_transaction?) do
+                Process.put(
+                  :ash_notifications,
+                  Process.get(:ash_notifications, []) ++ notifications
+                )
+              end
+
```
barnabasJ, line 229: This part is copied from the bulk update code.

barnabasJ:
Looking at the code, it looks kinda expected that the notifications are returned as part of the stream when `return_stream?: true` is set. 
 
Let me know how this should be resolved.

lib/ash/actions/create/bulk.ex
```diff
@@ -221,11 +224,18 @@ defmodule Ash.Actions.Create.Bulk do
...
-
             true ->
+              if opts[:notify?] do
+                Ash.Notifier.notify(notifications)
+              end
+
+              if Process.get(:ash_started_transaction?) do
+                Process.put(
+                  :ash_notifications,
+                  Process.get(:ash_notifications, []) ++ notifications
```
zachdaniel, line 235: I think we might want something like:

```elixir
notifications =
              if opts[:notify?] do
                 Ash.Notifier.notify(notifications)
               else
                 notifications
               end
```

in the section above so that we don't send some notifications and then also return them.

lib/ash/actions/create/bulk.ex
```diff
@@ -187,7 +187,10 @@ defmodule Ash.Actions.Create.Bulk do
             )
           after
             if opts[:notify?] && !opts[:return_notifications?] do
-              Ash.Notifier.notify(Process.delete({:bulk_create_notifications, ref}))
+              Process.put(
+                {:bulk_update_notifications, ref},
+                Ash.Notifier.notify(Process.delete({:bulk_update_notifications, ref}) || [])
```
zachdaniel, line 192: `bulk_update_notifications` -> `bulk_create_notifications` ?

barnabasJ:
@zachdaniel 

> Looking at the code, it looks kinda expected that the notifications are returned as part of the stream when return_stream?: true is set. So I'm not sure if the test was supposed to work or not.
>
> Let me know how this should be resolved.

Could you check the test and let me know if the test or the implementation is wrong?

barnabasJ:
> Spoke offline about it, don't see a failing test so not currently sure what to look at 😆

Ok, looks like the test was failing because the copy/paste I didn't fix up properly. In that case this should be ready to review

zachdaniel (approved):

zachdaniel:

PR #1955, opened 2025-04-08
improvement: Add Query.Function.now/1-2 for timezones

chazwatkins:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

- Update `Ash.Query.Function.Now` with `now/1-2` to allow passing in the timezone and timezone database to match `DateTime.now/1-2`
- `mix format` updated `ash.ex`

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -97,6 +97,17 @@ defmodule Ash do
                           """,
                           default: false
                         ],
+                        return_records?: [
+                          type: :boolean,
+                          doc: """
+                          If `true`, the the query will not be executed and not records will be returned.
+
+                          This can be used with `return_query?: true` to get the query without the records.
+
+                          Useful for using the code interface to get the query for aggregate functions like `Ash.exists?`
+                          """,
+                          default: true
+                        ],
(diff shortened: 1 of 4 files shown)


lib/ash.ex
```diff
@@ -97,6 +97,17 @@ defmodule Ash do
                           """,
                           default: false
                         ],
+                        return_records?: [
```
zachdaniel, line 100: Was this left in by accident?

zachdaniel:
This one is a bit complex. We do some complex stuff to retain the behavior of "elixir timestamps are always in utc" for functions like `start_of_day`. So that you always get back the UTC timestamp corresponding to the start of the day in a given timezone, not a date time at a particular time zone. I think it makes sense to support this kind of thing, but I think that we should probably have a new function here, like `now_at_tz` or something like that.

chazwatkins:
That's fair.  The utc timezones can always be shifted outside of an expression or in a custom module of some type.  Thanks for taking the time to review.  🙇‍♂️

PR #1960, opened 2025-04-09
fix: use correct exceptions and add missing fields to them

artiom:
While testing https://github.com/ash-project/ash_postgres/pull/524 found an issue in how exceptions are used (there are two `InvalidQuery` aliased in [lib/ash/query/query.ex](https://github.com/ash-project/ash/blob/main/lib/ash/query/query.ex#L248-L261)) and several fields were missing.

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/error/query/invalid_query.ex
```diff
@@ -4,7 +4,7 @@ defmodule Ash.Error.Query.InvalidQuery do
   """
   use Ash.Error.Exception
 
-  use Splode.Error, fields: [:field, :message], class: :invalid
+  use Splode.Error, fields: [:fields, :message, :value], class: :invalid
```
zachdaniel, line 7: This change from `field` to `fields` would be a breaking change for things that depend on the `field` field being present I believe. We can add `fields` *and* `field` potentially though. Why change `field` -> `fields`?

artiom: The only usage I have found in the code base is using [`fields`](https://github.com/ash-project/ash/blob/main/lib/ash/query/query.ex#L3388). That's the reason for changing to what code is expecting. For backwards compatibility you are correct, if someone relied on that module in their codebase it will break. Added `field` in addition to the `fields`.

zachdaniel: Interesting. We should probably update `AshPhoenix` and `AshGraphql` to use `fields` for this error in that case. like if `field` is `nil`, then check `fields. Could you check those repos out and follow that up?

zachdaniel:

zachdaniel (approved):

zachdaniel:
Looks like a small credo fix, will merge once that is handled 🙇

artiom:
> Looks like a small credo fix, will merge once that is handled 🙇

Pushed a fix 👍

lib/ash/error/query/invalid_query.ex
artiom: Will check `AshPhoenix` and `AshGraphql` for that usage.

lib/ash/query/query.ex
```diff
@@ -258,7 +258,8 @@ defmodule Ash.Query do
     Required
   }
 
-  alias Ash.Error.Load.{InvalidQuery, NoSuchRelationship}
+  alias Ash.Error.Load.InvalidQuery, as: LoadInvalidQuery
```
vonagam, line 261: Personally, I would just remove this alias and use fully qualified name in the code (it is used only 2 times anyway) to avoid confusion (because there is `Ash.Error.Query.InvalidLoad` for example).

artiom: Sounds like a great idea 👍 Pushed an update.

lib/ash/error/query/invalid_query.ex
artiom: [AshPhoenix](https://github.com/ash-project/ash_phoenix/pull/349) and [AshGraphql](https://github.com/ash-project/ash_graphql/pull/301) PRs

zachdaniel:
@vonagam if you have a few to review/merge these PRs that would be great 🙇

vonagam:
This PR seems ok.

Question about Phoenix/GraphQL PRs to Zach: in both you said to use `Map.get` to support previous versions. But theoretically PR can jump bump minimum required version of Ash for the packages, right? But you don't want to do it for such small thing, correct? Just want to confirm.

zachdaniel:
Yes, exactly.

PR #1961, opened 2025-04-10
fix(CodeInterface): logic error in `get_by` code interfaces.

jimsynz:
This change causes a bunch of "no such input `id`" test failures.  I'm considering filtering those out of the params before passing them into the action, but this would be a change of behaviour so I'm not sure what the correct thing to do is here.


zachdaniel:
The trick is to never merge the `params` and the args that are from filters at all.

So this work needs to happen up here, before we can't tell which is which: https://github.com/ash-project/ash/pull/1961/files#diff-1d9e19259691a4fc7f075f57875b4ddd8098e154132a72325e2bad771986ae63L657

zachdaniel:

PR #1963, opened 2025-04-10
fix: properly handle keyword variant in `Ash.Query.add_error/3`

vonagam:
Was getting complains when tried to use `Ash.Query.add_error(query, message: "Bad query")`.

Two things:

1) There were two aliases for `InvalidQuery` in the file - one for `Ash.Error.Query.InvalidQuery` and `Ash.Error.Load.InvalidQuery` with later taking a precedence. (I would have assumed that there should be a Credo rule for that.)

2) `Ash.Error.Query.InvalidQuery` does not accept `fields` or `value` - it has only `field` and `message`. And since `field` is guaranteed to be nil (because of the condition it is in) there is no need to pass it in.

### DIFF
--- a/lib/ash/query/query.ex
+++ b/lib/ash/query/query.ex
@@ -244,6 +244,7 @@ defmodule Ash.Query do
   alias Ash.Actions.Sort
 
   alias Ash.Error.Invalid.TimeoutNotSupported
+  alias Ash.Error.Load.NoSuchRelationship
 
   alias Ash.Error.Query.{
     AggregatesNotSupported,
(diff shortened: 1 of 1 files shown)


zachdaniel:
@vonagam there is a similar PR open about this, could you review that and see if this is still necessary? I'll leave it to you to merge that other one or close it and adjust this one etc 😄 


https://github.com/ash-project/ash/pull/1960

vonagam:
What a coincidence. I will close this one in favor of the mentioned PR - it seems more proper.

PR #1968, opened 2025-04-12
docs: Generic action implementation

capoccias:
Had some difficulty determining the implementation of the module that a generic action's `run` function takes so updating the docs to hopefully help someone else out

Could add a generator, taking inspiration from https://github.com/ash-project/ash/pull/1940, but unsure how useful that would be

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/actions/action/action.ex
+++ b/lib/ash/resource/actions/action/action.ex
@@ -75,7 +75,10 @@ defmodule Ash.Resource.Actions.Action do
                        {:spark_function_behaviour, Ash.Resource.Actions.Implementation,
                         {Ash.Resource.Action.ImplementationFunction, 2}},
                        {:spark, Reactor}
-                     ]}
+                     ]},
+                  doc: """
+                  Module may be an `Ash.Resource.Actions.Implementation` or `Reactor`.
+                  """
                 ]
               ]
               |> Spark.Options.merge(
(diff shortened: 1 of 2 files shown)


lib/ash/resource/actions/action/action.ex
```diff
@@ -75,7 +75,10 @@ defmodule Ash.Resource.Actions.Action do
                        {:spark_function_behaviour, Ash.Resource.Actions.Implementation,
                         {Ash.Resource.Action.ImplementationFunction, 2}},
                        {:spark, Reactor}
-                     ]}
+                     ]},
+                  doc: """
+                  Module may be an `Ash.Resource.Actions.Implementation` or `Reactor`.
+                  """
```
capoccias, line 81: This seems like a double-up but unfortunately the generated [documentation ](https://hexdocs.pm/ash/dsl-ash-resource.html#actions-action-run) doesn't specify which modules it accepts

image

zachdaniel: Doesn't hurt to double up the description, but perhaps we should add a version of this function: https://github.com/ash-project/spark/blob/main/lib/spark/options/docs.ex#L350

That makes for slightly better DSL docs? So like instead of `module` it could say `behaviour(module)` or something?

zachdaniel (approved):

zachdaniel:

PR #1982, opened 2025-04-17
Set doc false to internal public functions

shankardevy:
I figured out `manage_relationship_opts` and `run_atomic_change` public not because you want them as public interface for other developers to use but because there are other Ash modules that need to call them for internal implementation details. So marking them with `@doc false` as they show up on hexdocs which is confusing.

However, there are two more functions:

1. `set_on_upsert` 
2. `handle_params`

which looks like an internal public function but I'm not sure of the intent of the authors. If the authors could clarify if they intend to make these two publicly available api, I can write the docs or otherwise, mark them with `@doc false`.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1015,6 +1015,7 @@ defmodule Ash.Changeset do
     end)
   end
 
+  @doc false
   def run_atomic_change(
         changeset,
         %{change: {module, change_opts}, where: where},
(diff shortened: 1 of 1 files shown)


zachdaniel:
Yes those other two functions should also be marked `@doc false`, thank you 🙇

shankardevy:
Thanks for the clarification @zachdaniel. I have modified the commit to include these two functions as well.

zachdaniel:

PR #1998, opened 2025-04-27
doc: add ? to allow_forbidden_field_for_relationships_by_default

djthread:
# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

I noticed that my starter app config.exs was already declaring this option but with a trailing `?`.

Side note: the note seems fair enough about it not becoming `true` by default until 4.0, even though my starter app was already overriding that default to true.


zachdaniel (approved):

zachdaniel:
Yeah so the default change refers to old apps that were created before that config was added to the generators. Currently the config defaulting to `false` when being accessed means that they retain their old behavior. In 4.0 the config will default to `true` (and likely just be removed actually) so users will have to adjust their code accordingly.

zachdaniel:

PR #1659, opened 2024-12-13
fix - handle soft destroys with manage_relationship

lardcanoe:
Technically, [line 1689](https://github.com/lardcanoe/ash/blob/soft-deletes-manage-relationship/lib/ash/actions/managed_relationships.ex#L1689) also has this issue, but not sure how to target it.

# Contributor checklist

- [ ] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

PR #1999, opened 2025-04-28
fix: Allow soft-deleting of many-to-many relationships

sevenseacat:
This continues on from #1659.

Extracts the logic for deleting the destination after deleting the join to a separate function, as deleting the join can return both `{:ok, notifications}` and `{:ok, record, notifications}` if the join record is soft-deleted.

Expands on the logic to also handle `{:ok, record, notifications}` if the destination record is also soft deleted.

Includes tests for both hard/soft deleting many-to-many relationships - the hard delete passed and the soft delete failed before making any code changes, and now they both pass.

Also fixes a small bug where `all_notifications` was getting duplicated inside the `{:ok, notifications}` case clause.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/managed_relationships.ex
+++ b/lib/ash/actions/managed_relationships.ex
@@ -1694,31 +1694,30 @@ defmodule Ash.Actions.ManagedRelationships do
                 |> Ash.destroy(return_notifications?: true)
                 |> case do
                   {:ok, join_notifications} ->
-                    notifications = join_notifications ++ all_notifications
-
-                    record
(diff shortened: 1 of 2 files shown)


zachdaniel:
fixed.

sevenseacat:
ah my version was too close to the app I spotted the problem from, which used `use_identities` to match by something other than ID. Still, I'm kind of surprised it ever worked!

PR #2003, opened 2025-04-29
fix: only return the changeset

barnabasJ:
Having the :cont there lead to it breaking [here](https://github.com/ash-project/ash/blob/fafa5b4f9f8378c021668d53cf4e3a98d40f0a16/lib/ash/actions/update/bulk.ex#L3013-L3014)


zachdaniel (approved):

zachdaniel:

Issue #1970, opened 2025-04-13
Support mod/2 (or rem/2) in Ash.Expr for modulo-based filtering


jechol:
**Is your feature request related to a problem? Please describe.**
When working with resources that require filtering based on modulo logic (e.g. "even numbers", "every 3rd item"), it's currently impossible to express this directly in Ash's expr. For example, filtering records where i % 2 == 0 is not currently supported, which is a common condition.

**Describe the solution you'd like**
Support mod(left, right) (or rem(left, right)) as a first-class operator in Ash.Expr. This would allow expressions like:
```
expr(mod(i, 2) == 0)
```
This would unlock useful use cases like even/odd filtering, modulo group assignment, pagination keys, and more.

zachdaniel:
Agreed it's a good idea. You can do it with fragments but it would be a good idea to have div and rem. PRs welcome!

paulsullivanjr:
I can take this issue.

zachdaniel:
Sorry I forgot to close this. This is done.

PR #2004, opened 2025-04-29
feat: mod expr

ken-kost:
#1970 
This is just an initial try to get the ball rolling. I imagine there's things to change and or add.
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/query/mod_expression.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Ash.Query.ModExpression do
```
zachdaniel, line 1: We don't need a new expression type, only the function.

zachdaniel: No need. Its a function so the function handling logic will handle it.

ken-kost:
What else? Is this really it? 👺 
Also, should it be called mod since it's using rem? Maybe I should rename it to rem to be more consistent? 🐛

lib/ash/filter/filter.ex
```diff
@@ -3737,6 +3739,10 @@ defmodule Ash.Filter do
     end
   end
 
+  def do_hydrate_refs(%{op: :mod, left: _left, right: _right} = expr, context) do
```
zachdaniel, line 3742: This isn't necessary.

zachdaniel:
Once you remove the hydration code then yep that's all you need. However, we then need to add this to `ash_sql` 😄

zachdaniel:
Good question about mod vs rem. Let's call it rem 😄

zachdaniel:

zachdaniel:
Don't forget to add to `ash_sql` too ❤️ I'm releasing `Ash` again today to fix a bug w/ the last release so `ash_sql` can update its `ash` dep.

Issue #2012, opened 2025-05-02
Proposal: Support `limit` on `has_many` relationships


pikdum:
**Is your feature request related to a problem? Please describe.**

I'd like to be able to have relationships like `the 10 most recent posts, sorted by inserted_at desc`.
I'd also like to use those relationships in aggregates like `count of "the 10 most recent posts, sorted by inserted_at desc" that are unpublished`.

**Describe the solution you'd like**

```elixir
has_many ... do
  limit 10
  sort inserted_at: :desc
end
```

**Describe alternatives you've considered**

* limits on aggregate counts - no support in the DSL
* limits on inline calculation aggregates - throws an error if you try: https://github.com/ash-project/ash/blob/f6688fcc71b750f8d96222ec8eb157b0eff523ea/lib/ash/query/aggregate.ex#L523

**Express the feature either with a change to resource syntax, or with a change to the resource interface**

```elixir
has_many ... do
  limit 10
  sort inserted_at: :desc
end
```

**Additional context**

https://discord.com/channels/711271361523351632/1367931778144272657

pikdum:
Closing, since this was added. :)

PR #2016, opened 2025-05-07
enhancement: support has many limit

ken-kost:
#2012 

Not sure how correct this is regarding implementation details,. 
Also I feel the test is misplaced, but it felt most convenient to make the test in that file.

# Contributor checklist 

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -172,7 +172,6 @@ defmodule Ash.Actions.Read do
       end
 
     initial_query = query
-
     query = add_field_level_auth(query, query.domain, opts)
 
     query = %{
@@ -195,6 +194,7 @@ defmodule Ash.Actions.Read do
         source_context: query.context
       )
 
+    query = add_relationship_limit_to_loads(query)
     relationship? = Map.has_key?(query.context, :accessing_from)
 
     page_opts =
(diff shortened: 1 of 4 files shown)


lib/ash/actions/read/read.ex
```diff
@@ -2623,6 +2623,24 @@ defmodule Ash.Actions.Read do
     }
   end
 
+  defp add_relationship_limit_to_loads(%Ash.Query{load: []} = query), do: query
```
zachdaniel, line 2626: We should do this in `relationships.ex` when we build the related query, not here 😄

zachdaniel:
We will also have to ensure that in `ash_sql` in `join.ex` when we build the related subquery that we add a `limit`, and test this against things like `filter(related.foo == bar)` that the `limit` is applied in that context (we may need to do some work to detect said limit there 😄)

ken-kost:
> We will also have to ensure that in `ash_sql` in `join.ex` when we build the related subquery that we add a `limit`, and test this against things like `filter(related.foo == bar)` that the `limit` is applied in that context (we may need to do some work to detect said limit there 😄)

Could you expand on this? What would be a failing test example? The test should be in ash I assume.
From line 370 there is:
```
    |> Ash.Query.unset([:sort, :distinct, :select, :limit, :offset])
    |> handle_attribute_multitenancy(tenant)
    |> hydrate_refs(context[:private][:actor])
    |> then(fn query ->
      if sort? do
        Ash.Query.sort(query, relationship.sort)
      else
        Ash.Query.unset(query, :sort)
      end
    end)
```
should limit check be made somewhere there?

lib/ash/actions/read/relationships.ex
```diff
@@ -310,10 +310,15 @@ defmodule Ash.Actions.Read.Relationships do
...
       |> then(fn query ->
-        if relationship.cardinality == :one && Map.get(relationship, :from_many?) do
-          Ash.Query.limit(query, 1)
-        else
-          query
+        cond do
+          relationship.cardinality == :one && Map.get(relationship, :from_many?) ->
+            Ash.Query.limit(query, 1)
+
+          relationship.cardinality == :many && Map.get(relationship, :limit) ->
```
zachdaniel, line 317:
```suggestion
          Map.get(relationship, :limit) ->
```

zachdaniel:

For the testing, what I was getting at is that we should now go and confirm via tests in `ash_postgres` that `ash_sql` does the right thing with these relationships, specifically when used in filters. i.e `Ash.Query.filter(query, top_2_posts.name == "foo")` (just an example)

ken-kost:
Alright, I'll try it out. 👍

PR #2020, opened 2025-05-09
fix: add_new_code_interface/5 when do block occurs after resource

mbuhot:
Fixes issue with add_new_code_interface/5 where the `define` would be put in a do block outside the resources block.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/resource/igniter_test.exs
+++ b/test/resource/igniter_test.exs
@@ -111,4 +111,96 @@ defmodule Ash.Resource.IgniterTest do
       |> assert_unchanged()
     end
   end
+
+  describe "add_new_code_interface" do
+    setup do
+      igniter =
+        test_project()
+        |> Igniter.compose_task("ash.gen.resource", [
+          "MyApp.Accounts.User",
+          "--domain",
+          "MyApp.Accounts",
+          "--default-actions",
+          "create,read,update,destroy",
+          "--uuid-primary-key",
+          "id",
+          "--attribute",
+          "name:string:required:public"
+        ])
+        |> apply_igniter!()
+
+      %{igniter: igniter}
+    end
+
+    test "add_new_code_interface adds a new code interface to existing resource", %{
+      igniter: igniter
+    } do
+      igniter
+      |> Ash.Domain.Igniter.add_new_code_interface(
+        MyApp.Accounts,
+        MyApp.Accounts.User,
+        :create,
+        "define :create, action: :create"
+      )
+      |> assert_has_patch("lib/my_app/accounts.ex", """
+      - |    resource(MyApp.Accounts.User)
+      + |    resource(MyApp.Accounts.User) do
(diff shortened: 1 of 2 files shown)


test/resource/igniter_test.exs
```diff
@@ -111,4 +111,99 @@ defmodule Ash.Resource.IgniterTest do
       |> assert_unchanged()
     end
   end
+
+  describe "add_new_code_interface" do
+    setup do
```
mbuhot, line 116: Not sure if there's a simpler way to setup a Domain, Resource and Config sufficiently to exercise add_new_code_interface ?

test/resource/igniter_test.exs
```diff
@@ -111,4 +111,99 @@ defmodule Ash.Resource.IgniterTest do
...
+            extend: []
+          ]
+        })
+        |> Mix.Tasks.Ash.Gen.Resource.igniter()
+        |> apply_igniter!()
+
+      %{igniter: igniter}
+    end
+
+    test "add_new_code_interface adds a new code interface to existing resource", %{
```
mbuhot, line 138: This is the test case that would fail previously, without the do-block on the resource the define is potentially put on the next do-block in the file.

test/resource/igniter_test.exs
```diff
@@ -111,4 +111,99 @@ defmodule Ash.Resource.IgniterTest do
...
+          options: [
+            domain: "MyApp.Accounts",
+            default_actions: ~w[create read update destroy],
+            uuid_primary_key: "id",
+            attribute: ["name:string:required:public"],
+            relationship: [],
+            extend: []
+          ]
+        })
+        |> Mix.Tasks.Ash.Gen.Resource.igniter()
```
zachdaniel, line 132: What we'd do here is:

```elixir
mix igniter.compose_task("ash.gen.resource", ["--cli", "flags", ...])
```

test/resource/igniter_test.exs
```diff
@@ -111,4 +111,99 @@ defmodule Ash.Resource.IgniterTest do
...
+      )
+      |> assert_has_patch("lib/my_app/accounts.ex", """
+        |    resource(MyApp.Accounts.User) do
+        |      define(:create, action: :create)
+      + |      define(:update, action: :update)
+        |    end
+      """)
+    end
+
+    test "add_new_code_interface appends doesnt add code interface when one already exists",
```
mbuhot, line 185:
```suggestion
    test "add_new_code_interface doesnt add code interface when one already exists",
```

test/resource/igniter_test.exs
zachdaniel: Sorry, I said that weird. Just `Igniter.compose_task`, not `mix igniter.compose_task(...)`

Issue #2014, opened 2025-05-04
Add `debug?: true` opt to `Ash.Changeset.manage_relationship`


codonnell:
**Is your feature request related to a problem? Please describe.**
It is difficult to debug erroneous uses of `Ash.Changeset.manage_relationship`. There is quite a bit of business logic packed into it, and as a user all I see are my inputs and the result. There are a couple of ways to get some insight, like enabling logs in the data layer to see which queries are being run and adding debug logs to the actions being called. However, these approaches require the user to piece together what `manage_relationship` is doing from its side effects and can create excessively noisy output.

**Describe the solution you'd like**
I would love to add a `debug?: true` option in `Ash.Changeset.manage_relationship` which logs the decisions made. For example, a call with `type: :append` might log something along the lines of:
```
MATCH: no match found for `%{id: 1}`. Attempting to look up using action `:read`.
LOOKUP: found record with `%{id: 1}`. Attempting to relate it using action `:update`.
UPDATE: Record with `%{id: 1}` successfully related.
```
☝ This was written up quickly by someone relatively unfamiliar with how `manage_relationship` works. It shouldn't be interpreted as how the logs should actually be formatted.

**Describe alternatives you've considered**
There are similar debugging tools built for policies that are controlled by configuration. I don't think that would be ideal here, since I believe in most cases debugging `manage_relationship` is isolated to one call site, where we could add the `:debug?` option, minimizing noise.

**Express the feature either with a change to resource syntax, or with a change to the resource interface**
```elixir
change manage_relationship(:tags, type: :append)
```
would become
```elixir
change manage_relationship(:tags, type: :append, debug?: true)
```
when the developer needs to debug a problem.

**Additional context**
This was motivated by a [thread](https://elixir-lang.slack.com/archives/C0193AHFLPN/p1746375803627759) in the elixir slack on May 4, 2025. Anyone who's curious can reference it until it's no longer available.

PR #2021, opened 2025-05-09
enhancement: Support manage relationship debug

ken-kost:
#2014 
Not sure how good this is. 👺 

Maybe the logs should be more in line with other ones:

```
12:09:16.742 [debug] UPDATE: 09476719-41d9-42c4-86cd-0f0dda357a61. <- NEW

12:09:16.749 [debug] "Updating Ash.Test.ManageRelationshipTest.ParentResource "09476719-41d9-42c4-86cd-0f0dda357a61":

Setting %{name: "Even newer name"} <- OLD


12:09:16.749 [debug] UPDATE: 09476719-41d9-42c4-86cd-0f0dda357a61. <- NEW
```

In `manage_relationship_test.exs` add
```
  setup do
    Logger.configure(level: :debug)
    :ok
  end
```
and add `debug?: true` to manage relationship calls. Should some kind of test be made for this? 🤔 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
This is a great start! Lets do a few things:

1. create a `debug_log` function that looks like this:
```elixir
defp debug_log(relationship, message, opts) do
  if opts[:debug] do
    Logger.debug(...)
  end
end
```

and then ensure that it has a format that explains what its doing, like `"Managed Relationship Debug, #{inspect(resource)}.#{changeset.action.name}: <the message>"`

ken-kost:
> Managed Relationship Debug, #{inspect(resource)}.#{changeset.action.name}:

How should I get the changeset? I mean a lot of them are piped directly to the action all i.e. are not available currently. 😬

ken-kost:
For example, currently I have:
```
    relationship.destination
    |> Ash.Changeset.new()
    |> Ash.Changeset.set_context(%{
      accessing_from: %{source: relationship.source, name: relationship.name}
    })
    |> Ash.Changeset.for_create(action_name, input,
      require?: false,
      actor: actor,
      tenant: changeset.tenant,
      authorize?: opts[:authorize?],
      skip_unknown_inputs: Map.keys(input),
      domain: domain(changeset, relationship)
    )
    |> Ash.Changeset.set_context(relationship.context)
    |> Ash.create(return_notifications?: true)
    |> case do
      {:ok, created, notifications} ->
        if opts[:debug?], do: Logger.debug("CREATE: #{created.id}.")
```
I should `changeset =` then in case `Ash.create(changeset, return_notifications?: true) `
but then also what should be for the error case? Should I assume it's a changeset? 🤔 
Since you mentioned relationship in `debug_log(relationship, message, opts)`, maybe I should just fetch some info from that; but then another problem there is how do I know what resource from the relationship was involved. 😓

ken-kost:
What if the logging is moved deeper, in actions run? `Ash.create(return_notifications?: true, debug?: opts[:debug?])` or something like that. Sounds crazy. And I guess then you would also need a debug label for the `Managed Relationship Debug` part. 🤔 
idk, throwing an idea out there. 🐛

zachdaniel:
I was referring to the parent changeset, not the nested one :)

ken-kost:
> I was referring to the parent changeset, not the nested one :)

oh, now I see it. 😅

ken-kost:
So should it be `debug_log(resource, changeset, response, message, debug?)` and call it for example like:
```
debug_log(relationship.through, changeset, :ok, "Created #{created.id}, opts[:debug?])
```
and for error case it would be
```
debug_log(relationship.through, changeset, :error, "Failed to #{created.id}, opts[:debug?])
```
wdyt?

zachdaniel:
👍 Sounds good.

ken-kost:
Log example:
```
13:05:48.988 [debug] Managed Relationship Debug, Ash.Test.ManageRelationshipTest.ParentResource.update_with_existing_parent: Update success "b9afdf34-865e-42ec-a749-fd5cc1e9caba".
```
There were some problems with my idea, this is what I ended up with.

zachdaniel:
Can you merge back in `main` and check on the build issues?

zachdaniel:
For the error cases, let's include the error something like 
```elixir
error = Ash.Error.to_ash_error(error)
Exception.format(:error, error, error.stacktrace && error.stacktrace.stacktrace)
```

ken-kost:
The error passing broke inline function call. 👺 
Regarding my "crazy" idea, I noticed there are some logs already - not sure where they come from. But my idea is essentially 'this debug? on manage relationship is cool - why can't anything and everything have it's own debug option?'
since ash has declarative functionality it would be cool to be able to unveil the declarations in a way, at least regarding basic CRUD aspect.
But I'm not sure if this is already supported in a way. 🤔 For instance, if I have a complex create action that creates a lot of related resources, how would I get logs of what was created when? 🐛 
Would it make sense to have 'debug this create action'?

zachdaniel:
We could add something like `debug?: true` to all Ash actions, yes. Not sure what it would look like. We do have a `change debug_log` that can be placed in create/update/destroy actions.

lib/ash/actions/managed_relationships.ex
```diff
@@ -2015,5 +2289,47 @@ defmodule Ash.Actions.ManagedRelationships do
...
+      message =
+        case response do
+          :ok ->
+            "#{action} success #{inspect(identifier)}"
+
+          {:error, error} ->
+            ash_error = Ash.Error.to_ash_error(error)
+            stacktrace = ash_error.stacktrace && ash_error.stacktrace.stacktrace
+            error = Exception.format(:error, ash_error, stacktrace)
+            "Failed to #{action} #{inspect(identifier)}, #{error}"
```
zachdaniel, line 2325: The error should go on its own line I'd say.

ken-kost:
> We could add something like `debug?: true` to all Ash actions, yes. Not sure what it would look like. We do have a `change debug_log` that can be placed in create/update/destroy actions.

Yea, I'll think about it more, also I'll try `debug_log`. 👍

ken-kost:
@zachdaniel Otherwise, I think this is done, right? 👀

zachdaniel (changes requested):

zachdaniel:
I forgot to submit my previous review, sorry 😢

lib/ash/actions/managed_relationships.ex
```diff
@@ -2015,5 +2289,52 @@ defmodule Ash.Actions.ManagedRelationships do
...
+          record,
+          opts[:debug?]
+        )
+
+        {:error, error}
+    end
+  end
+
+  defp debug_log(resource, changeset, action, response, record, debug?) do
+    if debug? do
```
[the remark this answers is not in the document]
ken-kost: you mean the whole record or you mean something else?

lib/ash/actions/managed_relationships.ex
```diff
@@ -2015,5 +2289,52 @@ defmodule Ash.Actions.ManagedRelationships do
...
+            Failed to #{action} #{inspect(identifier)}
+
+            #{error}
+            """
+
+          _ ->
+            "#{action} #{inspect(identifier)}"
+        end
+
+      log = "#{inspect(resource)}.#{changeset.action.name}: #{message}"
```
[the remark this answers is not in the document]
zachdaniel: It should be something like `Managing relationship #{inspect(changeset.resource)}.#{relationship.name} with action #{changeset.action.name}`: #{message}`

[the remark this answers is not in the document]
ken-kost: For `RelatedResource` with   
```
change manage_relationship(:parent_resource, on_lookup: :relate_and_update, debug?: true)
```
now I get:
```
18:24:17.002 [debug] Managed Relationship Debug, Ash.Test.ManageRelationshipTest.RelatedResource's relation to Ash.Test.ManageRelationshipTest.ParentResource with action create_with_parent: Create success.
```

lib/ash/actions/managed_relationships.ex
```diff
@@ -2015,5 +2263,52 @@ defmodule Ash.Actions.ManagedRelationships do
...
+          :destroy,
+          {:error, error},
+          opts[:debug?]
+        )
+
+        {:error, error}
+    end
+  end
+
+  defp debug_log(relationship, changeset, action, response, debug?) do
```
zachdaniel, line 2284: This `relationship` argument is actually `destination` now, but I think we should just pass down the relationship *name* being edited, since there could be multiple relationships to the same destination being edited at one time. I think once we rename the argument it should be good to go 😄

ken-kost: Like this:
```
13:09:38.944 [debug] Managed Relationship Debug, Ash.Test.ManageRelationshipTest.RelatedResource:parent_resource with action create_with_parent: Create success.
```
?

ken-kost: Above is currently committed. 
But I also tried this (not committed):
```
13:34:56.493 [debug] [manage_relationship: :parent_resource] [action: :create_with_existing_parent] [resource: Ash.Test.ManageRelationshipTest.RelatedResource] [response: :ok]
```
in the format of `[debug]` wdyt? error message is in it's own line and it's `[response: :error]`.
This might be later colorized 🤠  but in general I think it would be cool to be able debug any change for any action for any resource and a consistent format around it would be also cool; like a more technical and direct log right... In the end it would be great to turn on a lot of `debug?` flags and still be able to figure out what's what easily from the logs.

zachdaniel: I think for now I prefer the current format. If we were to do that, it would be `[resource: ..., response: ...]` in a single keyword list probably. I do see your point about having some kind of standardized format etc, but I think that can be its own conversation.

zachdaniel:

lib/ash/actions/managed_relationships.ex
ken-kost: So then is this it? 😆

PR #2027, opened 2025-05-12
Fix batch order of bulk_create

### LINKED ISSUE #2026: `bulk_create` with `integer_primary_key` assigns IDs in reverse order of input due to `split_and_run_simple` batch prepending
**Describe the bug**

When using `bulk_create` for a resource that has an auto-generating `integer_primary_key`, the primary keys assigned to the created records are in the reverse order of the input parameters.

This appears to be caused by the `Ash.Actions.Helpers.split_and_run_simple/7` function. Specifically, in this section:

```elixir
{batch, must_be_simple} =
  Enum.reduce(batch, {[], []}, fn changeset, {batch, must_be_simple} ->
    if changeset.around_transaction in [[], nil] and changeset.after_transaction in [[], nil] and
         changeset.around_action in [[], nil] do
      changeset = Ash.Changeset.run_before_transaction_hooks(changeset)
      {[changeset | batch], must_be_simple} # <--- Batch is prepended here
    else
      {batch, [%{changeset | __validated_for_action__: action.name} | must_be_simple]}
    end
  end)
```

https://github.com/ash-project/ash/blob/d6a83a687ca08801cdeb8ef4c2ec88c0ecf3eb59/lib/ash/actions/helpers.ex#L15-L25

The batch accumulator is built by prepending changesets (`[changeset | batch]`). This list is then returned and eventually processed by the data layer without being reversed. Consequently, the data layer inserts the records in the reverse order of their original appearance in the input to `bulk_create`, leading to the reversed ID assignment for auto-incrementing integer primary keys.

**To Reproduce**

```elixir
defmodule MyApp.MyResource do
  use Ash.Resource,
    domain: MyApp.Domain,
    data_layer: AshPostgres.DataLayer # Or any other data layer supporting integer PKs

  postgres do
    table "my_resources"
    repo MyApp.Repo
  end

  attributes do
    integer_primary_key :id
  end

  actions do
    defaults [:read, create: :*]
  end
end

defmodule MyApp.Domain do
  use Ash.Domain

  resources do
    resource MyApp.MyResource
  end
end

%Ash.BulkResult{records: records} = 
  [
    %{name: "A"}
    %{name: "B"}
    %{name: "C"}
  ]
  |> Ash.bulk_create(input_params, MyApp.MyResource, :create, [return_records?: true, sorted?: true]) do
```

**Expected behavior**

The auto-generated integer_primary_key values should correspond to the order of the input parameters. 

**Runtime**
 - Elixir version: 1.18
 - Erlang version: 27
 - OS: mac
 - Ash version: 3.5.9

**Additional context**

nallwhy:
This PR resolves #2026.

Regarding testing: It seemed difficult to write a specific test for this bug using `Ash.DataLayer.Ets`, so I haven't added one.
I also noticed that `notifications` are prepended to a list in related code. Should their order also be corrected?

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/helpers.ex
+++ b/lib/ash/actions/helpers.ex
@@ -24,6 +24,9 @@ defmodule Ash.Actions.Helpers do
         end
       end)
 
+    batch = batch |> Enum.reverse()
+    must_be_simple = must_be_simple |> Enum.reverse()
+
     context =
       struct(
         Ash.Resource.Change.Context,


zachdaniel:
This type of thing can be pretty expensive, and I do believe that there are other cases where the list can end up out of order. There is an option that I think should help here called `sorted?: true`. Can you try that?

nallwhy:
@zachdaniel 

Are you asking if applying `sorted?: true` would actually fix this issue, or are you suggesting that the changes in this PR should only take effect when `sorted?: true` is set?

If it’s the former, `sorted?: true` only affects the result order, so it wouldn’t address this particular issue. (and we’re already using `sorted?`)

If it’s the latter, it’s possible to scope this change to apply only when `sorted?: true` is enabled, but that would likely require updating the documentation to clarify the expected behavior of this option.

As for the performance concern, my understanding is that the cost of reverse is generally negligible. According to the article below, reversing a list of 100,000 elements takes around 10ms, and for 10,000 elements, it’s about 0.1ms. Given the scale of typical batch operations, the overhead is unlikely to be significant compared to other processing steps.

[Lists:reverse/1 Performance in Erlang (emauton.org)](https://emauton.org/2015/01/25/lists-reverse-1-performance-in-erlang/)

Additionally, although we don’t use auto-incremented `id`, we do maintain a `seq` field to track the intended order of operations. However, due to this issue, the order of `seq` is currently unreliable.

zachdaniel:
Fair enough.

zachdaniel:
I think we may need to do a similar thing w/ `bulk_update` and `bulk_destroy`. Would you mind checking that if you get a few?

zachdaniel:
Looks like this caused an issue in `ash_oban`: https://github.com/ash-project/ash/actions/runs/14988525752/job/42106924217

nallwhy:
> I think we may need to do a similar thing w/ `bulk_update` and `bulk_destroy`. Would you mind checking that if you get a few?

`split_and_run_simple/7` is used in both `bulk_update` and `bulk_destroy` as well, so this change affects those as well.

https://github.com/nallwhy/ash/blob/2a241f6e5307736076fc92a611090e00d5f7aa99/lib/ash/actions/destroy/bulk.ex#L1431-L1438
https://github.com/nallwhy/ash/blob/2a241f6e5307736076fc92a611090e00d5f7aa99/lib/ash/actions/update/bulk.ex#L1710-1717

zachdaniel:
Nice 👍

zachdaniel:
@nallwhy I think the issue is with the ETS data layer, will fix.

nallwhy:
> Looks like this caused an issue in `ash_oban`: https://github.com/ash-project/ash/actions/runs/14988525752/job/42106924217

Let me check this.

zachdaniel:
Its fixed 👍 not a bug w/ your contribution, a bug w/ the ETS data layer.

PR #2036, opened 2025-05-13
feat: duration

matt-beanland:
Adds Ash.Type.Duration which is an Elixir Duration, shortname: :duration (also :duration in Ecto). The precision inherent in the interval is maintained. 

The idea is here is to allow Durations to be ash attributes persisted in the chosen DataLayer, noting that Ash has some existing support for durations by combining attributes of :duration_name and :integer type. 

Given that Elixir DateTime.shift, etc can be used with :duration and Comp.compare is implemented (using to_timeout such that durations are compared based on their total magnitude) I'm not sure whether Ash.Query.Function such as Ago, Date_Add, From_Now are needed.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ✔️] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/query/operator/basic.ex
+++ b/lib/ash/query/operator/basic.ex
@@ -13,10 +13,11 @@ defmodule Ash.Query.Operator.Basic do
       no_nils: true,
       evaluate_types: :numbers,
       types: [
+        [:same, :any],
         [:duration, :integer],
         [:integer, :duration]
       ],
-      returns: [:duration, :duration]
+      returns: [:same, :duration, :duration]
     ],
     minus: [
       symbol: :-,
@@ -71,15 +72,6 @@ defmodule Ash.Query.Operator.Basic do
         @moduledoc false
 
         require Decimal
-        require Duration
-        require Date
-        require DateTime
-        require NaiveDateTime
-        require Time
-        require Date
-        require DateTime
-        require NaiveDateTime
-        require Time
 
         use Ash.Query.Operator,
           operator: unquote(opts[:symbol]),
--- a/lib/ash/query/operator/basic.ex
+++ b/lib/ash/query/operator/basic.ex
@@ -11,7 +11,12 @@ defmodule Ash.Query.Operator.Basic do
     times: [
       symbol: :*,
       no_nils: true,
-      evaluate_types: :numbers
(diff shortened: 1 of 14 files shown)


zachdaniel:
This looks great 😄 We can support this type in core, but likely there would need to be work done to support them in the expression language. For example, to support `date + duration` etc. I think I'd want some basic test of use in a slew of expressions before merging.

As for things like `ago`, `date_add` and `from_now`, I think they will always be useful, because they don't require constructing a duration. i.e `ago(some_attribute, :day)`.

zachdaniel:
Could you confirm what the behavior in postgresql is for comparing durations? I assume that if they do support comparators that it is effectively just comparing the length of the duration?

zachdaniel:
Mind pulling in main and resolving conflicts?

zachdaniel:
That is fine, do you mind deleting it in your branch in that case? The diff now is too large to review 😄

zachdaniel:
you should be able to force push to this branch or something? Ultimately you'll want to rebase main or get this into a clean state some other way. No problem if you want to close this and open another PR or something along those lines.

lib/ash/query/operator/basic.ex
```diff
@@ -11,7 +11,12 @@ defmodule Ash.Query.Operator.Basic do
     times: [
       symbol: :*,
       no_nils: true,
-      evaluate_types: :numbers
+      evaluate_types: :numbers,
+      types: [
+        [:duration, :integer],
```
zachdaniel, line 16: We need to add a `[:same, :any]` with a `:same` return here as well.

lib/ash/query/operator/basic.ex
```diff
@@ -66,6 +71,15 @@ defmodule Ash.Query.Operator.Basic do
         @moduledoc false
 
         require Decimal
+        require Duration
```
zachdaniel, line 74: are we using all of these required modules? Doesn't look like it to me.

zachdaniel (approved):

zachdaniel (approved):

zachdaniel:

Issue #1983, opened 2025-04-08
Proposal: Add Scope module to put Ash opts in assigns like Phoenix 1.8


chazwatkins:
**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you'd like**

Phoenix 1.8 has introduced the concept of [scopes](https://hexdocs.pm/phoenix/1.8.0-rc.0/scopes.html), which Ash already handles with `actor`, `tenant`, `tracer`, and `context` when calling any action. But we're currently missing a solution for providing these options to our UI views so they can be easily used as opts in Ash actions.

Define a `Scope` module that takes a `Phoenix.LiveView.Socket` or `Plug.Conn` and sets any of the following in the assigns:
- tenant
- actor
- tracer
- context

The module would have convenience functions for setting all or each item like
- `Scope.set_actor(conn, actor)`
- `Scope.set_tenant(conn, tenant)`
- `Scope.set_scope(conn, actor: actor, tenant: tenant)`

We can provide a plug and hook to set the scope that can be easily used in the router, including a way to set the scope for LiveViews from a `Plug.Conn` session.

Then you can use the `assigns.scope` for any Ash actions like
`Ash.read!(Foo, MyApp.Scope.to_opts(socket.assigns.scope)`

The module could be `AshPhoenix.Scope` or maybe we provide a way to give the module name that will be used.

**Describe alternatives you've considered**

I made my own implementation of this in a project with inspiration from `Ash.PlugHelpers` and it seems to work pretty well.  Prior to that, I was adding each item to the assigns as `current_actor` or `current_tenant` in a plug and hook.

**Additional context**

I will submit a PR to show the idea in more detail to get feedback on the proposed solution.  Happy to integrate any feedback to make the solution nice.

PR #2050, opened 2025-05-19
improvement: add `Ash.Scope` for converting a scope to options

zachdaniel:
This would likely need to be incrementally adopted in various places, and the first place would be in `AshAuthentication`. We would likely make it an option in the `ash_authentication_live_session`, to take a `scope`, and expect that scope to define a function like `from_conn/1`. (or something like that)

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -63,6 +63,11 @@ defmodule Ash do
       type: :any,
       doc:
         "If an actor is provided, it will be used in conjunction with the authorizers of a resource to authorize access"
+    ],
+    scope: [
+      type: :any,
+      doc:
+        "A value that implements the `Ash.Scope` protocol. Will overwrite any actor, tenant or context provided. See `Ash.Context` for more."
     ]
   ]
 
@@ -726,6 +731,11 @@ defmodule Ash do
       The actor for handling `^actor/1` templates, supplied to calculation context.
       """
     ],
+    scope: [
+      type: :any,
+      doc:
+        "A value that implements the `Ash.Scope` protocol. Will overwrite any actor, tenant or context provided. See `Ash.Context` for more."
+    ],
     tenant: [
       type: {:protocol, Ash.ToTenant},
       doc: """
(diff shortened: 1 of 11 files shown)


zachdaniel:
Take a look at the moduledocs for `Ash.Scope` for docs on intended use cases.

zachdaniel:
Addresses https://github.com/ash-project/ash/issues/1983

lib/ash.ex
```diff
@@ -63,6 +63,11 @@ defmodule Ash do
       type: :any,
       doc:
         "If an actor is provided, it will be used in conjunction with the authorizers of a resource to authorize access"
+    ],
+    scope: [
+      type: :any,
+      doc:
+        "A value that implements the `Ash.Scope` protocol. Will overwrite any actor, tenant or context provided. See `Ash.Context` for more."
```
RoyalIcing, line 70: What do you think about making it raise if you pass both `scope` and any of `actor`/`tenant`? I think leaving it ambiguous could be confusing — I’d actually assume that passing both `scope` and `actor` would mean the `actor` would win.

Making it strict by raising would mean people wouldn’t get into edge cases. Raising wouldn’t be a breaking change as people have to opt-into passing `scope`.

chazwatkins:
The implementation looks straightforward.  Using a protocal makes sense with the defaults for actor, tenant, context with the option to add more if you want.

Once this is merged, we should be able to tweak the Plug and LiveView Hook I submitted to `ash_phoenix` or `ash_authentication_phoenix` to add the scope to the assigns.

zachdaniel:
The main complexity comes from the fact that an application could provide multiple scope modules, and many different ways of making them. So we will need to let users configure a function that takes the data produced from these plugs and creates or updates a scope.

lib/ash.ex
zachdaniel: The difficulty lies in where we do this specific processing, it happens a deeper down when all the options typically have values. Lifting it up to the edges could be good, but also requires a much more spread around change 

Often times automated things will call actions with options extracted from various places and so they will just by happenstance have `actor: nil` etc. I think what we can do that will be "close enough" is to add a check where if `actor` has two *non-nil* values that are *different* then we can raise. i.e `actor: nil, scope: scope` will get `scope.actor`, but `actor: user1, actor: user2` will be an error.

PR #2061, opened 2025-05-23
fix: pass tenant in bulk_update to managed_relationships

xanderio:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/managed_relationships.ex
+++ b/lib/ash/actions/managed_relationships.ex
@@ -42,8 +42,9 @@ defmodule Ash.Actions.ManagedRelationships do
               Enum.any?(relationships, fn {_, opts} -> opts[:authorize?] end)
 
           actor = engine_opts[:actor]
+          tenant = engine_opts[:tenant]
 
-          case Ash.load(acc, key, authorize?: authorize?, actor: actor, domain: domain) do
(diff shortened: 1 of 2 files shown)


xanderio:
After looking at the current test, I'm quite confused why this wasn't caught within them.

lib/ash/actions/managed_relationships.ex
```diff
@@ -42,8 +42,9 @@ defmodule Ash.Actions.ManagedRelationships do
               Enum.any?(relationships, fn {_, opts} -> opts[:authorize?] end)
 
           actor = engine_opts[:actor]
+          tenant = engine_opts[:tenant]
 
-          case Ash.load(acc, key, authorize?: authorize?, actor: actor, domain: domain) do
+          case Ash.load(acc, key, authorize?: authorize?, actor: actor, tenant: tenant, domain: domain) do
```
zachdaniel, line 47: Typically the tenant can be inferred from the data being passed in is why it probably wasn't an issue. In this case maybe we weren't properly attaching the tenant when we should somewhere? It should be in `record.__metadata__.tenant`

zachdaniel: I don't see a reason not to explicitly pass it through, but there may be a separate bug here somewhere.

PR #2064, opened 2025-05-26
fix: make Ash.Query.sort/3 accept string param.

Torkan:
Also adds some regression tests, and a heads-up in the docs that sort_input/3 should be used for user input.

Regarding the last test in `sort_test.exs`: I don't know of a better way to assert on the resulting calculation expressions, if there is a better way, just point me in the right direction and I'll update it 😅 

# Contributor checklist

- [ ] Chores
- [x ] Documentation changes
- [x ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/query/query.ex
+++ b/lib/ash/query/query.ex
@@ -3173,6 +3173,12 @@ defmodule Ash.Query do
       ["foo", "-bar", "++baz", "--buz"]
 
 
+  > ### Sorting on user input? {: .neutral}
+  >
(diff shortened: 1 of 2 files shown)


zachdaniel:
Thanks for the PR! I've decided to do this a bit differently and improve the general handling/validation of sorts in `Ash.Query.sort`. See 1f2113368a33c1411ee6d57f3e1acc4deef97a0b

Issue #1500, opened 2024-10-05
Add compiler error for `integer_primary_key` with Ets 


brianhv:
Following up on [this message in discord](https://discord.com/channels/711271361523351632/711271361523351636/1291944980167790673).

When using the ETS data layer and an `integer_primary_key`, rows get inserted with `id: nil`, and rows after the first will clobber the previous rows. In Discord it was suggested that this situation ought to be a compile time error.

Example of code with which I observed the issue:

```elixir
defmodule Campdbash.Camper do
  use Ash.Resource,
    domain: Campdbash,
    data_layer: Ash.DataLayer.Ets

  actions do
    defaults [:read]

    create :register do
      accept [:first_name, :last_name]
    end
  end

  attributes do
    uuid_primary_key :id

    attribute :first_name, :string do
      allow_nil? false
    end

    attribute :last_name, :string do
      allow_nil? false
    end

  end
end
```

Inserted with

```elixir
Campdbash.Camper |> Ash.Changeset.for_create(:register, %{first_name: "fname", last_name: "lname"}) |> Ash.create!()
```

Result

```
#Campdbash.Camper<
  __meta__: #Ecto.Schema.Metadata<:loaded>,
  id: nil,
  first_name: "fname",
  last_name: "lname",
  aggregates: %{},
  calculations: %{},
  ...
>
```

PR #2072, opened 2025-05-29
feat: don't allow integer_primary_key in ETS data layer

artiom:
Don't allow `integer_primary_key` in ETS data layer

Closes https://github.com/ash-project/ash/issues/1500

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/data_layer/ets/ets.ex
+++ b/lib/ash/data_layer/ets/ets.ex
@@ -42,7 +42,10 @@ defmodule Ash.DataLayer.Ets do
 
   use Spark.Dsl.Extension,
     sections: [@ets],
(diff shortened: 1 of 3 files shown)


zachdaniel:
I don't think we can solve this as shown.

You can still have integer primary keys, it's just expected that you do something yourself to set the value. What we could instead do is raise an error somewhere in the ETS data layer's create logic if a generated attribute's value isn't supplied by the action, since nothing gets generated for it by ETS.

artiom:
> You can still have integer primary keys, it's just expected that you do something yourself to set the value

Makes sense. For some reason, haven't thought about supplying my own value in this situation 🤦 
Thanks for suggestion.

PR #2086, opened 2025-06-01
Correct error message, example and doc for match constraint.

octowombat:
# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/dsls/DSL-Ash.Resource.md
+++ b/documentation/dsls/DSL-Ash.Resource.md
@@ -35,7 +35,7 @@ attributes do
     allow_nil? false
 
     constraints [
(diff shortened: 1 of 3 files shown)


octowombat:
BTW: In my case I have the regex's in a separate module as arity 0 functions that return the regex, which is why I spotted this minor issue. I need them to be that way for other parts of the code base so I in fact used this (not pretty I accept) approach to avoid duplications:

``` elixir
    attribute :name, :string do
      constraints min_length: 4, max_length: 128, match: Regex.source(Constants.name_regex())
      public? true
    end
```

PR #2083, opened 2025-06-01
Fix Ash.calculate/2 with opts

nallwhy:
This PR fixes following errors.

`{:error, %Spark.Options.ValidationError{message: "unknown options [:context], valid options are: [:args, :refs, :actor, :scope, :tenant, :authorize?, :tracer, :record, :data_layer?, :reuse_values?, :domain]", value: %{}, key: :context, __exception__: true, keys_path: []}}`

`{:error, %Spark.Options.ValidationError{message: "invalid value for :authorize? option: expected boolean, got: nil", value: nil, key: :authorize?, __exception__: true, keys_path: []}}`

# Contributor checklist

- [x] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

PR #2087, opened 2025-06-02
Fix Ash.run_action/2 with opts

nallwhy:
This PR is similar to #2083 and fixes the following error:

`{:error, %Spark.Options.ValidationError{message: "invalid value for :authorize? option: expected boolean, got: nil", value: nil, key: :authorize?, __exception__: true, keys_path: []}}`

One thing I’m curious about: I’ve noticed that there are two different ways of representing “boolean or nil” in Spark:

- `{:in, [true, false, nil]}`
- `{:or, [:boolean, {:literal, nil}]}`

While the latter seems a bit more explicit in terms of meaning, I’m wondering which style is preferred.
Either way, I think it might be good to standardize the usage throughout the codebase, and I’d like to include that change in this PR as well.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -819,7 +819,7 @@ defmodule Ash do
       """
     ],
     authorize?: [
-      type: :boolean,
+      type: {:in, [true, false, nil]},
       doc: """
       Whether or not the request should be authorized.
       """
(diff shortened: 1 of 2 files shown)


zachdaniel:
So, I actually have a different idea. Let's put these back to booleans, and modify the code in `Ash.Scope.ToOpts` to omit nil values when generating opts. Then we don't have to make these types more complex.

PR #2096, opened 2025-06-03
Inspect: put __meta__ and relationships at the end

madjar:
This is nicer in long list when records are truncated, because
relationships are likely to be `#Ash.NotLoaded<:relationship, field: ...>`
and `__meta__` to be `#Ecto.Schema.Metadata<:loaded>`.

Before:
```
%Ash.Test.Resource.ResourceTest.Leg{
  pants: #Ash.NotLoaded<:relationship, field: :pants>,
  __meta__: #Ecto.Schema.Metadata<:loaded>,
  id: "112d3351-852b-44e4-a715-6f46436c38b9",
  side: "right",
  pants_id: "2ac24627-542f-4e23-8cd4-f1b18634af53"
}
```

After:
```
%Ash.Test.Resource.ResourceTest.Leg{
  id: "f112cd6d-fe2f-4bec-98f2-13d2ad87cf3d",
  side: "right",
  pants_id: "d68fb517-f94c-4384-94c9-79e5a35e1a53",
  pants: #Ash.NotLoaded<:relationship, field: :pants>,
  __meta__: #Ecto.Schema.Metadata<:loaded>
}
```

# Test plan

I couldn't find any test related to this, so I tested this by adding a `dbg` to a test resource, and did a before/after.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource.ex
+++ b/lib/ash/resource.ex
@@ -214,42 +214,32 @@ defmodule Ash.Resource do
                        |> Enum.map(& &1.name)
                        |> Enum.concat([:__meta__, :calculations, :aggregates])
 
-          Code.ensure_loaded!(Inspect.Map)
+          def inspect(record, opts) do
+            record = %{
+              record
+              | calculations:
+                  Ash.Actions.Read.Calculations.map_without_calc_deps(record.calculations)
+            }
 
-          if function_exported?(Inspect.Map, :inspect, 4) do
-            def inspect(record, opts) do
-              record = %{
-                record
-                | calculations:
-                    Ash.Actions.Read.Calculations.map_without_calc_deps(record.calculations)
-              }
+            subtract_fields =
+              Enum.filter([:calculations, :aggregates], &(Map.get(record, &1) == %{}))
 
-              infos =
-                for %{field: field} = info <- record.__struct__.__info__(:struct),
-                    field in @show_fields,
-                    not (field == :aggregates && record.aggregates == %{}),
-                    not (field == :calculations && record.calculations == %{}),
-                    do: info
+            fields = @show_fields -- subtract_fields
 
-              Inspect.Map.inspect(record, inspect(record.__struct__), infos, opts)
+            fun = fn field, opts ->
+              Inspect.List.keyword({field, Map.get(record, field)}, opts)
             end
-          else
-            def inspect(record, opts) do
-              record = %{
-                record
-                | calculations:
-                    Ash.Actions.Read.Calculations.map_without_calc_deps(record.calculations)
-              }
 
-              infos =
-                for %{field: field} = info <- record.__struct__.__info__(:struct),
-                    field in @show_fields,
-                    not (field == :aggregates && record.aggregates == %{}),
-                    not (field == :calculations && record.calculations == %{}),
-                    do: info
+            open =
+              Inspect.Algebra.color_doc("%" <> inspect(record.__struct__) <> "{", :map, opts)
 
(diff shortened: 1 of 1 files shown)


lib/ash/resource.ex
```diff
@@ -242,11 +241,10 @@ defmodule Ash.Resource do
               }
 
               infos =
-                for %{field: field} = info <- record.__struct__.__info__(:struct),
-                    field in @show_fields,
+                for field <- @show_fields,
                     not (field == :aggregates && record.aggregates == %{}),
                     not (field == :calculations && record.calculations == %{}),
-                    do: info
+                    do: %{field: field}
```
zachdaniel: Couple of issues: 

1. we'll need to add `:aggregates` and `:calculations` if they aren't empty (those were in the struct fields but not `@show_fields`
2. I'm not sure what other fields are inside of the `info` that we get  back from `__info__(:struct)`, so we need to go and find that struct field's `info` and return that here.

madjar: Thanks for the quick review :)

1. It looks like aggregates and calculations and added to `@show_fields` on line 215 (`|> Enum.concat([:__meta__, :calculations, :aggregates])`)
2. On elixir 1.18, `__info__(:struct)` is a list of `%{field: ..., default: ...}`, and only `field` is used in the `Inspect` module. But yeah, this is a bit of internal hackings. I was trying to avoid having to create a map from the `__info__(:struct)` to avoid being quadratic, but I can do that

zachdaniel: Gotcha, okay we're good on #1.

For #2 I'm not sure. My thought is that we don't want to make assumptions about the internals of the struct inspect logic, i.e if someday that add some new field to the field map that we don't know about etc. For such small lists I think it should be fine to do N+1 style lookups etc probably? You could do some kind of `Map.new`.

Alternatively we could just in-line the inspect logic we're delegating to and then we don't have to use struct info at all. That would probably be easier TBH

madjar: I pushed a change to get back the fields from `__info__(:struct)`

madjar: Oh, sorry, I missed your message. I like the idea of inlining the logic better than what I have done, it removes the need for two variants, will be overall simpler. I'll do that instead.

madjar: Here, I just pushed the version that inlines the inspect logic. Tell me if you like it better, and I'll squash all that together nicely.

lib/ash/resource.ex
```diff
@@ -214,42 +214,33 @@ defmodule Ash.Resource do
...
+          def inspect(record, opts) do
+            record = %{
+              record
+              | calculations:
+                  Ash.Actions.Read.Calculations.map_without_calc_deps(record.calculations)
+            }
+
+            fields =
+              for field <- @show_fields,
+                  not (field == :aggregates && record.aggregates == %{}),
```
zachdaniel: I think we can clean this up even more by doing something like this

```elixir
subtract_fields = Enum.filter([:calculations, :aggregates], &(Map.get(record, &1)  == %{}))

fields = @show_fields -- subtract_fields
```

zachdaniel (approved):
This looks much better. One small feedback then its good to go.

madjar:
Incorporated your suggestion, squashed, and rebased on main.

zachdaniel (approved):

zachdaniel:

PR #2102, opened 2025-06-04
fix: add one more verifier for code_interface if args are missing

TwistingTwists:
if args are missing from code_interface `define` but it is defined in the action, then the error is raised. 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -658,9 +658,38 @@ defmodule Ash.CodeInterface do
             arg_params = unquote(arg_params)
 
             params =
-              if is_list(params),
-                do: Enum.map(params, &Map.merge(&1, arg_params)),
-                else: Map.merge(params, arg_params)
+              if is_list(params) do
+                Enum.map(params, &Map.merge(&1, arg_params))
+              else
+                try do
+                  Map.merge(params, arg_params)
+                rescue
+                  Protocol.UndefinedError ->
+                    action = unquote(Macro.escape(action))
+                    interface = unquote(Macro.escape(interface))
+                    
+                    reraise Ash.Error.Framework.AssumptionFailed.exception(
+                      message: """
+                      Possible code interface argument mismatch for #{interface.name}.
+                      
+                      You called the function with what appears to be a positional argument,
+                      but the interface expects a params map/keyword list.
+                      
+                      Action `#{action.name}` has arguments: #{inspect(Enum.map(action.arguments, & &1.name))}
+                      Interface `#{interface.name}` declares args: #{inspect(interface.args || [])}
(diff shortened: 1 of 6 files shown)


lib/ash/resource/verifiers/verify_code_interface_argument_completeness.ex
```diff
@@ -0,0 +1,95 @@
+defmodule Ash.Resource.Verifiers.VerifyCodeInterfaceArgumentCompleteness do
```
zachdaniel: It is not necessary for users to define all required arguments as positional arguments in their code interfaces. it is perfectly acceptable to do things like `MyApp.Accounts.register_user!(%{email:  "email", password: "password"})`. We can add the error handling, but not this verifier.

lib/ash/code_interface.ex
```diff
@@ -658,9 +658,38 @@ defmodule Ash.CodeInterface do
...
-                else: Map.merge(params, arg_params)
+              if is_list(params) do
+                Enum.map(params, &Map.merge(&1, arg_params))
+              else
+                try do
+                  Map.merge(params, arg_params)
+                rescue
+                  Protocol.UndefinedError ->
+                    action = unquote(Macro.escape(action))
+                    interface = unquote(Macro.escape(interface))
```
zachdaniel: unquoting these is really expensive at compile time as it shoves the entire interface into the module body for every function. Let's simplify the error message. IIRC I recommended the try rescue, but I've changed my mind. I think what we should do is something like this:

```elixir
params = 
  if is_list(params) do
    Enum.map(params, fn item -> 
      if is_map(item) do
        Map.merge(item, arg_params)
      else
        raise ArgumentError, """
        Expected `params` to be a map or a list of maps.
        Got:  #{inspect(params)}
        """
      end
    end)
  else
     if is_map(params) do
        Map.merge(params, arg_params)
      else
        raise ArgumentError, """
        Expected `params` to be a map or a list of maps.
        Got:  #{inspect(params)}
        """
      end
  end
```

lib/ash/resource/verifiers/verify_code_interface_argument_completeness.ex
TwistingTwists: I see. I almost never do things this way, so. I'll remove the verifier. and only have error handling.

lib/ash/code_interface.ex
TwistingTwists: updated PR

zachdaniel (approved):

zachdaniel:

PR #2105, opened 2025-06-05
chore(CI): Make the community files check optional.

jimsynz:


zachdaniel:
Don't mind making it optional, but which repos are we planning on allowing to bypass the checks?

PR #2107, opened 2025-06-05
Feat/assert stripped

swrenn:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -25,6 +25,7 @@ spark_locals_without_parens = [
   ash_step: 1,
   ash_step: 2,
   ash_step: 3,
+  assert_stripped: 1,
   assume_casted?: 1,
   async?: 1,
   atomic_update: 1,
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -25,7 +25,6 @@ spark_locals_without_parens = [
   ash_step: 1,
   ash_step: 2,
   ash_step: 3,
-  assert_stripped: 1,
   assume_casted?: 1,
   async?: 1,
   atomic_update: 1,
@@ -330,6 +329,11 @@ spark_locals_without_parens = [
   writable?: 1
 ]
 
+locals_without_parens = [
+  {:assert_stripped, 1}
+  | spark_locals_without_parens
+]
+
 [
   import_deps: [:spark, :reactor],
   inputs: [
@@ -337,8 +341,8 @@ spark_locals_without_parens = [
     "{config,lib,test,benchmarks,flames}/**/*.{ex,exs}"
   ],
   plugins: [Spark.Formatter],
-  locals_without_parens: spark_locals_without_parens,
+  locals_without_parens: locals_without_parens,
   export: [
-    locals_without_parens: spark_locals_without_parens
+    locals_without_parens: locals_without_parens
   ]
 ]
--- a/lib/ash/test.ex
+++ b/lib/ash/test.ex
@@ -4,6 +4,7 @@ defmodule Ash.Test do
   """
 
   require ExUnit.Assertions
+  import ExUnit.Assertions
 
   @doc """
   Assert that the given changeset, query, or action input has a matching error.
@@ -26,7 +27,7 @@ defmodule Ash.Test do
     error = splode.to_class(error)
 
     if error_class do
-      ExUnit.Assertions.assert(error.__struct__ == error_class,
+      assert(error.__struct__ == error_class,
         message:
           "Expected the value to have errors of class #{inspect(error_class)}, got: #{inspect(error.__struct__)}"
       )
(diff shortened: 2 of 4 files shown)


swrenn:
@zachdaniel, is the new error okay, `Ash.Error.Framework.Unsupported`? I didn't see an existing error that seemed to fit.

swrenn:
Also, perhaps I should import `ExUnit.Assertions.assert/1` in `lib/ash/test.ex`, as the code delegates to `assert/1`?

lib/ash/test.ex
```diff
@@ -208,4 +208,82 @@ defmodule Ash.Test do
...
+
+  defmacro assert_stripped({:not, _meta, [{:in, _meta2, [left, right]}]}) do
+    quote do
+      assert strip_metadata(unquote(left)) not in strip_metadata(unquote(right))
+    end
+  end
+
+  defmacro assert_stripped(expr) do
+    quote do
+      raise Ash.Error.Framework.Unsupported,
```
zachdaniel: I think in this case just an `ArgumentError` is fine. This error would never be raised from application code/need to be handled in any special way.

zachdaniel:
> Also, perhaps I should import ExUnit.Assertions.assert/1 in lib/ash/test.ex, as the code delegates to assert/1?

Yes, good idea 👍

swrenn:
> > Also, perhaps I should import ExUnit.Assertions.assert/1 in lib/ash/test.ex, as the code delegates to assert/1?
> 
> Yes, good idea 👍

I just noticed the code is making fully-qualified calls to `ExUnit.Assertions.assert/1`. Should I follow that pattern or import?

zachdaniel:
Your import change is good, but we actually have to do it inside the `quote` blocks for it to be effective for those cases 😄

swrenn:
> Your import change is good, but we actually have to do it inside the `quote` blocks for it to be effective for those cases 😄

Whoops 🤦‍♂️. I added fully-qualified calls to `ExUnit.Assertions.assert/1` and reverted the import change so this PR isn't touching unrelated code.

lib/ash/test.ex
```diff
@@ -208,4 +208,84 @@ defmodule Ash.Test do
...
+
+  defmacro assert_stripped({:!==, _meta, [left, right]}) do
+    quote do
+      ExUnit.Assertions.assert(strip_metadata(unquote(left)) !== strip_metadata(unquote(right)))
+    end
+  end
+
+  defmacro assert_stripped({:in, _meta, [left, right]}) do
+    quote do
+      ExUnit.Assertions.assert(strip_metadata(unquote(left)) in strip_metadata(unquote(right)))
```
zachdaniel, line 272: Even w/ fully qualified calls, its a macro, so these macros should either import or require `ExUnit.Assertions`

swrenn: I apologize for my noob mistakes. What is the impact of injecting import statements at every call site? Is that the right solution? Is there a case here for `__using__`?

zachdaniel: It's trivial IMO, especially as it's test only code. But I would keep it fully qualified and just do `require ExUnit.Assertions`.

zachdaniel: Also, not noob mistakes this kind of metaprogramming has a bunch of associated trickeries that hits seasoned devs all the time, myself included.

.formatter.exs
```diff
@@ -25,6 +25,7 @@ spark_locals_without_parens = [
   ash_step: 1,
   ash_step: 2,
   ash_step: 3,
+  assert_stripped: 1,
```
zachdaniel: Ah, just noticed this. This will need to go in a separate list below, and get added to this list. This list of locals is generated from the DSL definitions.

Then its good 😄

swrenn: Do you see something like this often in Elixir code?

```
locals_without_parens = [
  {:assert_stripped, 1}
  | spark_locals_without_parens
]
```

Or is this better?

```
other_locals_without_parens = [
  assert_stripped: 1
]

locals_without_parens =
  spark_locals_without_parens ++
    other_locals_without_parens
```

zachdaniel: First one is best 👍

swrenn:
One last question, will the functions and macros be automatically ordered alphabetically in the documentation? Should I maintain alphabetical order in the code?

zachdaniel:
Try it and see 😄 `mix docs && open doc/index.html`

swrenn:
> Try it and see 😄 `mix docs && open doc/index.html`

Cool. I didn't know it was that easy. After seeing the docs, I went back and forth with Claude a few times to improve them.

zachdaniel:

PR #2117, opened 2025-06-07
Fix typespecs for Ash.get! and Ash.load!

moxley:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

`Ash.get!()` and `Ash.load!()` actually allow more values than their typespecs suggested. This was causing Dialyzer errors to leak into my application that was using Ash.

- `Ash.get!()` can return `nil` under a certain condition (i.e., with the `error?: false` option).
- `Ash.load!()` can take `nil` as the first argument.


zachdaniel:

PR #2128, opened 2025-06-11
fix(calculations): correct path construction order for nested relationship rewrites

### LINKED ISSUE #2127: Error Loading three nested relationships with `no_attributes?: true`
**Describe the bug**
Given a resource structure in which `Grandparent` has_many `Parent` has_many `Child` has_many `Toy`, 

And in which the `Grandparent` also has_many `:grandchild_toys` using a `no_attributes?: true` relationship,

When the `Child` tries to simultaneously load three separate Calculations (`:foo`, `:bar` and `:baz`) which themselves all need to load `[parent: [grandparent: [:grandchild_toys]]`,

The following code raises an exception:
```elixir

child |> Ash.load!([:foo, :bar, :baz])

# raises

%Ash.Error.Unknown.UnknownError{
  error: "** (KeyError) key :grandparent not found in: %ErrorDemo.Demo.Child
 ...
}
```

**To Reproduce**
[A minimal set of resource definitions and calls that can reproduce the bug.](https://github.com/enoonan/error_demo)

**Expected behavior**
I expect the calculations and relationships to load successfully without raising an exception.

**Runtime**
 - Elixir version: 1.18.2
 - Erlang version: 27
 - OS: MacOS, Sonoma 14.6.1
 - Ash version: 3.5.17
 - AshPostgres version: 2.6.6

**Additional context**
It's all on the github demo:
https://github.com/enoonan/error_demo

I ran into a very similar bug many months back, which Zach was able to fix, but I recall it was thorny.

zachdaniel:
When loading multiple calculations that depend on nested relationships with no_attributes?: true, the rewrite system was constructing relationship traversal paths in reverse order, causing KeyError when trying to access relationships that don't exist on the current record.

The issue occurred in get_rewrites/4 where relationship names were being appended to paths instead of prepended, creating paths like [{:rel, :grandparent}, {:rel, :parent}] instead of the correct [{:rel, :parent}, {:rel, :grandparent}].

This caused the system to try accessing :grandparent directly on Child records instead of traversing Child → Parent → Grandparent.

Fixed by changing path construction from append to prepend:
- `path ++ [{:rel, name}]` → `[{:rel, name} | path]`

This ensures rewrite paths match actual relationship traversal order.

Closes #2127

### DIFF
--- a/lib/ash/actions/read/calculations.ex
+++ b/lib/ash/actions/read/calculations.ex
@@ -861,15 +861,16 @@ defmodule Ash.Actions.Read.Calculations do
           query
           |> get_all_rewrites(calculation, path)
           |> Enum.map(fn {{path, data, calc_name, calc_load}, source} ->
-            {{path ++ [{:rel, name}], data, calc_name, calc_load}, source}
+            {{[{:rel, name} | path], data, calc_name, calc_load}, source}
           end)
         end
 
       {name, query} ->
-        query
-        |> get_all_rewrites(calculation, path)
+        nested_rewrites = query |> get_all_rewrites(calculation, path)
+
+        nested_rewrites
         |> Enum.map(fn {{path, data, calc_name, calc_load}, source} ->
-          {{path ++ [{:rel, name}], data, calc_name, calc_load}, source}
+          {{[{:rel, name} | path], data, calc_name, calc_load}, source}
         end)
     end)
     |> Enum.concat(load_relationship_rewrites(ash_query, calculation, path))
@@ -898,7 +899,7 @@ defmodule Ash.Actions.Read.Calculations do
                     path
                   )
                   |> Enum.map(fn {{path, data, calc_name, calc_load}, source} ->
-                    {{path ++ [{:rel, name}], data, calc_name, calc_load}, source}
+                    {{[{:rel, name} | path], data, calc_name, calc_load}, source}
                   end)
               ]
 
@@ -925,7 +926,7 @@ defmodule Ash.Actions.Read.Calculations do
         path
       )
       |> Enum.map(fn {{path, data, calc_name, calc_load}, source} ->
-        {{path ++ [{:rel, calculation.opts[:relationship]}], data, calc_name, calc_load}, source}
+        {{[{:rel, calculation.opts[:relationship]} | path], data, calc_name, calc_load}, source}
       end)
     end)
   end
--- a/test/calculations/calculation_test.exs
+++ b/test/calculations/calculation_test.exs
@@ -419,6 +419,164 @@ defmodule Ash.Test.CalculationTest do
     end
   end
 
+  # Test resources for no_attributes relationship calculation bug
+  defmodule Toy do
+    use Ash.Resource,
+      domain: Domain,
+      data_layer: Ash.DataLayer.Ets
+
+    ets do
+      private?(true)
+    end
+
+    actions do
+      default_accept :*
+      defaults([:read, :destroy, create: :*, update: :*])
+    end
+
+    relationships do
+      belongs_to :child, Ash.Test.CalculationTest.Child, public?: true
+    end
+
+    attributes do
+      uuid_primary_key :id
+    end
+  end
+
+  defmodule Child do
+    use Ash.Resource,
+      domain: Domain,
+      data_layer: Ash.DataLayer.Ets
+
+    ets do
+      private?(true)
+    end
+
+    relationships do
+      belongs_to :parent, Ash.Test.CalculationTest.Parent, public?: true
+      has_many :toys, Ash.Test.CalculationTest.Toy, public?: true
+    end
+
+    actions do
+      default_accept :*
+      defaults([:read, :destroy, create: :*, update: :*])
+    end
+
+    attributes do
+      uuid_primary_key :id
+    end
+
+    calculations do
+      calculate :calc_foo, :string, Ash.Test.CalculationTest.Child.CalcFoo
+      calculate :calc_bar, :string, Ash.Test.CalculationTest.Child.CalcBar
+      calculate :calc_baz, :string, Ash.Test.CalculationTest.Child.CalcBaz
+    end
+
+    defmodule CalcFoo do
+      use Ash.Resource.Calculation
+
+      @impl true
+      def load(_, _, _) do
+        [
+          parent: [grandparent: [:grandchild_toys]]
+        ]
+      end
+
+      @impl true
+      def calculate(records, _, _) do
+        records |> Enum.map(fn _ -> "foo" end)
+      end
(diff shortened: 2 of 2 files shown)


zachdaniel:
For the curious, here is the plan file. Claude wrote all of this code:

# Fix: KeyError when loading multiple calculations with no_attributes relationships

## Bug Summary
When loading three calculations simultaneously on a Child resource, where each calculation needs to load `[parent: [grandparent: [:grandchild_toys]]]` and the `:grandchild_toys` relationship uses `no_attributes?: true`, a KeyError is raised: "key :grandparent not found in: %Child{...}".

## Root Cause
The issue occurs in the calculation rewrite logic in `ash/lib/ash/actions/read/calculations.ex`. When processing calculation dependencies that involve `no_attributes?: true` relationships, the rewrite system was constructing relationship traversal paths in reverse order.

The problem was in the path construction logic where relationships were being **appended** to paths instead of **prepended**, causing the system to try accessing relationships in the wrong order:
- **Expected**: `[{:rel, :parent}, {:rel, :grandparent}]` (Child → Parent → Grandparent)  
- **Actual**: `[{:rel, :grandparent}, {:rel, :parent}]` (trying to find :grandparent directly on Child)

## Existing Usage Rules Violations
This violates the principle that relationship paths should be constructed in the correct traversal order, following the natural flow from source to destination through intermediate relationships.

## Reproduction Test
```elixir
test "loading multiple calculations with no_attributes relationship should not raise KeyError", %{
  child: child
} do
  # This should not raise a KeyError
  result = Ash.load!(child, [:calc_foo, :calc_bar, :calc_baz])

  assert result.calc_foo == "foo"
  assert result.calc_bar == "bar"
  assert result.calc_baz == "baz"
end
```

## Test Output (Before Fix)
```
** (KeyError) key :grandparent not found in: %Child{...}
  (erts 15.1.2) :erlang.map_get(:grandparent, %Child{...})
  (elixir 1.18.4) lib/map.ex:318: Map.update!/3
  (ash 3.5.17) lib/ash/actions/read/calculations.ex:762: Ash.Actions.Read.Calculations.rewrite_at_path/2
```

## Proposed Solution
Modify the path construction logic in `get_rewrites/4` function to use **prepend** instead of **append** when building relationship paths. This ensures the paths are constructed in the correct traversal order.

The fix involves changing:
```elixir
# Before (incorrect):
{{path ++ [{:rel, name}], data, calc_name, calc_load}, source}

# After (correct):
{{[{:rel, name} | path], data, calc_name, calc_load}, source}
```

## Changes Required
1. File: `ash/lib/ash/actions/read/calculations.ex` - Fix 4 instances of path construction in `get_rewrites/4` function

## Potential Side Effects
- Minimal risk: The change only affects the order of path construction
- Should not impact normal relationship loading behavior  
- May fix other edge cases where nested relationship paths were constructed incorrectly

## Regression Prevention
- The existing test should pass after the fix
- All existing calculation loading tests should continue to pass
- Add specific test coverage for `no_attributes?: true` relationship scenarios

## Implementation Log

### Analysis
The root cause was traced to incorrect path construction order in the calculation rewrite system. When processing nested relationships like `[parent: [grandparent: [:grandchild_toys]]]`, the system was building rewrite paths in reverse order through recursive calls to `get_all_rewrites`.

### Root Cause Location
The issue was in multiple path construction locations in `get_rewrites/4` function:
- Line 863: `{{path ++ [{:rel, name}], data, calc_name, calc_load}, source}` (append)
- Line 870: `{{path ++ [{:rel, name}], data, calc_name, calc_load}, source}` (append)  
- Line 901: `{{path ++ [{:rel, name}], data, calc_name, calc_load}, source}` (append)
- Line 926: `{{path ++ [{:rel, calculation.opts[:relationship]}], data, calc_name, calc_load}, source}` (append)

This caused the rewrite system to try accessing `:grandparent` directly on Child records instead of traversing Child → Parent → Grandparent.

### Solution Implemented
Changed all path construction from **append** to **prepend**:
```elixir
# Before (incorrect):
{{path ++ [{:rel, name}], data, calc_name, calc_load}, source}

# After (correct):
{{[{:rel, name} | path], data, calc_name, calc_load}, source}
```

This ensures that when relationships are processed recursively, the paths are built in the correct order: the outermost relationship (closest to the source record) comes first in the path.

### Files Changed
1. `ash/lib/ash/actions/read/calculations.ex` - Fixed 4 instances of incorrect path construction

### Final Implementation
Fixed path construction order by prepending relationship names instead of appending them. This ensures the rewrite paths match the actual relationship traversal order from the source record through nested relationships.

### Test Results
- Reproduction test: PASSING ✅
- Full calculation test suite: 64 tests, 0 failures ✅  
- Full test suite: 1668 tests, 0 failures ✅
- New tests added: 2 test cases integrated into `test/calculations/calculation_test.exs`

### Verification Checklist
- [x] Bug is fixed - KeyError no longer occurs when loading multiple calculations with no_attributes relationships
- [x] No regressions introduced - All existing tests pass
- [x] Tests cover the fix - Added specific test cases integrated into main calculation test suite
- [x] Code follows patterns - Used existing Elixir list prepend pattern `[item | list]`
- [x] Test integration - Merged test cases into existing `calculation_test.exs` instead of standalone file

## Questions for Zach
1. ~~Should we provide a default value (like `%Ash.NotLoaded{}`) when the relationship field is missing, or handle it differently?~~ **RESOLVED**: The issue was incorrect path construction, not missing fields.
2. ~~Are there other places in the codebase where similar `Map.update!/3` patterns might cause issues with `no_attributes?: true` relationships?~~ **RESOLVED**: The `Map.update!/3` calls were correct - the paths being passed to them were wrong.

## Technical Details

### Debug Process
The issue was discovered through systematic debugging:
1. Added debug output to trace rewrite path construction
2. Identified that paths like `[{:rel, :grandparent}, {:rel, :parent}]` were backwards
3. Traced the path construction through the recursive `get_all_rewrites` calls
4. Found that `path ++ [{:rel, name}]` was appending instead of prepending
5. Changed to `[{:rel, name} | path]` to build paths in correct order

### Why the Fix Works
The calculation system processes nested loads recursively:
1. `[:grandchild_toys]` creates path `[]`
2. `[grandparent: [:grandchild_toys]]` processes child path `[]` and should prepend `:grandparent` → `[{:rel, :grandparent}]`
3. `[parent: [grandparent: [:grandchild_toys]]]` processes child path `[{:rel, :grandparent}]` and should prepend `:parent` → `[{:rel, :parent}, {:rel, :grandparent}]`

This creates the correct traversal order: Child → Parent → Grandparent.

### Impact
This fix resolves not just the specific `no_attributes?: true` issue, but any scenario where nested relationship calculations might have had incorrect path construction. The fix is minimal, safe, and follows Elixir's natural list prepend patterns.

PR #2130, opened 2025-06-12
fix(code_interface): add notify?: true for bulk creates

chazwatkins:
## Changes

- Code Interface: Add `Keyword.put(:notify?, true)` to `bulk_options` for bulk creates so notifications are sent by default like `:update` and `:destroy` actions

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -926,6 +926,7 @@ defmodule Ash.CodeInterface do
                         bulk_opts =
                           opts
                           |> Keyword.delete(:bulk_options)
+                          |> Keyword.put(:notify?, true)
                           |> Keyword.merge(Keyword.get(opts, :bulk_options, []))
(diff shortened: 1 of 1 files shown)


.tool-versions
```diff
@@ -1,2 +0,0 @@
-erlang 27.1.2
```
zachdaniel: I think you deleted this by accident.

zachdaniel (approved):

.tool-versions
chazwatkins: woops, sorry.  I'll add it back

chazwatkins:
@zachdaniel Reverted .tool-versions so it only includes the notify changes for bulk create.

zachdaniel:

PR #2132, opened 2025-06-12
add nil to Ash.get spec

diogomrts:
we can get {:ok, nil} if error? opt is false


zachdaniel (approved):

zachdaniel:

Issue #2081, opened 2025-05-31
Proposal: Use relationships in PubSub topics to fan out


pikdum:
**Is your feature request related to a problem? Please describe.**

I'd like a way to use relationships in pubsub topics, especially with has_many relationships through a join table.

Using a social media site as an example, this would allow broadcasting new posts to all followers using a relationship like `post.author.followers.user_id` to publish to `post:$follower_id` topics.

**Describe the solution you'd like**

* some way use relationships in publish topics
* list of publish topics is generated from that + event is published to all

**Describe alternatives you've considered**

* publish to a more broad topic + filter once a process receives it
* have the client dynamically generate specific topics to follow
* custom notifier

**Express the feature either with a change to resource syntax, or with a change to the resource interface**

Since `publish` and `publish_all` use lists for combinations, maybe tuples could be used for relationship attributes?

```elixir
  pub_sub do
    prefix "post"
    module SocialMediaWeb.Endpoint    
    publish_all :create, [{:author, :followers, :user_id}]
    publish :update, ["foobar", [{:author, :followers, :user_id}, :_pkey]]
  end
```

**Additional context**

https://discord.com/channels/711271361523351632/1298629277524295700/1377340286405181461

One thing I'm not certain about is performance - doing queries + generating these topics + publishing to each could be slow with larger relationshpis.

dev-guy:
Performance will be awful for popular sites. Consider integrating with ash oban optionally.

zachdaniel:
I think providing choice on this front makes sense. However I think what we should do here instead is allow configuring a function or a module that implements a behavior that returns a list of topics, that way the logic can just be "whatever you want".

```elixir
publish_all :create, {Fanout,...opts}
```

I won't likely do this myself any time soon, but I'd support the latter implementation.

jechol:
It seems a PR was opened for this, but it hasn’t been merged, and the issue remains open.
I just wanted to check if this means the feature hasn’t been fully addressed yet.

barnabasJ:
I started working on it, but there is more work needed, and priorities shifted, so I haven't had time to work on it more.

PR #2139, opened 2025-06-13
feat: add function and module-based topic generation to PubSub

barnabasJ:
This addresses GitHub issue #2081

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/create/create.ex
+++ b/lib/ash/actions/create/create.ex
@@ -21,6 +21,7 @@ defmodule Ash.Actions.Create do
     else
       {changeset, opts} = Ash.Actions.Helpers.set_context_and_get_opts(domain, changeset, opts)
       changeset = Helpers.apply_opts_load(changeset, opts)
+      changeset = Ash.Notifier.PubSub.Helpers.add_topic_generator_loads(changeset)
 
       Ash.Tracer.span :action,
                       fn ->
--- a/lib/ash/notifier/pub_sub/helpers.ex
+++ b/lib/ash/notifier/pub_sub/helpers.ex
@@ -1,0 +1,110 @@
+defmodule Ash.Notifier.PubSub.Helpers do
+  @moduledoc """
+  Helper functions for PubSub notifier integration.
+  """
+
+  alias Ash.Notifier.PubSub
+
+  @doc """
+  Adds required loads from topic generators to a changeset or query.
+
+  This function inspects all PubSub publications configured for the resource
+  and action, extracts load requirements from topic generators that implement
+  the `required_loads/0` callback, and adds them to the changeset/query.
+
+  This ensures that when notifications are created, the data already has
+  the required relationships loaded, preventing N+1 queries in bulk operations.
+  """
+  def add_topic_generator_loads(%Ash.Changeset{} = changeset) do
+    add_loads_for_action(changeset, changeset.resource, changeset.action)
+  end
+
+  def add_topic_generator_loads(%Ash.Query{} = query) do
+    # For queries, we might need to handle this differently
+    # For now, return unchanged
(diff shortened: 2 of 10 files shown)


zachdaniel:
So, I think there is one particular thing we should do before we support this, to solve for bulk actions, which is that we'd add a `load` callback. However, what we need to do for that is actually pretty complicated. Specifically, we can't clobber existing load statements on the data. The "trick" would be to create a new calculation for each notifier that requires loads, and dynamically add it for each of those notifiers, and have that dynamic calculation return the loads required by the notifier as a map. This would look kind of like the `LoadRelationship` calculation. The notifier would have an interface like this:
```elixir
@callback required_loads :: %{term() => load_statement()}
```

The key of that map would be to disambiguate for the notifier if multiple separate notifications trigger on that action.

So in the read action, we'd add calculations like `Ash.Query.Calculation.new({:notifier_dependencies, Notifier, key}, {ABuiltInModuleCalculationForThis, load: [load]})` for each notifier/key combination, that returns a map where the key is each load in `load`. Then, before passing the data into the notifier, we'd merge those loads with/ the calculation data.

Writing this all up has pointed out how complex the problem space is 😆. However, it is the pattern we'd need for notifiers to automatically load their dependencies. We can potentially side-line this for now, but I worry about having this pattern in place in a way that plays so poorly with bulk actions.

lib/ash/notifier/pub_sub/topic_generator.ex
```diff
@@ -80,25 +82,44 @@ defmodule Ash.Notifier.PubSub.TopicGenerator do
...
+    [:author, :tags, :comments]
   end
   ```
   """
-  @callback generate_topics(notification :: Ash.Notifier.Notification.t(), opts :: keyword()) ::
-              String.t()
-              | [
-                  String.t()
-                ]
+  @callback required_loads() :: term()
```
zachdaniel, line 120: This should be a callback on `Ash.Notifier` I think.

lib/ash/actions/create/create.ex
```diff
@@ -21,6 +21,7 @@ defmodule Ash.Actions.Create do
     else
       {changeset, opts} = Ash.Actions.Helpers.set_context_and_get_opts(domain, changeset, opts)
       changeset = Helpers.apply_opts_load(changeset, opts)
+      changeset = Ash.Notifier.PubSub.Helpers.add_topic_generator_loads(changeset)
```
zachdaniel, line 24: We shouldn't have behavior for a specific notifier here, it should be calling a callback on all of the notifiers for this resource and action.

lib/ash/notifier/pub_sub/topic_generator.ex
zachdaniel: Ah, nvm. This makes sense here. But my other comment about having a callback on the notifier itself to drive this is the thing.

lib/ash/notifier/pub_sub/helpers.ex
```diff
@@ -0,0 +1,110 @@
...
+          single_load ->
+            [single_load]
+        end
+
+      case changeset_or_query do
+        %Ash.Changeset{} = changeset ->
+          Ash.Changeset.load(changeset, load_list)
+
+        %Ash.Query{} = query ->
+          Ash.Query.load(query, load_list)
```
zachdaniel, line 104: We can't just load the things here, because the caller of the action could request the same load key but with different parameters which can lead to non deterministic behavior. This is what I was getting at with needing to add a custom calculation.

zachdaniel:
Closing due to staleness.

PR #2154, opened 2025-06-20
Read Action multitenancy :bypass_all

xantrac:
This PR allows the user to setup a new value `bypass_all` for the `multitenancy` option in the read action. With `bypass_all` the user can recursively bypass the tenancy of all the resources successively loaded.

Also this PR removes the `bypass` value from the other actions type since it wasn't actually implemented.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -354,7 +354,7 @@ defmodule Ash.Actions.Read do
                  opts[:authorize?]
                ),
              {:ok, data} <-
-               load_relationships(data, query, opts),
+               load_relationships(data, query_ran, opts),
              {:ok, data} <-
                Ash.Actions.Read.Calculations.run(
                  data,
@@ -366,7 +366,7 @@ defmodule Ash.Actions.Read do
                load_through_attributes(
                  data,
                  %{
-                   query
+                   query_ran
                    | calculations: Map.new(calculations_at_runtime, &{&1.name, &1}),
                      load_through: Map.delete(query.load_through || %{}, :attribute)
                  },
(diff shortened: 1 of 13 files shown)


zachdaniel:
I like the idea, but I think we're going to have to do it with shared context. The tenant can be any value, so for all we know `:bypass` is a valid tenant value for some user.

xantrac:
> I like the idea, but I think we're going to have to do it with shared context. The tenant can be any value, so for all we know `:bypass` is a valid tenant value for some user.

That makes sense

lib/ash/actions/read/read.ex
```diff
@@ -2425,6 +2428,11 @@ defmodule Ash.Actions.Read do
 
       :bypass ->
         {:ok, %{query | tenant: nil, to_tenant: nil}}
+
+      :bypass_all ->
+        query = Ash.Query.set_context(query, %{shared: %{multitenancy: :bypass_all}})
```
zachdaniel, line 2433: using shared context means it will apply to all children, regardless of how deep they are nested. Is this the intent? IIRC you made a good point about it being less confusing if it only extends to direct relationships.

xantrac: yeah I think that was uncertain and you wanted this type of behavior. I think it can be confusing in both ways.

This pattern grew on me because it could be easier to debug knowing the all the descending queries will be untenanted instead of having to check each intermediate step.  You know you users better so I'll leave the final decision to you.

zachdaniel: This way is better 👌

zachdaniel:

xantrac:
> 🚀 Thank you for your contribution! 🚀

Always!

PR #2155, opened 2025-06-20
fix min/max length constraint handling in Ash.Type for array types

kernel-io:
run the min/max length constraints for `{:array, _}` types after all the other constraints have been applied

### DIFF
--- a/lib/ash/type/type.ex
+++ b/lib/ash/type/type.ex
@@ -1203,54 +1203,49 @@ defmodule Ash.Type do
   def apply_constraints({:array, type}, term, constraints) when is_list(term) do
     type = get_type(type)
 
-    list_constraint_errors = list_constraint_errors(term, constraints)
     item_constraints = item_constraints(constraints)
 
-    case list_constraint_errors do
-      [] ->
-        nil_items? = Keyword.get(constraints, :nil_items?, false)
-        remove_nil_items? = Keyword.get(constraints, :remove_nil_items?, false)
+    nil_items? = Keyword.get(constraints, :nil_items?, false)
+    remove_nil_items? = Keyword.get(constraints, :remove_nil_items?, false)
 
-        term
-        |> Enum.with_index()
-        |> Enum.reduce({[], []}, fn {item, index}, {items, errors} ->
-          if type.custom_apply_constraints_array?() do
-            maybe_handle_nil_item(item, index, items, errors, nil_items?, remove_nil_items?)
-          else
-            case apply_constraints(type, item, item_constraints) do
-              {:ok, value} ->
-                maybe_handle_nil_item(value, index, items, errors, nil_items?, remove_nil_items?)
+    {terms, errors} =
+      term
+      |> Enum.with_index()
+      |> Enum.reduce({[], []}, fn {item, index}, {items, errors} ->
+        if type.custom_apply_constraints_array?() do
+          maybe_handle_nil_item(item, index, items, errors, nil_items?, remove_nil_items?)
+        else
+          case apply_constraints(type, item, item_constraints) do
+            {:ok, value} ->
+              maybe_handle_nil_item(value, index, items, errors, nil_items?, remove_nil_items?)
 
-              {:error, new_errors} ->
-                new_errors =
-                  new_errors
-                  |> List.wrap()
-                  |> Ash.Helpers.flatten_preserving_keywords()
-                  |> Enum.map(fn
-                    string when is_binary(string) ->
-                      [message: string, index: index]
+            {:error, new_errors} ->
+              new_errors =
+                new_errors
+                |> List.wrap()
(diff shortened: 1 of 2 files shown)


lib/ash/type/type.ex
```diff
@@ -1285,12 +1285,23 @@ defmodule Ash.Type do
     end
   end
 
+  @doc false
+  defp remove_empty_items(term, constraints) do
+    empty_values = Keyword.get(constraints, :empty_values, [])
```
zachdaniel: lets create one list up front of empty values to be removed. If the list is empty, we don't need to do `Enum.filter` we can just return the `term` unchanged.

zachdaniel: Hm...well...doesn't this mean we're doubling up the remove empty items logic? One place to remove it, and once place to remove it and check length? wouldn't this issue just be an "order of operations" issue in that case? remove empty items before checking length?

kernel-io: yeah I will recommit what I originally had, was having a bit of a brainfart earlier and overwrote it...

lib/ash/type/type.ex
```diff
@@ -1203,6 +1203,8 @@ defmodule Ash.Type do
   def apply_constraints({:array, type}, term, constraints) when is_list(term) do
     type = get_type(type)
 
+    term = remove_empty_items(term, constraints)
```
zachdaniel: I don't think we want to remove empty items *before* applying item constraints, because we don't know what will and won't be empty until we do that.

kernel-io: ahh gothcha, I misread you previous comment, I'll put it back within the length checking part as originally if you'd like.

lib/ash/type/type.ex
```diff
@@ -1285,8 +1285,27 @@ defmodule Ash.Type do
...
+          Enum.filter(term, &(&1 not in [nil | values]))
+        else
+          Enum.filter(term, &(&1 not in values))
+        end
+    end
+  end
+
   @doc false
   def list_constraint_errors(term, constraints) do
+    term = remove_empty_items(term, constraints)
```
zachdaniel: this term is discarded though isn't it? how does this result in the nil items being removed?

zachdaniel: oh, there is already code for this. I think we just need to move the max/min length checks to somewhere like this: https://github.com/ash-project/ash/pull/2155/files#diff-dd4491b1dc2629842f30d2edee72161d609bc73dff8e10a3b3e61f60b4ffc756R1250

After all the other stuff has applied.

kernel-io: the core issue is that nothing currently removes the empty values, the only place it's currently used is here https://github.com/ash-project/ash/blob/e6394c34248b2d8376617f572031b10f886198e6/lib/ash/type/type.ex#L1049-L1051

So this was done to make the length checks work without modifying any other functionality.

I can go ahead and put the list constraints at the end of the constraints checking function, but then I'll still need to drop the empty items.

In the meantime I've modified the PR to not assign unnecessarily and removed some useless tests that I added.

zachdaniel: The point is not doing an extra pass on the list, not about assigning unnecessarily. 

We need something like this:

```elixir
def apply_constraints({:array, type}, term, constraints) when is_list(term) do
    type = get_type(type)

    item_constraints = item_constraints(constraints)

    nil_items? = Keyword.get(constraints, :nil_items?, false)
    remove_nil_items? = Keyword.get(constraints, :remove_nil_items?, false)

    {terms, errors} =
      term
      |> Enum.with_index()
      |> Enum.reduce({[], []}, fn {item, index}, {items, errors} ->
        if type.custom_apply_constraints_array?() do
          maybe_handle_nil_item(item, index, items, errors, nil_items?, remove_nil_items?)
        else
          case apply_constraints(type, item, item_constraints) do
            {:ok, value} ->
              maybe_handle_nil_item(value, index, items, errors, nil_items?, remove_nil_items?)

            {:error, new_errors} ->
              new_errors =
                new_errors
                |> List.wrap()
                |> Ash.Helpers.flatten_preserving_keywords()
                |> Enum.map(fn
                  string when is_binary(string) ->
                    [message: string, index: index]

                  vars ->
                    Keyword.put(vars, :index, index)
                end)

              {[item | items], List.wrap(new_errors) ++ errors}
          end
        end
      end)

    case errors ++ list_constraint_errors(terms, constraints) do
      [] ->
        if type.custom_apply_constraints_array?() do
          case type.apply_constraints_array(Enum.reverse(terms), item_constraints) do
            :ok -> {:ok, term}
            other -> other
          end
        else
          {:ok, Enum.reverse(terms)}
        end

      errors ->
        {:error, errors}
    end
  end
```

zachdaniel:

PR #2161, opened 2025-06-23
chore: Use the same parent query timeout for the count query, for `read` actions

sevenseacat:
As discussed [in Slack](https://discord.com/channels/711271361523351632/1386661506904621147).

When supplying the `timeout` option to `Ash.read`, that timeout is not being used for the `count` query that also gets run (when fetching a total count, for pagination purposes).

Doing some testing via a Cinder table (as that's what the original error was coming from), when specifying `query_opts={[timeout: 10000]}` this gets passed through to the `count` query.

If no `timeout` is specified from the table, `query.timeout` is `:infinity`.

# Contributor checklist

- [x] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -651,7 +651,7 @@ defmodule Ash.Actions.Read do
              results <- add_keysets(query, results, query.sort),
              {:ok, results} <- run_authorize_results(query, results),
              {:ok, results, after_notifications} <- run_after_action(query, results),
-             {:ok, count} <- maybe_await(count) do
+             {:ok, count} <- maybe_await(count, query.timeout) do
           notify_callback.(query, before_notifications ++ after_notifications)
           {:ok, results, count, calculations_at_runtime, calculations_in_query, query}
         else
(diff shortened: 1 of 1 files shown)


lib/ash/actions/read/read.ex
```diff
@@ -651,7 +651,7 @@ defmodule Ash.Actions.Read do
              results <- add_keysets(query, results, query.sort),
              {:ok, results} <- run_authorize_results(query, results),
              {:ok, results, after_notifications} <- run_after_action(query, results),
-             {:ok, count} <- maybe_await(count) do
+             {:ok, count} <- maybe_await(count, query.timeout) do
```
zachdaniel, line 654: It may not necessarily have a timeout, so we would likely want to do something like `query.timeout || :infinity`. Unless we set query.timeout to infinity automatically or something?

sevenseacat: I'm not sure tbh, but in my testing if I didn't set a timeout, it was infinity?

zachdaniel:

PR #2165, opened 2025-06-25
improvement: allow pure-binary sigils in expr fragments

frankdugan3:
# Contributor checklist

- [x] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/expr/expr.ex
+++ b/lib/ash/expr/expr.ex
@@ -823,6 +823,11 @@ defmodule Ash.Expr do
     soft_escape(Ash.CiString.sigil_i(str, mods), escape?)
   end
 
+  def do_expr({:fragment, opts, [{_, _, [{:<<>>, _, [query]}, []]} | rest]}, escape?)
+      when is_binary(query) do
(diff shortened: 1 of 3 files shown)


lib/ash/expr/expr.ex
```diff
@@ -823,6 +823,11 @@ defmodule Ash.Expr do
     soft_escape(Ash.CiString.sigil_i(str, mods), escape?)
   end
 
+  def do_expr({:fragment, opts, [{_, _, [{:<<>>, _, [query]}, []]} | rest]}, escape?)
+      when is_binary(query) do
+    do_expr({:fragment, opts, [Macro.expand(query, __ENV__) | rest]}, escape?)
```
zachdaniel: So this discards whatever compile time transformations are made by the sigil doesn't it? I think we'd want to pass the AST through right?

zachdaniel (dismissed):

lib/ash/expr/expr.ex
frankdugan3: Good catch, I made a few changes that should address that properly.

zachdaniel:

PR #2166, opened 2025-06-26
tests: add change-tracking tests for embedded attributes

rellen:
These tests check that change-tracking in changesets works as expected for changes to attributes that are embedded resources.

We are tracking down a strange bug somewhere in Ash or AshPaperTrail, and these tests imply that the issue is not in Ash, but thought it would be worthwhile to add them to the test suite.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/embedded_resource_test.exs
+++ b/test/embedded_resource_test.exs
@@ -586,4 +586,144 @@ defmodule Ash.Test.Changeset.EmbeddedResourceTest do
     assert Enum.map(applied_author.union_tags_with_id, & &1.value.id) ==
              Enum.map(author.union_tags_with_id, & &1.value.id)
   end
+
+  test "embedded resource change detection works correctly when updating with identical struct" do
+    author =
+      Author
+      |> Changeset.for_create(:create, %{
+        profile: %{first_name: "John", last_name: "Doe"}
+      })
+      |> Ash.create!()
+
+    original_profile = author.profile
+    changeset = Changeset.for_update(author, :update, %{profile: original_profile})
+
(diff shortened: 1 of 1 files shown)


test/embedded_resource_test.exs
```diff
@@ -586,4 +586,144 @@ defmodule Ash.Test.Changeset.EmbeddedResourceTest do
...
+      |> Ash.create!()
+
+    changeset =
+      Changeset.for_update(author, :update, %{
+        profile: %{first_name: "John", last_name: "Doe"}
+      })
+
+    {_result, notifications} = Ash.update!(changeset, return_notifications?: true)
+
+    assert List.first(notifications).changeset.context[:changed?] == false
```
rellen, line 684: Is this an acceptable way to get access to the changeset after the change has been executed?

Another way could be to add an `after_action` hook to send a message with the changeset back to the test process.

zachdaniel: Hmm...yeah, I think the latter would be the preferred way in this scenario. I don't want to necessarily lean on the notification.changeset particulars

zachdaniel: actually this is fine.

zachdaniel: I think it might also be good to assert on the creation or non-creation of versions as that is the primary use case we're concerned with here. However, that can be done in addition to what's already been done, so I'll merge this.

zachdaniel:

PR #2200, opened 2025-07-15
test: added test for modify_query in aggregate with read_action

TommasoAmici:
This is a reproduction for an issue I ran into.
It seems that modify_query is not applied when a read_action is specified for an aggregation.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/actions/aggregate_test.exs
+++ b/test/actions/aggregate_test.exs
@@ -6,6 +6,12 @@ defmodule Ash.Test.Actions.AggregateTest do
 
   alias Ash.Test.Domain, as: Domain
 
+  defmodule Comment.ReadActionModifyQuery do
+    def modify(_ash_query, _ecto_query) do
+      raise "Should raise!"
+    end
+  end
+
   defmodule Comment do
     use Ash.Resource,
       domain: Domain,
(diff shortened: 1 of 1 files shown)


test/actions/aggregate_test.exs
```diff
@@ -469,5 +484,24 @@ defmodule Ash.Test.Actions.AggregateTest do
...
+        })
+        |> Ash.create!(authorize?: false)
+
+      assert_raise Ash.Error.Unknown, ~r/Should raise!/, fn ->
+        Ash.read!(Comment, action: :with_modify_query, authorize?: false)
+      end
+
+      assert_raise RuntimeError, fn ->
+        Ash.load!(post, :count_of_comments_modify_query, authorize?: false)
+      end
```
TommasoAmici, line 504: this one still doesn't raise, it seems it's not going down that path.

https://github.com/ash-project/ash/actions/runs/16304466863/job/46047188470?pr=2200#step:13:398

zachdaniel:

zachdaniel:
@TommasoAmici I've fixed this for the ets data layer, but we also need to reproduce it with AshPostgres as I believe it will have the same issue.

PR #2205, opened 2025-07-17
feat: short-circuit load function when load is empty

barnabasJ:
Add early return clauses to the load function when the load parameter is nil or an empty list, avoiding unnecessary processing.


barnabasJ:
There is already a branch for this that I overlooked.

PR #2212, opened 2025-07-20
improvement: Add Ash.Subject to abstract Changeset, Query, ActionInput

chazwatkins:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Adds `Ash.Subject` to help abstract common functionality between `Changeset`, `Query`, and `ActionInput`.  Most functions act as a router to the subject's corresponding function.  For any functions that seemed to have generic implementations, I moved those into Ash.Subject directly.

No changes have been made to `Changeset`, `Query`, or `ActionInput`.  This PR only adds `Ash.Subject` so it can be used when refactoring other functionality that has to pattern match on the subject type.

Let me know if any changes are needed.

Future PRs will aim to refactor modules that are currently pattern matching on subject type to call the common functions and eventually pull out the logic from the subject type modules into `Ash.Subject` so it's all centralized.


lib/ash/subject.ex
```diff
@@ -0,0 +1,539 @@
...
+  @spec add_error(
+          t(),
+          Ash.Error.error_subject() | list(Ash.Error.error_subject()),
+          Ash.Error.path_subject()
+        ) :: t()
+  def add_error(subject, errors, path \\ [])
+
+  def add_error(subject, [], _path), do: subject
+
+  def add_error(%Ash.Changeset{} = subject, errors, path) when is_list(errors) do
```
zachdaniel: I think we can simplify this interface by *not* accepting path here. You can just `Ash.Error.set_path` on the error before calling these.

lib/ash/subject.ex
```diff
@@ -0,0 +1,539 @@
...
+  @spec set_actor(t(), Keyword.t() | term()) :: t()
+  def set_actor(subject, opts) when is_list(opts) do
+    if Keyword.has_key?(opts, :actor) do
+      set_actor(subject, opts[:actor])
+    else
+      subject
+    end
+  end
+
+  def set_actor(subject, actor) do
```
zachdaniel: AFAIK this is a private function on `Ash.Query` for example. `set_actor|tenant|authorize?|tracer` should be removed here.

zachdaniel: not private w/ `defp` but with `@doc false`

lib/ash/subject.ex
```diff
@@ -0,0 +1,539 @@
...
+
+  Used to store action results for later retrieval.
+
+  ## Parameters
+
+    * `subject` - The subject to set result on
+    * `result` - The result data to store
+  """
+  @spec set_result(t(), term()) :: t()
+  def set_result(subject, result) do
```
zachdaniel: We don't support this on generic actions yet, so we either need to support it there or remove this here.

lib/ash/subject.ex
```diff
@@ -0,0 +1,539 @@
...
+
+  @doc """
+  Clears any stored result from the subject's private context.
+
+  ## Parameters
+
+    * `subject` - The subject to clear result from
+  """
+  @spec clear_result(t()) :: t()
+  def clear_result(subject) do
```
zachdaniel: same as above.

chazwatkins:
Addressed all the feedback items

chazwatkins:
Fixed the dialyzer issue for `add_error` type specs.  Should be good to review once ci completes.

chazwatkins:
I went ahead and added `set_result` and `clear_result` for all subject types

lib/ash/subject.ex
```diff
@@ -0,0 +1,420 @@
...
+  def set_result(%Ash.Changeset{} = subject, result) do
+    Ash.Changeset.set_result(subject, result)
+  end
+
+  def set_result(%Ash.Query{} = subject, result) do
+    Ash.Query.set_result(subject, result)
+  end
+
+  def set_result(subject, result) do
+    put_context(subject, :private, %{action_result: result})
```
zachdaniel: Generic actions don't support this AFAIK. I think we should leave this one out for now 😄

chazwatkins: Ok.  I'll drop the commit and push again

chazwatkins: I dropped that commit.  Let me know if there are any other changes needed.

zachdaniel (approved):

zachdaniel:

zachdaniel:
Awesome 😄 PRing the usage of this to the builtin preparations, validations, and changes would be amazing ❤️

chazwatkins:
> Awesome 😄 PRing the usage of this to the builtin preparations, validations, and changes would be amazing ❤️

I'll work on that this week and get it submitted.

PR #2215, opened 2025-07-21
refactor: Migrate to Ash.Subject for Changeset, Query, ActionInput

chazwatkins:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

This PR was artisanally crafted without AI as it would be too easy for AI to run a muck with this many changes.

There were a couple instances where `Ash.Changeset.add_error/3` is being used, but we removed `Ash.Subject.add_error/3`.  So, I left those function calls as-is.

I removed `Ash.Subject.run_before_actions` and `Ash.Subject.run_after_actions` as I noticed they are marked with `@doc false`, so they are likely private functions.

The next step would be to evaluate `Ash.Action.Helpers` and potentially port some of it into `Ash.Subject`.


zachdaniel:
I think this may be a bit aggressive. When we know 100% that the value is a changeset, query or action input we should use the appropriate underlying interface. Only when it could be multiple types of subjects and we would otherwise have case statements etc. should we then use `Ash.Subject`

chazwatkins:
> I think this may be a bit aggressive. When we know 100% that the value is a changeset, query or action input we should use the appropriate underlying interface. Only when it could be multiple types of subjects and we would otherwise have case statements etc. should we then use `Ash.Subject`

I reverted all changes except for built-ins that were pattern matching on subject type.

zachdaniel:

PR #2221, opened 2025-07-24
improvement: Add prepend? opt to hooks and Ash.Subject transaction hooks

chazwatkins:
## Changes
- Add support for `prepend?` option to Changeset, ActionInput, and Query hooks.  Except for `Ash.Query.after_action` which uses `@read_action_after_action_hooks_in_order?`
- Add transaction hooks to `Ash.Subject` since Changeset, ActionInput, and Query now all support them.
- Update all functions to delegate to the subject type module function
- Align type specs with subject type module types
- Update `get_argument/3` to only return the default value when `:error` is returned from `fetch_argument/2`
- Add `Ash.ActionInput.delete_argument/3`
- Add `Ash.Changeset.fetch_attribute/2`
- Add `Ash.Changeset.fetch_data/2`
- Add `Ash.Changeset.fetch_argument_or_attribute/2`
- Update `Ash.Query.fetch_argument/2` to first call `Ash.Query.new(query)` like other functions to ensure the subject is a query.
- Remove `set_private_argument/3` tests because there is a bug where `new/1` doesn't add the `arguments`, but `set_argument/3` is using dot syntax to access the arguments.  So, there's no way to use `set_private_argument/3` without getting the warning message.  I'll submit another PR to fix this"

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/action_input.ex
+++ b/lib/ash/action_input.ex
@@ -998,10 +998,15 @@ defmodule Ash.ActionInput do
   """
   @spec after_action(
           input :: t(),
-          fun :: after_action_fun()
+          fun :: after_action_fun(),
+          opts :: Keyword.t()
         ) :: t()
-  def after_action(input, func) do
-    %{input | after_action: input.after_action ++ [func]}
+  def after_action(input, func, opts \\ []) do
+    if opts[:prepend?] do
+      %{input | after_action: [func | input.after_action]}
+    else
+      %{input | after_action: input.after_action ++ [func]}
+    end
   end
 
   @doc """
@@ -1025,9 +1030,17 @@ defmodule Ash.ActionInput do
   - `around_transaction/2` for hooks that wrap the entire transaction
   - `before_action/3` for hooks that run before the action (inside transaction)
   """
-  @spec before_transaction(t, before_transaction_fun) :: t
-  def before_transaction(input, func) do
-    %{input | before_transaction: input.before_transaction ++ [func]}
+  @spec before_transaction(
+          input :: t(),
+          fun :: before_transaction_fun(),
+          opts :: Keyword.t()
+        ) :: t()
+  def before_transaction(input, func, opts \\ []) do
+    if opts[:prepend?] do
+      %{input | before_transaction: [func | input.before_transaction]}
+    else
+      %{input | before_transaction: input.before_transaction ++ [func]}
+    end
   end
 
   @doc """
@@ -1051,9 +1064,17 @@ defmodule Ash.ActionInput do
   - `around_transaction/2` for hooks that wrap the entire transaction
   - `after_action/2` for hooks that run after the action (inside transaction)
   """
-  @spec after_transaction(t, after_transaction_fun) :: t
-  def after_transaction(input, func) do
-    %{input | after_transaction: input.after_transaction ++ [func]}
+  @spec after_transaction(
+          input :: t(),
+          fun :: after_transaction_fun(),
+          opts :: Keyword.t()
+        ) :: t()
+  def after_transaction(input, func, opts \\ []) do
+    if opts[:prepend?] do
+      %{input | after_transaction: [func | input.after_transaction]}
+    else
+      %{input | after_transaction: input.after_transaction ++ [func]}
+    end
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


lib/ash/subject.ex
```diff
@@ -356,4 +431,232 @@ defmodule Ash.Subject do
...
+  def before_transaction(%Ash.Changeset{} = subject, callback, opts) do
+    Ash.Changeset.before_transaction(subject, callback, opts)
+  end
+
+  def before_transaction(%Ash.Query{} = subject, callback, opts) do
+    Ash.Query.before_transaction(subject, callback, opts)
+  end
+
+  def before_transaction(subject, callback, opts) do
+    if opts[:prepend?] do
```
zachdaniel: We should probably be delegating this to `ActionInput` too. Starting to think we should just be making this a protocol...

zachdaniel: The only reason a protocol doesn't make as much sense to me is that wedon't want people actually implemetning that protocol aside from us.

Either way, these fallback clauses should be updated to delegate to `ActionInput. fun`

chazwatkins:
I removed the default implementations and delegate to the subject type modules.

lib/ash/subject.ex
```diff
@@ -174,32 +166,24 @@ defmodule Ash.Subject do
...
-        Map.get(subject.arguments, to_string(argument))
-      end
-
-    if is_nil(value) do
-      default
-    else
-      value
+  def get_argument(%Ash.Changeset{} = subject, argument, default) do
+    case Ash.Changeset.get_argument(subject, argument) do
+      nil -> default
```
zachdaniel: Actually this isn't right. `get_argument` with a default is different.

```elixir
case fetch_argument(subject, arg) do
  {:ok, v} - v
   :error -> nil
end
```

chazwatkins: I don't understand what you mean.  Are you saying you don't want the default?  The subject type modules `get_argument` don't currently support providing a default, so i just put it in `Ash.Subject`.  Or are you saying you'd prefer `get_argument` to call `mod.fetch_argument`, match on the result, and return the result when there is one and the default when nil?

lib/ash/action_input.ex
```diff
@@ -463,7 +463,7 @@ defmodule Ash.ActionInput do
   - `set_argument/3` for setting argument values
   - `for_action/4` for providing initial arguments
   """
-  @spec fetch_argument(t, atom | String.t()) :: {:ok, term()} | :error
+  @spec fetch_argument(t, atom | binary) :: {:ok, term()} | :error
```
zachdaniel: `String.t()` is the right type for this unless I'm missing something.

lib/ash/subject.ex
zachdaniel: Sorry, yeah my bad, I messed up the ecample. it should return the default on `:error`.

zachdaniel: This is important because if the user provided the arg as `nil`, we shouldn't use the default.

lib/ash/action_input.ex
chazwatkins: I thought I reverted that.  Sorry.

lib/ash/subject.ex
chazwatkins: Good call.  I'll have the `get_argument/2` delegate to the subject type modules.  The `get_argument/3` will use `fetch_argument/2` for the subject type modules and return the default if an error is returned.

chazwatkins:
Ok, It should be good to go now. Let me know if you need anymore changes.

Few items for future PRs:
- Fix issue where you can't call `set_argument` or `set_private_argument` for a `Changeset` or `ActionInput` when you only do `new/1`.  This isn't intended behavior as the error message you receive when using `Ash.Changeset.for_create(MyResource, :create) |> Ash.Changeset.set_private_argument(...)` is that you should do `Ash.Changeset.new(MyResource) |> Ash.Changeset.set_private_argument(...)`, but you can't because `action.arguments` key doesn't exist yet.  This should be an easy fix.
- Add `Ash.Query.set_private_argument`
- Determine the functions that are the same between subjects and pull them out into another module
- Break-up the subject modules into smaller modules to reduce compilation time.

chazwatkins:
@zachdaniel Thanks for the time and patience reviewing the PRs.  I've been off work this week and wanted to contribute a bit.

zachdaniel:
> Determine the functions that are the same between subjects and pull them out into another module

For this one, I think it will be better to leave them in their respective places. We'll never have another subject type, and it is possible their implementations will intentionally drift. Having the unified `Ash.Subject` implementation should be sufficient to simplify callers. Open to doing it differently later but for now I think the separation is best.

zachdaniel (approved):

zachdaniel:

One small note: we want to be careful to avoid scope creep on individual PRs since it can make it a bit tougher to review.

> Thanks for the time and patience reviewing the PRs. I've been off work this week and wanted to contribute a bit.

Not at all!  these are great changes, so thank _you_!

chazwatkins:
> > Determine the functions that are the same between subjects and pull them out into another module
> 
> For this one, I think it will be better to leave them in their respective places. We'll never have another subject type, and it is possible their implementations will intentionally drift. Having the unified `Ash.Subject` implementation should be sufficient to simplify callers. Open to doing it differently later but for now I think the separation is best.

Fair enough.

For the scope creep, I definitely realized this one was getting a pretty big.  I'll keep them smaller in the future to reduce the risk of y'all missing something.

PR #2222, opened 2025-07-24
improvement: Add Ash.Type.Enum label/2, description/2, details/2 callbacks

### LINKED ISSUE #2172: docs: using `gettext` in Ash.Type.Enum is misleading
Using `gettext` as label/description in `Ash.Type.Enum` will always use the default locale, because the options are evaluated at compile time. 

The example gives the impression that the `value/1` and `description/1` functions would return the value based on the current locale at runtime. It would be great if this worked, but I don't see an easy fix, right?

https://github.com/ash-project/ash/blob/f9b6b5cc2511f8726d89a06654c444b5a40caacf/lib/ash/type/enum.ex#L73-L75

chazwatkins:
## Changes

- Provides optional callbacks to allow users to provide options that can be used to transform the label and description at runtime.  i.e. using gettext to translate

closes docs: using `gettext` in Ash.Type.Enum is misleading Fixes #2172

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/type/enum.ex
```diff
@@ -118,6 +191,12 @@ defmodule Ash.Type.Enum do
   @doc "finds the valid value that matches a given input term"
   @callback match(term) :: {:ok, atom} | :error
 
+  @optional_callbacks [
```
zachdaniel: Hm...this is confusing to me. I don't think these are optional callbacks, they are *overridable* callbacks.

chazwatkins: After thinking more about it, it's probably better to revert the callback changes and just remove the reference of `gettext` from the docs.  Users could always define a `label/2`, `description/2`, `details/2` themselves without the need for the optional callbacks and achieve the same thing.

zachdaniel: But the callbacks need to be marked as overridable so that the y can define their own.

zachdaniel: Oh do we not have these callbacks currently? I think its a good addition to have them TBH.

chazwatkins: We have them as callbacks.  Overriding them leaks some implementation details.  For instance, if you override 'details/1' you have to access the '@values_map' module attribute.  So, I opted for the optional 2-arity functions to let them define them if they wanted.

Overriding the 1-arity doesn't provide any value.

I'll add overridable to them though and make sure the implementation doesn't leak through.

chazwatkins:
Ok, this one is good to go.

zachdaniel:
@chazwatkins Sorry to be an ass but now that I see what you mean about this just being a new callback that anyone could define, since we don't actually use it...you were right 😢 Sorry for the whiplash, but your plan of just removing gettext from the docs is the right way.

chazwatkins:
> @chazwatkins Sorry to be an ass but now that I see what you mean about this just being a new callback that anyone could define, since we don't actually use it...you were right 😢 Sorry for the whiplash, but your plan of just removing gettext from the docs is the right way.

No worries, man.  Don't take it as you being an ass at all.  This one didn't take long anyway.

PR #2223, opened 2025-07-24
fix: Change relate_actor to use `force_change_attribute`

chazwatkins:
`relate_actor` uses `manage_relationship/4` when the action is run non-atomically, which requires running a read query on the actor resource.  When users have policies on the actor resource, this means they must add a read policy granting access to the actor for their own record.  The atomic version of the change sets the value directly from the actor map, so this change aligns the non-atomic version to work the same.

- Add feature flag `keep_manage_relationship_for_relate_actor?` that allows users to opt-in to using `force_change_attribute` instead of `manage_relationship` for `relate_actor`

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


documentation/topics/development/backwards-compatibility-config.md
```diff
@@ -7,6 +7,20 @@ will have their defaults changed to the new value.
 
 The ash installer automatically sets all of these.
 
+## keep_manage_relationship_for_relate_actor?
+
+```elixir
+config :ash, keep_manage_relationship_for_relate_actor?: true
```
zachdaniel: No need for this. It was never intentional to require readability to relate the actor, nor was it documented AFAIK.

lib/ash/resource/change/relate_actor.ex
```diff
@@ -62,12 +62,20 @@ defmodule Ash.Resource.Change.RelateActor do
...
 
       actor ->
-        Changeset.manage_relationship(
-          changeset,
-          relationship.name,
-          actor,
-          type: :append_and_remove
-        )
+        if Application.get_env(:ash, :keep_manage_relationship_for_relate_actor?, true) do
+          Changeset.manage_relationship(
```
zachdaniel: We can just add `authorize?: false` to this, there is more happening under the hood in manage_relationship here that we'd want to keep.

chazwatkins:
Changed as requested

zachdaniel:

PR #2226, opened 2025-07-25
Update docs on default pagination behavior.

Torkan:
Adds some more details on default behavior regarding pagination, and the current default config setup.

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/topics/advanced/pagination.livemd
+++ b/documentation/topics/advanced/pagination.livemd
@@ -13,6 +13,10 @@ Ash has built-in support for two kinds of pagination: `offset` and `keyset`. You
 
 Pagination support is configured on a per-action basis. A single action can support both kinds of pagination if desired, but typically you would use one or the other. Read actions generated with `defaults [:read]` support both offset and keyset pagination, for other `read` actions you have to configure the [`pagination` section](https://hexdocs.pm/ash/dsl-ash-resource.html#actions-read-pagination).
 
+> ### Default Pagination Type
+>
+> When an action supports both pagination types, the behavior depends on your application configuration. See the ["Default Pagination Behavior"](#default-pagination-behavior-when-both-types-are-supported) section below for details on how Ash determines which type to use.
+
 > ### Check the updated query return type!
 >
 > Pagination will modify the return type of calling the query action.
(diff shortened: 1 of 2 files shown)


documentation/topics/development/backwards-compatibility-config.md
```diff
@@ -66,14 +66,24 @@ config :ash, default_page_type: :keyset
...
-You can always force returning an `%Ash.Page.Offset{}` by providing the offset option,
-i.e `page: [offset: 0]`
+With this configuration, Ash will now return an `%Ash.Page.Keyset{}` when the pagination
+type is ambiguous (only `limit` is provided).
+
+You can always force a specific pagination type by providing type-specific parameters:
+- Force offset: `page: [limit: 10, offset: 0]` 
+- Force keyset: `page: [limit: 10, after: cursor]`
+
+> **Note**: This configuration is automatically set to `:keyset` when using `mix igniter.install ash`
```
zachdaniel: All of these are set to the "new behavior" by the installer, which is stated above 

> The ash installer automatically sets all of these.

So I think we can remove this section.

documentation/topics/advanced/pagination.livemd
```diff
@@ -895,9 +899,45 @@ last_page.more?
...
+```
+
+- If `config :ash, :default_page_type, :keyset` → returns `Ash.Page.Keyset`
+- If `config :ash, :default_page_type, :offset` → returns `Ash.Page.Offset`
+- If no configuration is set → defaults to `Ash.Page.Offset`
 
-If an action supports both offset and keyset pagination (e.g. default read actions), offset pagination is used by default when page options only contain `limit`. However, the records will have the keyset in the metadata, so keyset pagination can be performed on next pages.
+> #### ⚠️ Configuration Warning
+>
+> Never set `config :ash, :default_page_type, nil` as this will cause a runtime crash.
```
zachdaniel: This warning is a bit specific. I'd suggest trimming down these docs a bit WRT the configuration and instead just link to the backwards compatbility config guide. No need to explain both options here. In this case this reads more like a "feature" rather than a backwards compatibility tool.

documentation/topics/advanced/pagination.livemd
```diff
@@ -895,9 +899,40 @@ last_page.more?
...
+**1. Explicit pagination parameters take precedence:**
+- If `after` or `before` is provided → keyset pagination
+- If `offset` is provided → offset pagination
+
+**2. Configuration-based default for ambiguous cases:**
+When only `limit` is provided, Ash checks your application configuration:
+
+```elixir
+# Check your config/config.exs
+config :ash, :default_page_type, :keyset  # or :offset
```
zachdaniel: What I'm getting at is that for backwards compatible configurations, we shouldn't call this out as a configuration that we want people to set. So at most, there can be a note that a backwards compatibility configuration exists. We don't want to model this as something that the user might set as a "feature". These configuarations will be *removed* in future versions of Ash and only the new behavior will be available. So we really shouldn't explain it here except at most saying something like "note that this behavior is subject to a backwards compatibility configuration: <link>"

Torkan: Ok, so keyset will be the default in the future without it being explicitly being stated in the config, is that what you're saying?

I get that in general in terms of performance characteristics, keyset is preferred over offset, but this basically just explains to the user that unless if you have a config for this setting, currently Ash will default to using offset pagination, which is the current behavior?

The main premise for the PR was to clarify that when you run a query with `page: [limit: 10]` with the default setup, why do I get either an offset or keyset result.

zachdaniel: For new users, when they generate an app, it defaults to keyset and we don't ever want them to change it back, so we don't want to explain really how or why they might want to at all.

zachdaniel (approved):

zachdaniel:

PR #2233, opened 2025-07-27
Feat: Support through for many to many relationship gen

ken-kost:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


ken-kost:
This kinda locks where you should put through modifier, and also the code kind of repeats. I was thinking to switch it a reduce so I could better pattern match in one go but for now this. Didn't want to meddle too much...

lib/mix/tasks/gen/ash.gen.resource.ex
```diff
@@ -416,6 +416,51 @@ if Code.ensure_loaded?(Igniter) do
     defp add_relationships_to_resource(igniter, resource, options) do
       Enum.reduce(options[:relationship] || [], igniter, fn relationship, igniter ->
         case String.split(relationship, ":") do
+          ["many_to_many", name, destination, "through", through_destination | modifiers] ->
```
zachdaniel: Hmm...so this is close, but there is a bit of an issue with it. We don't require that modifiers be ordered in any particular way. This shouldn't be considered a modier, instead it should just be a requirement that you do `many_to_many:comments:MyApp.Blog.PostComment:MyApp.Blog.Comment:...<other_modifiers>`. Then its not a modifer, just an extra positional element.

ken-kost: Makes sanse. So I should just update the docs or...?

ken-kost: oh I see, raising an error when it's not provided

zachdaniel: I don't think we support it yet, so you'd have to add the matcher for many to many relationships and that `through` option. Just saying thats how we should do it instead of `through:Through`

zachdaniel:
@ken-kost mind wrapping this up or closing?

ken-kost:
ごめんなさい先輩 🙇

lib/mix/tasks/gen/ash.gen.resource.ex
```diff
@@ -416,6 +416,63 @@ if Code.ensure_loaded?(Igniter) do
     defp add_relationships_to_resource(igniter, resource, options) do
       Enum.reduce(options[:relationship] || [], igniter, fn relationship, igniter ->
         case String.split(relationship, ":") do
+          ["many_to_many", name, through, destination | modifiers] ->
+            cond do
+              !String.contains?(through, ".") ->
+                raise ArgumentError, """
```
zachdaniel: We don't do this validation for other relationship types, let's not do it here either. It's up to the user really. We can potentially match it as a regex instead if we want, but if we do we should do it everywhere. Technically just `Foo` is a valid module name, so it's not sufficient to check for a `.`. My vote is to leave it out for now.

zachdaniel (approved):

PR #2241, opened 2025-07-31
docs: add info about deadlocks in tests to usage rules

barnabasJ:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


usage-rules.md
```diff
@@ -1131,3 +1131,62 @@ When testing resources:
 - Use `authorize?: false` in tests where authorization is not the focus
 - Write generators using `Ash.Generator`
 - Prefer to use raising versions of functions whenever possible, as opposed to pattern matching
+
+### Preventing Deadlocks in Concurrent Tests with Ash.Generator
+
+When running tests concurrently, using fixed values for identity attributes can cause postgres deadlock errors. Multiple tests attempting to create records with the same unique values will conflict. Ash provides `Ash.Generator` with built-in utilities for generating unique test data.
+
+#### Using Ash.Generator.sequence/3
```
barnabasJ: Since this is unique only on a per-process basis, I'm not entirely sure we should include it here in this form. I mean, all the info is here, but I'm not sure if only having the System.unique_integer might be better. WDYT? :thinking:

zachdaniel: I agree, this will probably make it wrong more often.

zachdaniel: Let's not use the term `postgres` here directly though. We'll want to use data layer agnostic terminology (can use the term deadlock though)

mix.exs
```diff
@@ -406,7 +406,8 @@ defmodule Ash.MixProject do
       {:mix_audit, ">= 0.0.0", only: [:dev, :test], runtime: false},
       {:mix_test_watch, "~> 1.0", only: [:dev, :test], runtime: false},
       {:benchee, "~> 1.1", only: [:dev, :test]},
-      {:tz, "~> 0.28", only: [:test]}
+      {:tz, "~> 0.28", only: [:test]},
+      {:usage_rules, "~> 0.1"}
```
zachdaniel: This shouldn't be added here.

zachdaniel:

PR #2245, opened 2025-08-02
Feat: Add `ash.gen.resources`

ken-kost:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel (commented):
This wouldn't actually work in practice. It only works because you're passing in a single string to the compose task but args come in as a split list. This task would be very unwieldy even if it worked. I'd prefer to explore something like `mix igniter.do` like `mix do'

PR #2249, opened 2025-08-03
filter out invalid changeset, before running the action

barnabasJ:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -487,6 +487,14 @@ defmodule Ash.Actions.Create.Bulk do
       else
         batch
       end
+      |> Enum.reject(fn
+        %{valid?: false} = changeset ->
+          store_error(ref, changeset, opts)
+          true
+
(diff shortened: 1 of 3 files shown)


barnabasJ:
@zachdaniel I'm not sure the tests that fail should work. If the changeset is invalid due to incorrect attributes, should the hooks be executed? Because it's not the action failing, but it's failing beforehand.

lib/ash/actions/create/bulk.ex
```diff
@@ -536,16 +544,6 @@ defmodule Ash.Actions.Create.Bulk do
         end
       )
 
-    batch =
```
zachdaniel: I believe we may need to do this in both places?

zachdaniel:
Hooks should not run on invalid changesets *except* for after transaction hooks.

zachdaniel:

PR #2250, opened 2025-08-04
feat: Add `field?: false` option to Ash.Resource.Calculation DSL

chazwatkins:
Creating as a draft for now

- Add `field?` option to `Ash.Resource.Calculation` DSL
- Exclude `field?: false` calculations from Resource struct and direct loading

This seems to work, but I'd like a review to double check.  Still need to add a validation for `Ash.Error.InvalidLoad` when trying to directly load.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


chazwatkins:
ok, the validation is in and you now get an `InvalidLoad` error when you attempt to load a `field?: false` resource calculation.

zachdaniel:
Just had a thought: we may actually have a bit of an issue here 😢 

Specifically, AshGraphql and AshJsonApi will need to be updated to support this, by not allowing loads for these non-loadable calculations. We'll want PRs for those done before we release this.

chazwatkins:
> Just had a thought: we may actually have a bit of an issue here 😢 
> 
> 
> 
> Specifically, AshGraphql and AshJsonApi will need to be updated to support this, by not allowing loads for these non-loadable calculations. We'll want PRs for those done before we release this.

No experience with those packages, but I'll take a look at them.

chazwatkins:
Both api libraries only expose `public?: true` calculations.  So, I made a small change to have `public_calculations` and `public_calculation` only return back `field?: true` calculations.

Do you see any cases where a `field?: false` calculation should be exposed outside the resource?

zachdaniel:
Hmm...yes. It could be supported in filters exposed over GraphQL for example without being exposed as a field.

PR #2256, opened 2025-08-05
pass actor: nil to ash.get in get_and_lock_for_update

jsw800:
My domain uses `require_actor? true`, which means I'm actually unable to use these changes right now, since they don't pass the actor to `Ash.get`

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


jsw800:
Alternatively, since this uses authorize?: false, I could set actor to nil explicitly - open to feedback!

jsw800:
Actually thinking more about this, I think that's the correct approach, I'll update this to pass `nil` for the actor

zachdaniel (approved):

zachdaniel:

Issue #2252, opened 2025-08-05
Return multiple errors (i.e for all fields) on struct, map, and keyword type casting


zachdaniel:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

these cast differently than most change sets which provide eager errors.

### Describe the solution you'd like

Right now we only return the first error, but we should instead return all errors. We should also move the constraint checking to happen when casting each individual input, instead of in the apply constraints step.

PR #2260, opened 2025-08-06
fix: multiple errors

TwistingTwists:
# Contributor checklist
fixes: #2252

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/composite_type_helpers.ex
+++ b/lib/ash/type/composite_type_helpers.ex
@@ -1,0 +1,135 @@
+defmodule Ash.Type.CompositeTypeHelpers do
+  @moduledoc """
+  Shared utilities for composite type validation and error handling.
+
+  This module provides common functions used by composite types like
+  TypedStruct, Map, Keyword, and Struct for handling validation errors
+  and converting them to consistent formats.
+  """
+
+  @doc """
+  Converts validation errors from composite types to InvalidAttribute exceptions.
+
+  Handles multiple error formats:
+  - Single keyword list errors
+  - Lists of keyword list errors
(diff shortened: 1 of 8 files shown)


TwistingTwists:
@zachdaniel : What is the difference between Invalid and Unknown errors?

one hacky way to fix this would be to change the error type. but that feels not right. 

```
diff --git a/test/typed_struct_test.exs b/test/typed_struct_test.exs
index 0c002e82..d2f6a66e 100644
--- a/test/typed_struct_test.exs
+++ b/test/typed_struct_test.exs
@@ -91,7 +91,7 @@ defmodule Ash.TypedStructTest do
     end
 
     test "new! raises on errors" do
-      assert_raise Ash.Error.Invalid, fn ->
+      assert_raise Ash.Error.Unknown, fn ->
         UserStruct.new!(%{
           id: "123e4567-e89b-12d3-a456-426614174000",
           name: "John Doe",
```

update:  202ffae commit might be useful. though it assumes that if there is a list of errors, we can raise InvalidAttrribute Exception for all of them.

zachdaniel:
@TwistingTwists in this context, yes.

zachdaniel:
This looks good, I'm wondering if this would be a good time to consolidate some of this logic into a central place, like `Ash.Type.CompositeTypeHelpers` or something. What do you think?

TwistingTwists:
> This looks good, I'm wondering if this would be a good time to consolidate some of this logic into a central place, like `Ash.Type.CompositeTypeHelpers` or something. What do you think?

this is a good idea. will do this 

update: ready for your review. relevant commit:  243d283

zachdaniel:

PR #2261, opened 2025-08-07
fix: support top-level `:shared` key in `Ash.Scope.ToOpts` for `Map`

jimsynz:
## Summary

Fixes a regression in `Ash.Scope.ToOpts` implementation for `Map` where top-level `:shared` keys were not being properly converted to context options.

## Problem

The `Ash.Scope.ToOpts` implementation for `Map` was only checking for `:shared` nested under a `:context` key:
```elixir
%{context: %{shared: %{application: "test"}}}  # ✅ Worked
%{shared: %{application: "test"}}              # ❌ Didn't work
```

However, the deprecated `Ash.Context.to_opts/1` documentation indicates it should handle both patterns, and client code was expecting the top-level pattern to work.

## Solution

Updated the `get_context/1` function in `Ash.Scope.ToOpts` for `Map` to check both patterns:
1. First checks for nested `:context` key (existing behavior)
2. Falls back to top-level `:shared` key (new behavior)
3. Maintains backward compatibility and proper priority

## Changes

- **lib/ash/scope.ex**: Updated `Ash.Scope.ToOpts` implementation for `Map`
- **test/scope_test.exs**: Added comprehensive regression tests (7 test cases)

## Testing

- ✅ All existing tests pass (1952 tests, 0 failures)
- ✅ New regression tests cover both patterns and edge cases
- ✅ Verified fix with client reproduction case
- ✅ Backward compatibility maintained

## Contributor Checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/scope.ex
+++ b/lib/ash/scope.ex
@@ -186,11 +186,13 @@ defmodule Ash.Scope do
     def get_actor(map), do: Map.fetch(map, :actor)
     def get_tenant(map), do: Map.fetch(map, :tenant)
 
-    def get_context(map) do
-      with {:ok, context} <- Map.fetch(map, :context) do
-        {:ok, Map.take(context, [:shared])}
-      end
-    end
+    def get_context(%{shared: shared0, context: %{shared: shared1}}),
+      do: {:ok, %{shared: Map.merge(shared1, shared0)}}
+
+    def get_context(%{shared: shared}), do: {:ok, %{shared: shared}}
+    def get_context(%{context: %{shared: shared}}), do: {:ok, %{shared: shared}}
+    def get_context(%{context: _}), do: {:ok, %{}}
+    def get_context(_), do: :error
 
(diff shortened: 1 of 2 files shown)


zachdaniel:
I think this gets pretty confusing. It isn't obvious what should happen if you have `%{context: %{shared: %{a: 1}}, shared: %{a: 2}}`. I'd suggest that we should merge them all up in that scenario, treating the global `shared` key as having priority. Not perfect but avoids "losing" a key.

jimsynz:
> I think this gets pretty confusing. It isn't obvious what should happen if you have `%{context: %{shared: %{a: 1}}, shared: %{a: 2}}`. I'd suggest that we should merge them all up in that scenario, treating the global `shared` key as having priority. Not perfect but avoids "losing" a key.

Agreed. Updated.

zachdaniel:

PR #2265, opened 2025-08-08
improvement: TypedStruct required field pattern matching

### LINKED ISSUE #2253: Modify the function signature of `new!` in `Ash.TypedStruct` in such a way that it will provide dialyzer errors on missing fields
### Code of Conduct

### AI Policy

We can provide better compile time checking of typed struct initialization

We should generate a function head that matches on the required fields, so that you get a dialyzer (and eventually Elixir type system) error at compile time in the generated `new!` function in typed structs.

_No response_

### Additional context

_No response_

chazwatkins:
I think this covers the requirement in #2253.  Let me know if any changes are needed.

### Improvements
- Adds generated function head for `new!/1` for required fields

closes #2253

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/typed_struct.ex
+++ b/lib/ash/typed_struct.ex
@@ -149,6 +149,12 @@ defmodule Ash.TypedStruct do
             end
           end)
 
+        map_required_fields_match =
+          enforce_keys
+          |> Enum.reject(&(&1 in Map.keys(defaults)))
+          |> Enum.map(&{&1, {:_, [], Elixir}})
+          |> then(&{:%{}, [], &1})
+
         map_constraints =
           [
             fields:
@@ -181,11 +187,15 @@ defmodule Ash.TypedStruct do
                constraints: unquote(Macro.escape(map_constraints))
 
              @doc "Create a new #{__MODULE__}, raising any error"
+             def new!(unquote(map_required_fields_match) = fields) do
+               fields
+               |> new()
+               |> Ash.Helpers.unwrap_or_raise!()
+             end
+
              def new!(fields) do
-               case new(fields) do
-                 {:ok, value} -> value
-                 {:error, error} -> raise Ash.Error.to_error_class(error)
-               end
+               {:error, error} = new(fields)
+               raise Ash.Error.to_error_class(error)
              end
 
              @doc "Create a new #{__MODULE__}, or return an error"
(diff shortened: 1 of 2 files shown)


zachdaniel:
So my main concern here is that this will provide a pretty crappy error instead of the current behavior. I think we should consider making a fallback clause that always raises an error. The issue is whether or not that will be picked up but the type system or not since I'm pretty sure this one would 🤔

zachdaniel:
Or maybe this could be opt-in/opt-out when defining the type. `new_pattern?: true` (some better name than that).

chazwatkins:
> So my main concern here is that this will provide a pretty crappy error instead of the current behavior. I think we should consider making a fallback clause that always raises an error. The issue is whether or not that will be picked up but the type system or not since I'm pretty sure this one would 🤔

I thought the same.  Sucks you can't intercept `FunctionClauseError`.  The fallback could be fine for now, but kind of defeats the purpose.  When the type system allows us to write the type expressions, we can update it to generate those and it should _just work_ without any opt-in.

If we're after dialyzer warnings, should we just generate the typespec and eventually migrate that to the type system expressions?  If so, does Spark have anything for generating typespecs?

lib/ash/typed_struct.ex
```diff
@@ -181,11 +187,16 @@ defmodule Ash.TypedStruct do
...
+             end
+
              def new!(fields) do
-               case new(fields) do
-                 {:ok, value} -> value
-                 {:error, error} -> raise Ash.Error.to_error_class(error)
-               end
+               fields
+               |> new()
+               |> Ash.Helpers.unwrap_or_raise!()
```
zachdaniel: I think if we want dialyzer or type system help, we'd need something like

`{:error, error} = new(fields)` and `raise Ash.Error.to_error_class(fields)` so that this fall-through *always* raises.

chazwatkins: Doh! 🙃

chazwatkins:
I made the update you requested.  Let me know if you want any more changes.  Thanks for taking the time to review ☺️

zachdaniel (approved):

zachdaniel:

PR #2269, opened 2025-08-11
[POC] Introduce typed_struct_generator

ekosz:
This is a generator Ash developers can used to generate test versions of their typed structs without need to write manual fixtures.

This is current just a POC and is missing the required documentation and such for this to be ready to merge. I wanted to get feedback from the Ash core contributors if 
1. This is even something that Ash would like as part of its API
2. This feature is even necessary?

To be honest because `typed_structs` are just NewType structs underneath the hood I feel there may be a way to access their generator already, but for the life of me I couldn't get it to work. Not only that I'm currently unaware how one may replace some of the attribute generators with custom ones like this API provides. But if an Ash core contributor could point that out I'd also be happy to close this PR and open a separate one updating the TypedStruct docs to show how one could create generators for them.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/generator/generator.ex
+++ b/lib/ash/generator/generator.ex
@@ -519,6 +519,18 @@ defmodule Ash.Generator do
     end
   end
 
+  def typed_struct_generator(module, opts \\ []) do
+    generators =
+      opts[:defaults]
+      |> Kernel.||([])
+      |> Map.new()
+      |> Map.merge(Map.new(opts[:overrides] || %{}))
+
+    Spark.Dsl.Extension.get_entities(module, [:typed_struct])
+    |> generate_attributes(generators, true, :create, [])
+    |> StreamData.map(fn data -> module.new!(data) end)
+  end
+
   @doc """
   Generate globally unique values.
 
--- a/test/generator/generator_test.exs
+++ b/test/generator/generator_test.exs
@@ -231,6 +231,21 @@ defmodule Ash.Test.GeneratorTest do
     end
   end
 
+  defmodule TypedEvent do
+    use Ash.TypedStruct
+
+    typed_struct do
+      field(:event_source, :atom,
+        allow_nil?: false,
+        constraints: [one_of: [:first_party, :third_pary]]
+      )
+
+      field(:external_event_id, :string, allow_nil?: false, constraints: [min_length: 3])
+      field(:metadata, :map)
+      field(:timestamp, :datetime, allow_nil?: false)
+    end
+  end
+
   defmodule Generator do
     use Ash.Generator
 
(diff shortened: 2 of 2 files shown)


zachdaniel:
🤔 as you say, the builtin generator ought to "just work". Does it not? i.e `Ash.Type.generator(YourType) |> Enum.take(1)`

ekosz:
@zachdaniel Thanks for taking a look!  Unfortunately I tried going down that path and it didn't work. For example:

```iex
iex(1)> defmodule TestStruct do
...(1)>   use Ash.TypedStruct
...(1)>
...(1)>   typed_struct do
...(1)>     field :username, :string, allow_nil?: false
...(1)>   end
...(1)> end
{:module, TestStruct,
 <<70, 79, 82, 49, 0, 0, 135, 176, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 9, 96,
   0, 0, 0, 180, 17, 69, 108, 105, 120, 105, 114, 46, 84, 101, 115, 116, 83,
   116, 114, 117, 99, 116, 8, 95, 95, 105, 110, ...>>,
 [Ash.TypedStruct.Dsl.TypedStruct.Field, :ok]}
iex(2)> Ash.Type.generator(TestStruct) |> Enum.take(1)
** (UndefinedFunctionError) function Ash.Type.generator/1 is undefined or private. Did you mean:

      * generator/2

    (ash 3.5.34) Ash.Type.generator(TestStruct)
    iex:2: (file)
iex(2)> Ash.Type.constraints(TestStruct)
[
  instance_of: [
    type: :atom,
    doc: "The module the struct should be an instance of"
  ],
  preserve_nil_values?: [
    type: :boolean,
    default: false,
    doc: "If set to true, when storing, nil values will be kept. Otherwise, nil values will be omitted."
  ],
  fields: [
    type: :keyword_list,
    keys: [
      *: [
        type: :keyword_list,
        keys: [
          type: [
            type: {:spark_type, Ash.Type, :builtins,
             [{"{:array, inner_type}", "list", "{:array, ${1:inner_type}}"}]},
            required: true
          ],
          allow_nil?: [type: :boolean, default: true],
          description: [type: :string],
          constraints: [type: :keyword_list, default: []]
        ]
      ]
    ],
    doc: "The types of the fields in the struct, and their constraints.\n\nFor example:\n\n    fields:  [\n      amount: [\n        type: :integer,\n        description: \"The amount of the transaction\",\n        constraints: [\n          max: 10\n        ]\n      ],\n      currency: [\n        type: :string,\n        allow_nil?: false,\n        description: \"The currency code of the transaction\",\n        constraints: [\n          max_length: 3\n        ]\n      ]\n    ]\n\nallow_nil? is true by default\n"
  ]
]
```

`Ash.Type.generator` currently takes two arguments, the module and it's constraints. But when you ask the typed_struct for it's constraints you currently get back the constraints of meta fields not the derived ones. I look a looked at how `Ash.Resource.Info.attributes` does this and it looks at the Spark extension to get it's information. That's why I went with that approach for this PR. I don't know too much about Spark and when what information is available when. But if you think something else can be done to fix this and point me in the right direction I'm happy to give it a crack!

zachdaniel:
Ah, so you can get the appropriate constraints with `Ash.Type.init(Type, [])`. Those constraints can then be passed into the generator.

I think regardless we'd make this work using the builtin types generator, even if there is something that needs to be fixed or made more ergonomic there. Thank you for the PR though 🙇

ekosz:
@zachdaniel TIL about `Ash.Type.init(Type, [])` to get the type constraints! Thank you so much

Issue #2276, opened 2025-08-18
Proposal: TypedStruct Info


matt-beanland:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

I've got some resources where an attribute is user defined and I've been using Ash.TypedStruct DSL and also extending it. I'm wanting to allow my users to configure stuff using TypedStruct extensions such as jason and outstanding, and this would be cleaner with some TypedStruct introspection. I

### Describe the solution you'd like

 Ash.TypedStruct.Info with fields method, etc, allowing access to the Ash.TypeStruct.Field struct so TypedStruct extensions can use this.

### Describe alternatives you've considered

In an TypedStruct extension you can just create a TypedStruct with new, and take fields using Map.from_struct |> Map.keys, however the fields aren't ordered in the dsl creation order and don't have the rich information available in existing %Ash.TypedStruct.Field{}

matt-beanland:
I've raised a PR https://github.com/ash-project/ash/pull/2277

matt-beanland:
Thanks for merging this. I have made a minor mistake with the typespec for Ash.TypedStruct.Info.field_names/1, the return type should be list(atom) not MapSet.t(). I've fixed this with PR https://github.com/ash-project/ash/pull/2286, again there are some dialyzer issues that are unrelated.

PR #2277, opened 2025-08-18
typed_struct_info

matt-beanland:
# Contributor checklist

- [ *] AI was not used in the creation of this PR.
- [ ] Chores
- [* ] Documentation changes
- [* ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies


matt-beanland:
There are some dialyzer issues which I believe are on main. The related issue is https://github.com/ash-project/ash/issues/2276

lib/ash/typed_struct/info.ex
```diff
@@ -0,0 +1,56 @@
...
+  @moduledoc "Introspection for typed structs"
+
+  alias Spark.Dsl.Extension
+
+  @doc """
+  The trace_name of the typed struct
+  """
+  @spec trace_name(Spark.Dsl.t() | Ash.Type.NewType.t()) :: String.t() | nil
+  def trace_name(typed_struct) do
+    Extension.get_opt(typed_struct, [:typed_struct], :trace_name, nil) ||
```
zachdaniel: I don't think that typed structure supports configuring a trace name does it?

zachdaniel: or a short name, for example.

matt-beanland: No it doesn't. Can remove these or just have them set to the default from Module.split. What would you like me to do?

zachdaniel: We should remove anything that doesn't explicitly map to the DSL for typed structs.

matt-beanland: I pushed a change to the branch adding a test for trace_name. While neither short_name or trace_name can be overridden they are potentially useful in extensions which won't break if we do add the ability to set these (and plural_name), which I haven't added to Info as no logical default.

zachdaniel: 🤔 I think let's leave them out at least of this PR at least because I don't know of any current justification for their existence. We can always open a separate PR to discuss them that won't block the addition of the other stuff 😄

matt-beanland: Thanks, have taken out short_name and trace_name.

zachdaniel (approved):

zachdaniel:

PR #2281, opened 2025-08-19
chore: add test for nested calculation with a non-nillable argument used by an expression that guards for nils

skanderm:
Adds a test for using a calculation with a non-nillable argument in another calculation with a check for nils


zachdaniel:
Fixed in `main`

PR #2285, opened 2025-08-21
fix: allow module validations in generic actions

skanderm:
Module validations are described in the documentation for generic actions but were missing `Ash.ActionInput` in `supports/1`. https://hexdocs.pm/ash/generic-actions.html#validations

image

# Contributor checklist


zachdaniel:
It's not that module validations are missing, you need to put in your validation what it supports. That's for you to define on your end.

zachdaniel:
Like by default we assume the validation was only written with change sets in mind.

skanderm:
Oh oops, I missed that it was a callback!

Issue #2274, opened 2025-08-15
Using `exists()` in policy with syntax error produces very obtuse stack trace


olivermt:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

main

### Operating system

All

### Current Behavior

Broken syntax inside `exists()` like for example:

`exists(foo.bar == "bar")` as opposed to correct syntax `exists(foo, bar == "bar")` gives an obtuse stack trace that gives no information about whats wrong:

```[error] 082eebac-545b-48b4-bb58-2671c09ef121: Exception raised while resolving query.

** (Ash.Error.Unknown) 
Bread Crumbs:
  > Exception raised in: Safari.Outcrop.PaleoMap.read
  > Exception raised in: Safari.Outcrop.Outcrop.read

Unknown Error

* ** (ArgumentError) `nil` is not a Spark DSL module.

  nil.persisted(:data_layer)
  (spark 2.2.67) lib/spark/dsl/extension.ex:138: Spark.Dsl.Extension.persisted!/3
  (ash 3.5.25) lib/ash/filter/filter.ex:3881: Ash.Filter.do_hydrate_refs/2
  (ash 3.5.25) lib/ash/filter/filter.ex:3763: Ash.Filter.do_hydrate_refs/2
  (safari 0.1.0) lib/safari/policy/end_user_read_access.ex:2: Safari.Policy.EndUserReadAccess.try_strict_check/3
  (ash 3.5.25) lib/ash/policy/policy.ex:183: Ash.Policy.Policy.fetch_or_strict_check_fact/2
  (ash 3.5.25) lib/ash/policy/policy.ex:427: Ash.Policy.Policy.handle_constants/2
  (ash 3.5.25) lib/ash/policy/policy.ex:390: Ash.Policy.Policy.handle_constants/2
  (ash 3.5.25) lib/ash/policy/policy.ex:402: Ash.Policy.Policy.handle_constants/2
  (ash 3.5.25) lib/ash/policy/policy.ex:390: Ash.Policy.Policy.handle_constants/2
  (ash 3.5.25) lib/ash/policy/policy.ex:29: Ash.Policy.Policy.solve/1
  (ash 3.5.25) lib/ash/policy/checker.ex:65: Ash.Policy.Checker.strict_check_scenarios/1
[...]

PR #2291, opened 2025-08-24
enhancement: exists expressions verifier

ken-kost:
#2274 
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
We don't need to special case exists expressions, I think we could just use `Ash.Filter.hydrate_refs` and if it returns an error we raise that.

PR #2292, opened 2025-08-25
Fix bug when casting a non-existing atom from datalayer when `unsafe_to_atom?` is set

Rutgerdj:
The `unsafe_to_atom?` constraint was not used by the `Ash.Type.Atom.cast_stored/2` function so when casting a stored string to an atom it would throw errors. 

I've added a test that demonstrates the problem. The test uses the ETS directly to insert a non-existing atom. I've put this in a new test module specifically for Atom type tests.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/atom.ex
+++ b/lib/ash/type/atom.ex
@@ -81,17 +81,7 @@ defmodule Ash.Type.Atom do
 
   def cast_input("", _), do: {:ok, nil}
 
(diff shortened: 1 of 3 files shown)


zachdaniel:

PR #2293, opened 2025-08-25
fix: handle notification without set domain in telemetry spans

vonagam:
https://github.com/ash-project/ash/commit/27aca515eb53f2ecb2dfc6d2c98f133a57e6ce71 added telemetry spans and uses `notification.domain` for span names but based on previous state, `Notification.new` signature and [documentation](https://github.com/ash-project/ash/blob/d10bd592905702438bc2536f1691fffbd424f065/lib/ash/action_input.ex#L933-L937) in `ActionInput` I would assume that `resource` is the only required field for `Ash.Notification`.

So this adds fallback to `Ash.Resource.Info.domain` if `notification.domain` is not set.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/notifier/notifier.ex
+++ b/lib/ash/notifier/notifier.ex
@@ -57,19 +57,14 @@ defmodule Ash.Notifier do
 
   defp do_notify(notifier, notification) do
     tracer = notification.changeset && notification.changeset.context[:private][:tracer]
+    domain = notification.domain || Ash.Resource.Info.domain(notification.resource)
 
     Ash.Tracer.span :notification,
-                    fn ->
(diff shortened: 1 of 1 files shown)


zachdaniel:

vonagam:
Ah, just force pushed a test...

Doesn't matter now, just wasn't sure if "Bug fixes include regression tests" means that all bug fixes should include tests now.

zachdaniel:
Ideally yes, but you can open a PR or not worry about it, your call :)

PR #2297, opened 2025-08-28
fix: Allow ^actor() in upsert_condition

### LINKED ISSUE #1445: actor not available to upsert_condition
When you refer to an actor in upsert condition with the following: 
    upsert_condition expr(is_nil(^actor(:id)))

This produces the following error even if the actor is present on the query:
{:error, %Ash.Error.Invalid{changeset: "#Changeset<>",  errors: [%Ash.Error.Changes.ActionRequiresActor{splode: Ash.Error, bread_crumbs: [], vars: [], path: [], stacktrace: #Splode.Stacktrace<>, class: :invalid}]}}

FugiTech:
[The docs for upserts](https://hexdocs.pm/ash/create-actions.html#upserts) say to use `upsert_condition expr(user_id == ^actor(:id))` to restrict access to updating resources to the owner of the resource, as the create policy will be used and not the update policy when an upsert occurs.

When I attempted to do something similar I would consistently get an `Ash.Error.Changes.ActionRequiresActor` error, despite always passing in an `actor`. Upon digging in, the issue is that `filter` attempts to expand the `upsert_condition` expression and replace the `{:_actor, field}` placeholders immediately. However `Ash.Changeset.for_create` isn't populating the context with `actor`, I suspect because `do_for_action` is responsible for that logic.

I was unsure the best approach and opted for the simplest fix of adding `actor` to the context passed in to `filter`. 

This fixes #1445

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1896,10 +1896,6 @@ defmodule Ash.Changeset do
         upsert_condition: upsert_condition
       }
     })
-    |> then(fn
-      changeset when upsert_condition != nil -> filter(changeset, upsert_condition)
-      changeset -> changeset
-    end)
     |> do_for_action(action, params, opts)
   end
 
@@ -2541,6 +2537,7 @@ defmodule Ash.Changeset do
                 changeset
                 |> prepare_changeset_for_action(action, opts)
                 |> handle_params(action, params, opts)
+                |> handle_upsert()
                 |> run_action_changes(
                   action,
                   opts[:actor],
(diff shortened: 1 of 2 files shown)


zachdaniel:
🤔 I see where you're going with it but I don't want to do the scope -> opts work multiple times if it can be avoided.

It seems like somewhere we are doing logic out of order or not setting the new private context before exapnding the upsert condition. Could you look to see if we can fix it that way?

FugiTech:
Understood, I originally chose to duplicate the work as I was afraid of touching `do_for_action`.

I've updated the PR to instead move the `filter` call inside `do_for_action` using a new helper, `handle_upsert`, that checks for the `upsert_condition` in the private context. I _believe_ this better matches the intended flow, and avoids duplicate work.

zachdaniel:

Issue #2298, opened 2025-08-28
Paginated query with `count: true` should skip the count query if the non-count query was skipped due to policy


drtheuns:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Given a query such as:

```
query = Ash.Query.page(query, limit: x, offset: y, count: true)
Ash.read(query)
```

Then if the query is cancelled due to a policy issue, then the count query should also be skipped:

### Operating system

Linux

### Current Behavior

```
[debug] App.Accounts.User.read: skipped query run due to filter being false"

[debug] QUERY OK source="users" db=0.4ms queue=15.4ms idle=1335.0ms
SELECT coalesce(count(*), $1::bigint)::bigint FROM "users" AS u0 WHERE (FALSE) [0]
```

### Expected Behavior

If the normal query is skipped, then the count query is also skipped.

zachdaniel:
Thanks for the issue, PRs welcome to fix!

StephanH90:
This issue can be closed.

PR #2303, opened 2025-08-30
skip count of records when paginating if filter evaluates to false

rtorresware:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

I was not able to test that the count query was not made. I tried: 
- shutting down the ETS layer: action fails due to table not found
- checking telemetry events for read: no changes between code with fix / without fix in the read events
- capture logs: debug events are not logged because of global logger settings.

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -3376,6 +3376,15 @@ defmodule Ash.Actions.Read do
            (page[:count] != false and action.pagination.countable == :by_default))
 
     cond do
+      query.filter && query.filter.expression == false ->
+        result = if needs_count?, do: 0, else: nil
+
+        if return? do
+          {:ok, {:ok, result}}
+        else
+          {:ok, fn -> {:ok, result} end}
+        end
+
       Map.has_key?(query.context, :accessing_from) and needs_count? ->
         if return? do
           # Relationship count is fetched by the parent using aggregates, just return nil here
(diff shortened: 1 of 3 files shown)


rtorresware:
solves https://github.com/ash-project/ash/issues/2298

lib/ash/actions/read/read.ex
```diff
@@ -3376,6 +3376,17 @@ defmodule Ash.Actions.Read do
            (page[:count] != false and action.pagination.countable == :by_default))
 
     cond do
+      query.filter && query.filter.expression == false ->
+        Logger.debug("""
+        #{inspect(query.resource)}.#{query.action.name}: skipped count query run due to filter being false"
+        """)
+
+        if return? do
+          {:ok, {:ok, nil}}
```
zachdaniel: if `needs_count?` then this should return `0`, and `nil` otherwise.

lib/ash/actions/read/read.ex
```diff
@@ -3376,6 +3376,19 @@ defmodule Ash.Actions.Read do
            (page[:count] != false and action.pagination.countable == :by_default))
 
     cond do
+      query.filter && query.filter.expression == false ->
+        Logger.debug("""
```
zachdaniel: Since we log the main query being skipped, no need to log here also.

zachdaniel:

PR #2305, opened 2025-09-01
fix: add default values to manual action changesets

Torkan:
Ensure that changesets for manual update actions get default values applied before handing up to manual update module's `update/3`

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/update/update.ex
+++ b/lib/ash/actions/update/update.ex
@@ -480,12 +480,8 @@ defmodule Ash.Actions.Update do
                     if result = changeset.context[:private][:action_result] do
                       result
                     else
-                      if Enum.empty?(changeset.attributes) &&
-                           Ash.DataLayer.data_layer_can?(changeset.resource, :atomic_update) do
-                        Ash.Changeset.atomic_defaults(changeset)
-                      else
-                        Ash.Changeset.set_defaults(changeset, :update, true)
-                      end
+                      changeset
+                      |> Ash.Changeset.set_defaults(:update, true)
                       |> mod.update(
                         action_opts,
                         %Ash.Resource.ManualUpdate.Context{
--- a/lib/ash/actions/update/update.ex
+++ b/lib/ash/actions/update/update.ex
@@ -480,8 +480,13 @@ defmodule Ash.Actions.Update do
                     if result = changeset.context[:private][:action_result] do
                       result
                     else
-                      mod.update(
-                        changeset,
+                      if Enum.empty?(changeset.attributes) &&
+                           Ash.DataLayer.data_layer_can?(changeset.resource, :atomic_update) do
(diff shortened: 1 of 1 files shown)


lib/ash/actions/update/update.ex
```diff
@@ -480,8 +480,13 @@ defmodule Ash.Actions.Update do
                     if result = changeset.context[:private][:action_result] do
                       result
                     else
-                      mod.update(
-                        changeset,
+                      if Enum.empty?(changeset.attributes) &&
+                           Ash.DataLayer.data_layer_can?(changeset.resource, :atomic_update) do
```
zachdaniel: The assumption in this logic branch is currently that the update is not atomic, so I think we only need the set_defaults logic.

Torkan: Hm, are you sure?

The match pattern above does a check against the `can_atomic_update?` value, but not this one.

Also, the same pattern is used below in the else branch (if it's not a manual action).

zachdaniel: Right, it makes sense to do that if its not a manual action, but not if it is. Which match pattern above are you referring to?

Torkan: I was referring to this:

```
fn
          %{atomics: atomics} when atomics != [] and not can_atomic_update? ->
            {:error,
             Ash.Error.Invalid.AtomicsNotSupported.exception(
               resource: changeset.resource,
               action_type: :update
             )}
```

Since there are no later checks using `can_atomic_update?`, I assumed it meant that "from here on out atomics are supported" 😅 

But okay, I assume you are saying that that condition can't ever occur for manual update actions then, which is good enough for me 😄

Torkan: Slightly unrelated, but worth double-checking:

Currently we have:
```
can_atomic_update? =
      Ash.DataLayer.data_layer_can?(changeset.resource, {:atomic, :update})
```

And further down when handling regular update actions:

```
changeset =
                          if Enum.empty?(changeset.attributes) &&
                               Ash.DataLayer.data_layer_can?(changeset.resource, :atomic_update) do
                            Ash.Changeset.atomic_defaults(changeset)
                          else
                            Ash.Changeset.set_defaults(changeset, :update, true)
                          end
```

Is `:atomic_update` an actual thing, or is it a typo?
I couldn't find any reference to it in `ash_postgres` for example, I could only find references to `{:atomic, :update}`

zachdaniel: 😓 ah, yeah so one is "can you support setting fields atomically" and the other is "can you support fully atomic atomic updates".

zachdaniel (approved):

PR #2310, opened 2025-09-08
Runtime upsert_fields option is not passed to data layer in certain cases

jlgeering:
following up on https://elixirforum.com/t/anybody-working-on-or-interested-in-ash-mongo-ash-datalayer-for-mongodb/72189/4

when creating a resource directly without creating a changeset first, upsert_fields is not passed to the datalayer:  (new test case in commit 72b161d)

```
Ash.create!(
  Post,
  %{
   title: "foo",
   contents: "test"
  },
  upsert?: true,
  upsert_identity: :unique_title,
  upsert_fields: [:contents] # this gets lost
)
```

the fix in commit c7e6376 is my best guess on how to fix it (and all tests pass) but:
- I am not 100% confident that this is the right place or way to do it
- there might be a different helper that I haven't seen that should be called

=> i.e. I understand what I am doing, but not why upsert_fields need to be in the changeset, when they are already in the opts
=> I would appreciate a review here.

Next, I have updated an existing test to make the behaviour of "reverts to default" for fields missing in the changeset more explicit. I can remove that if not appropriate.

Finally, I have a doubt about `upsert?: true` in

```
# lib/ash/actions/create/create.ex 133:140
    changeset =
      Ash.Changeset.set_context(changeset, %{
        private: %{
          upsert?: true,
          upsert_identity: upsert_identity,
          upsert_fields: upsert_fields
        }
      })
```

I have tried to remove it, set it to false, or to `upsert?: opts[:upsert?]` (which my instincts tell me it should be) and all tests still pass. I have not yet been able to write a failing test case.

=> not sure what to do ... maybe that should be a separate issue?

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/create/create.ex
+++ b/lib/ash/actions/create/create.ex
@@ -111,6 +111,13 @@ defmodule Ash.Actions.Create do
         opts[:upsert_identity] || get_in(changeset.context, [:private, :upsert_identity])
       end
 
+    upsert_fields =
+      Ash.Changeset.expand_upsert_fields(
+        opts[:upsert_fields] || get_in(changeset.context, [:private, :upsert_fields]) ||
+          action.upsert_fields,
+        changeset.resource
+      )
+
     opts =
       if get_in(changeset.context, [:private, :return_skipped_upsert?]) do
         Keyword.put(opts, :return_skipped_upsert?, true)
@@ -123,7 +130,11 @@ defmodule Ash.Actions.Create do
 
     changeset =
       Ash.Changeset.set_context(changeset, %{
-        private: %{upsert?: true, upsert_identity: upsert_identity}
+        private: %{
+          upsert?: true,
+          upsert_identity: upsert_identity,
+          upsert_fields: upsert_fields
+        }
       })
 
     with %{valid?: true} = changeset <- changeset(changeset, domain, action, opts),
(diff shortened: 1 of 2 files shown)


jlgeering:
I do not understand the failed ci check (file not formatted) => I do not see the line it is complaining about

zachdaniel:
Its a line you didn't touch, I pushed some code without formatting recently 🥶.

lib/ash/actions/create/create.ex
```diff
@@ -111,6 +111,15 @@ defmodule Ash.Actions.Create do
         opts[:upsert_identity] || get_in(changeset.context, [:private, :upsert_identity])
       end
 
+    upsert_fields =
+      if opts[:upsert?] do
```
zachdaniel: I don't think we need the conditional do we? Should just be able to always use:

```elixir
opts[:upsert_fields] || get_in(changeset.context, [:private, :upsert_fields]) ||
  action.upsert_fields
```

jlgeering: ok, seems reasonable. I changed it, tests still pass. (will push)

zachdaniel:

PR #2311, opened 2025-09-08
fix(nif rustler_precompiled): update targets to ubuntu-22.04 and extend release list to prevent stuck builds

shahryarjb:
After researching I found some targets like `ubuntu-20.04` has problem and i think kinda deprecated and some another release are not in our list
 
- Based on https://github.com/tessi/wasmex/blob/main/.github/workflows/release.yml

I tested the list inside my test repo and works! and I hope fixes the IgniterJs build

Thank you in advance


zachdaniel:

PR #2312, opened 2025-09-08
build(ci): remove unnecessary targets from `rustler_precompiled` release

shahryarjb:
I removed some **unnecessary** targets for `rustler_precompiled`
Since the Rust part is currently only used within the Elixir library, these targets are unnecessary. Removing them reduces resource usage, and fixing their related errors would require significant changes across multiple parts of the codebase.

Thank you in advance


zachdaniel:

PR #2315, opened 2025-09-09
Update behaviour in preparation.ex

m0rt3nlund:
Since `init/1` and `supports/1` have default implementation they can be optional callbacks

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/preparation/preparation.ex
+++ b/lib/ash/resource/preparation/preparation.ex
@@ -109,7 +109,7 @@ defmodule Ash.Resource.Preparation do
               Ash.Query.t() | Ash.ActionInput.t()
   @callback supports(opts :: Keyword.t()) :: [module()]
 
-  @optional_callbacks init:1, supports:1
+  @optional_callbacks init: 1, supports: 1
 
   defmacro __using__(_) do
     quote do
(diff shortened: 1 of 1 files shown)


zachdaniel:
That's not really how it works actually. The behavior requires these implementations. The `use` macro adds them, but they are still required.

m0rt3nlund:
Then I am a bit confused, but thats okay!
So the default implementation is just a fallback not to crash when not implemented?

The reason I thought this was optional was that the docs at https://hexdocs.pm/ash/Ash.Resource.Preparation.html says that `init` is optional and since `supports` had a default implementation I assumed the same for this.

zachdaniel:
Well, any given module that adopts that behavior must define that function. But the default implementation ensures it is always defined when you `use Ash.Resource.Preparation`. So its optional for it to be defined manually, but not optional to actually exist.

m0rt3nlund:
Okay! I assumed the same philosophy that for example is used for `GenServer`.
Here `handle_info` callback is defined as optional even if it has a default implementation that is overridable.

But for `Ash.Resource.Preparation` we dont want to assume that the developer will only implement a `prepare` function that only supports the default `[Ash.Query]` as returned by the default implementation of `supports`?

zachdaniel:
> Here handle_info callback is defined as optional even if it has a default implementation that is overridable.

There is no default definition for handle_info 😁

PR #2320, opened 2025-09-13
fix: handle validation init errors without overriding them with validation message

### LINKED ISSUE #2319: Mesage in validatoin overrides internal option validation error message
### Code of Conduct

### AI Policy

### Versions

ash 3.5.40

### Operating system

Mac

### Current Behavior

The following validation

```elixir
      validations do
        validate present([:first_name, :last_name], where: [absent(:full_name)]) do
          message "must have either first_name and last_name or full_name"
        end
      end
```
returns the error "must have either first_name and last_name or full_name"

### Reproduction

The attached diff adds a test to reproduce the problem

[validation_changes.patch](https://github.com/user-attachments/files/22311484/validation_changes.patch)

### Expected Behavior

The error message should be 
`unknown options [:where], valid options are: [:at_least, :at_most, :exactly, :attributes]`

Because the option validation error is overridden by the message, it was quite hard to find the actual problem.

marot:
Fixes #2319

I was thinking about trying to check if options are valid earlier, but saw a commit message that the option initialization was moved to runtime intentionally.

Still, it would be great for DX if such a mistake is pointed out earlier. 

The function is quite nested. Please let me know if another approach to handle the init error is preferred. The first case match is on error, to improve readability.

Edit: Force pushed to fix `mix check` that I forgot to run

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -3858,54 +3858,56 @@ defmodule Ash.Changeset do
               changeset: changeset
             )
 
-          with {:ok, opts} <- validation.module.init(opts),
-               :ok <-
-                 Ash.Resource.Validation.validate(
-                   validation.module,
-                   changeset,
-                   opts,
-                   struct(
-                     Ash.Resource.Validation.Context,
-                     Map.put(context, :message, validation.message)
-                   )
-                 ) do
-            changeset
-          else
-            :ok ->
-              changeset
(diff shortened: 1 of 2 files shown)


zachdaniel:

zachdaniel:

PR #2324, opened 2025-09-16
feat: implemented 'has' and 'intersects' functions

moutikabdessabour:
Initial implementation of has and overlaps operators to allow for intersection and inclusion checks against array fields

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -25,7 +25,9 @@ defmodule Ash.Filter do
     Fragment,
     FromNow,
     GetPath,
+    Has,
     If,
+    Intersects,
     IsNil,
     Lazy,
     Length,
@@ -41,9 +43,7 @@ defmodule Ash.Filter do
     StringSplit,
     StringTrim,
     Today,
-    Type,
-    Has,
-    Intersects
+    Type
   }
 
   alias Ash.Query.Operator.{
@@ -72,8 +72,10 @@ defmodule Ash.Filter do
     Fragment,
     FromNow,
     GetPath,
+    Has,
     IsNil,
     If,
+    Intersects,
     Lazy,
     Length,
     Minus,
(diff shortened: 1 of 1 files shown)


zachdaniel:
Actually this should probably be done as a function not an operator.

moutikabdessabour:
@zachdaniel can you expand on that?

zachdaniel:
These should be `use Ash.Query.Function` instead of `Ash.Query.Operator` because `has` and `intersects` are not valid Elixir operators and so these are `has(a, b)` not `a has b`. We also need to test these in expressions 😄

moutikabdessabour:
@zachdaniel I don't need to add ETS tests, right?

I saw that only the operators are tested which makes sense

zachdaniel:
Yeah we need tests showing the actual filter working, which for core would be against ets backed resources.

moutikabdessabour:
@zachdaniel The way I understood the ets data layer is that it falls back to the evaluate function. i.e if I implement the tests in filter/filter_test.exs and query/function/{has,intersects}_test.exs they should be enough.

moutikabdessabour:
@zachdaniel I added some rudementary tests to the `ets_test.exs`, let me know if this needs improvements.

zachdaniel:
Looks good, please resolve any warnings and pull `main`.

moutikabdessabour:
@zachdaniel I resolved the credo warning, but the dialyzer ones exist in main, 2 of them should be reported to dialyxir, and there is one that's valid

for reference:
```
Please file a bug in https://github.com/jeremyjh/dialyxir/issues with this message.

Unknown error occurred: %FunctionClauseError{module: Dialyxir.Warnings.CallWithoutOpaque, function: :format_long, arity: 1, kind: nil, args: nil, clauses: nil}
```

from a first look on the valid issue `Ash.ActionInput.new` to_tenant is missing from the typing. Done this but dialyzer still complains. fyi  I am on OTP 28

zachdaniel:

moutikabdessabour:
@zachdaniel what are the next steps, to make this available in ash_sql

zachdaniel:
You'd need to modify `expr.ex` in `ash_sql` to have an implementation for them 😄

Issue #2326, opened 2025-09-17
Ash.Notifier requires_original_data? action type discrepancy in verifier


meeq:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Elixir 1.18.4
Erlang 27.3.4.2
Ash 3.5.40

### Operating system

macOS

### Current Behavior

Ash.Notifier specifies that the `requires_original_data?` callback takes Resource module and an Action struct:

https://github.com/ash-project/ash/blob/9dd40357ed479caa017abce7d00fbf8aadab42ac/lib/ash/notifier/notifier.ex#L1-L6

https://github.com/ash-project/ash/blob/9dd40357ed479caa017abce7d00fbf8aadab42ac/lib/ash/resource/actions/actions.ex#L1-L7

However, verify_actions_atomic calls `requires_original_data?` with the action name:

https://github.com/ash-project/ash/blob/9dd40357ed479caa017abce7d00fbf8aadab42ac/lib/ash/resource/verifiers/verify_actions_atomic.ex#L71-L75

### Reproduction

This leads to compilation errors when implementing an Ash.Notifier that tries to access fields on the action argument:

```elixir
defmodule ModusLedger.Notifiers.CoachSlack do
  use Ash.Notifier

  alias ModusLedger.Clients.ClientInvoice

  @impl Ash.Notifier
  def requires_original_data?(ClientInvoice, action) do
    action.name in [:create, :update, :upsert]
  end

  # ...
end
```

```
warning: using map.field notation (without parentheses) to invoke function :update.name() is deprecated, you must add parentheses instead: remote.function()
  lib/modus_ledger/notifiers/coach_slack.ex:13: ModusLedger.Notifiers.CoachSlack.requires_original_data?/2
  (elixir 1.18.4) lib/enum.ex:4390: Enum.filter_list/2
  (ash 3.5.40) lib/ash/resource/verifiers/verify_actions_atomic.ex:73: anonymous fn/3 in Ash.Resource.Verifiers.VerifyActionsAtomic.verify/1
  (elixir 1.18.4) lib/enum.ex:4442: Enum.flat_map_list/2
  (ash 3.5.40) lib/ash/resource/verifiers/verify_actions_atomic.ex:18: Ash.Resource.Verifiers.VerifyActionsAtomic.verify/1

[...]

PR #2327, opened 2025-09-17
Fix usage of Ash.Notifier.requires_original_data/2 in VerifyActionsAtomic

meeq:
Fixes https://github.com/ash-project/ash/issues/2326

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/verifiers/verify_actions_atomic.ex
+++ b/lib/ash/resource/verifiers/verify_actions_atomic.ex
@@ -71,7 +71,7 @@ defmodule Ash.Resource.Verifiers.VerifyActionsAtomic do
           dsl
           |> Ash.Resource.Info.notifiers()
           |> Enum.filter(fn notifier ->
-            notifier.requires_original_data?(module, action.name)
+            notifier.requires_original_data?(module, action)
           end)
           |> case do
             [] ->


lib/ash/resource/verifiers/verify_actions_atomic.ex
```diff
@@ -71,7 +71,7 @@ defmodule Ash.Resource.Verifiers.VerifyActionsAtomic do
           dsl
           |> Ash.Resource.Info.notifiers()
           |> Enum.filter(fn notifier ->
-            notifier.requires_original_data?(module, action.name)
+            notifier.requires_original_data?(module, action)
```
zachdaniel, line 74: 🤔 We need to keep this the same but change the type spec to match the implementation, otherwise this is a breaking change.

meeq: To be clear, the type spec for `requires_original_data?/2` matches the usage in this fix, which is consistent with the implementation in `Ash.Notifier.PubSub` and [usage of the callback in `Ash.Changeset`](https://github.com/ash-project/ash/blob/9dd40357ed479caa017abce7d00fbf8aadab42ac/lib/ash/changeset/changeset.ex#L1008-L1012). The second argument should be a struct, not an atom:

https://github.com/ash-project/ash/blob/9dd40357ed479caa017abce7d00fbf8aadab42ac/lib/ash/notifier/notifier.ex#L1-L6

https://github.com/ash-project/ash/blob/9dd40357ed479caa017abce7d00fbf8aadab42ac/lib/ash/resource/actions/actions.ex#L1-L7

`Ash.Notifier.PubSub` wasn't triggering a compilation error because it was falling-through a match, but it was returning a spurious `false` result:

https://github.com/ash-project/ash/blob/9dd40357ed479caa017abce7d00fbf8aadab42ac/lib/ash/notifier/pub_sub/pub_sub.ex#L237-L240

https://github.com/ash-project/ash/blob/9dd40357ed479caa017abce7d00fbf8aadab42ac/lib/ash/notifier/pub_sub/pub_sub.ex#L532-L538

zachdaniel: Thanks for clarifying.

zachdaniel:

PR #2336, opened 2025-09-23
Attempt at adding bulk_create for Ash.DataLayer.Mnesia

zentourist:
# Contributor checklist

- [ ] Chores
- [x] Documentation changes (@doc for bulk_create/3)
- [ ] Refactoring
- [ ] Update dependencies

This is my attempt add adding `Ash.DataLayer.Mnesia.bulk_create/3`. I'm still wrapping my head around a lot of it and the `upsert` portion is heavily inspired by the ETS DataLayer. There is also a deficiency in the `upsert` where when you pass in `upsert_fields`, it happily ignores it. This was the current behavior, but something I noticed when I started adding some Unit tests around the code. I'm happy to keep hacking at this, but wanted some early feedback to make sure I'm on the right path. ✌️ 

Dan


lib/ash/data_layer/mnesia/mnesia.ex
```diff
@@ -317,9 +318,114 @@ defmodule Ash.DataLayer.Mnesia do
...
+            {:halt, {:error, error}}
+        end
+      end)
+    end)
+  end
+
   @doc false
   @impl true
-  def create(resource, changeset) do
+  def create(resource, changeset, with_transaction \\ true) do
```
zachdaniel: lets make this an opts list.

zachdaniel:
Looks great so far, would love to have that `upsert_fields` bug fixed 😄

zentourist:
> Looks great so far, would love to have that `upsert_fields` bug fixed 😄

Thanks for the feedback @zachdaniel. I'll have a look at the upsert bug in a separate PR. I was thinking about doing some work around the upsert as well to make that more performant so we're not loading everything in memory. I don't think Mnesia has a great way to handle all of the upsert features natively but I'm thinking over some approaches that should work well.

lib/ash/data_layer/mnesia/mnesia.ex
```diff
@@ -69,7 +69,7 @@ defmodule Ash.DataLayer.Mnesia do
       :filter,
       :limit,
       :tenant,
-      :sort,
+      sort: [],
```
zentourist, line 72: @zachdaniel - this fixed the upsert issue I was having. I was also not sending in the data correctly in the test for upsert because I wasn't sending the `:id`. 🤦‍♂️

test/ash/data_layer/mnesia_test.exs
```diff
@@ -0,0 +1,176 @@
...
+                   roles: [:admin, :manager]
+                 }
+               ],
+               notifications: nil,
+               error_count: 0
+             } = resp
+    end
+
+    test "bulk_create/3 with UPSERT without returning records" do
+      user_a =
```
zentourist, line 130: This fixes the crux of the issue I was having on upsert by including the `:id` in the attributes.

zachdaniel:

PR #2340, opened 2025-09-29
Fix with_transaction bug in Mnesia data layer

zentourist:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

When I moved the `with_transaction` in the Mnesia DataLayer to kw opts, I did not add a default value. It was not caught in the test because `Ash.create`, must be wrapping it already. I added a default value and also changed the test to use `Ash.DataLayer.Mnesia.create/2` directly.

One other thing I noticed while working through some of the Mnesia stuff is that the pkey is a List and the values are a single value in Mnesia that is wrapped in a Map. I assume this is not intentional, but wanted to check before I started working on it. It would be a backward incompatible change to move to proper Mnesia fields. Happy to have this conversation elsewhere too. ✌️

### DIFF
--- a/lib/ash/data_layer/mnesia/mnesia.ex
+++ b/lib/ash/data_layer/mnesia/mnesia.ex
@@ -445,7 +445,7 @@ defmodule Ash.DataLayer.Mnesia do
                Ash.DataLayer.Mnesia.Info.table(resource),
                pkey,
                values,
-               opts[:with_transaction]
+               Keyword.get(opts, :with_transaction, true)
              ) do
           # If with_transaction is false, we are in a transaction and will only return :ok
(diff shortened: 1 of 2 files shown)


zachdaniel:
> One other thing I noticed while working through some of the Mnesia stuff is that the pkey is a List and the values are a single value in Mnesia that is wrapped in a Map. I assume this is not intentional, but wanted to check before I started working on it. It would be a backward incompatible change to move to proper Mnesia fields. Happy to have this conversation elsewhere too. ✌️

Not fully following this issue, perhaps you could open a PR/issue w/ an example of the issue?

PR #2341, opened 2025-09-30
feat:  rollback transactions on error option

barnabasJ:
# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -3803,16 +3803,6 @@ defmodule Ash do
 
       To send notifications use `Ash.Notifier.notify(notifications)`. It sends any notifications that can be sent, and returns the rest.
       """
-    ],
-    rollback_on_error?: [
-      type: :boolean,
-      default: Application.compile_env(:ash, :transaction_rollback_on_error?, false),
-      doc: """
-      Whether or not to rollback the transaction on error, if the resource is in a transaction.
-
-      If the callback returns `{:error, _}`, automatically calls `Ash.DataLayer.rollback/2`.
-      Defaults to the value of `:transaction_rollback_on_error?` in your `:ash` config, or `false` if not set.
-      """
     ]
   ]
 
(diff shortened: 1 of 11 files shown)


zachdaniel:
I think we should add `transact/1` and deprecate `transaction/1`. What do you think?

barnabasJ:
> I think we should add `transact/1` and deprecate `transaction/1`. What do you think?

Do you mean only `Ash.transaction` ? Some of the code is making use of the fact that `Ash.DataLayer.transaction` doesn't necessarily rollback automatically.

zachdaniel:
Correct, we'd only deprecate `Ash.transaction`.

barnabasJ:
> Correct, we'd only deprecate `Ash.transaction`.

Ok, and transact would always rollback_on_error, or would it have an option for it?

zachdaniel:
Yeah `transact` would just always rollback on error, mirroring recent Ecto changes.

barnabasJ:
> Yeah `transact` would just always rollback on error, mirroring recent Ecto changes.

Alright, sounds good, I'll make the change.

lib/ash.ex
```diff
@@ -3814,25 +3814,130 @@ defmodule Ash do
     use Spark.Options.Validator, schema: transaction_opts
   end
 
+  @doc deprecated: """
```
zachdaniel: These docs are strangely combined. The `@doc deprecated: ...` should be separate from the function docs.

barnabasJ: Ah, I wasn't sure about the syntax and the docs don't show an example with both deprecated and regular docs. Do I just add another @doc before the function or how does that work?

zachdaniel: Ohh I see what you were going for. Yeah its `@doc deprecated: "string"` and `@doc "the docs"` 👍

zachdaniel:

PR #2350, opened 2025-10-02
fix: remove non public arguments in generator action_input

rapidfsub:
Simple bug fix


zachdaniel:

PR #2352, opened 2025-10-03
fix: Incorrect return type checking for generic action hooks

jimsynz:
Generic action after_action, after_transaction and around_transaction hooks were not correctly validating return types for generic actions with no return type.

# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/action_input.ex
+++ b/lib/ash/action_input.ex
@@ -46,7 +46,9 @@ defmodule Ash.ActionInput do
   """
   @type after_action_fun ::
           (t, term ->
-             {:ok, term}
+             :ok
+             | {:ok, [Ash.Notifier.Notification.t()]}
+             | {:ok, term}
              | {:ok, term, [Ash.Notifier.Notification.t()]}
              | {:error, any})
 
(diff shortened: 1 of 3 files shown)


lib/ash/action_input.ex
```diff
@@ -1482,52 +1506,94 @@ defmodule Ash.ActionInput do
...
+
             #{inspect(other)}
             """
         end
       end
     )
   end
 
+  defp merge_after_action_notifications(input, result, old_notifications, new_notifications) do
+    Enum.map(
```
zachdaniel, line 1577: This is kind of a new feature that I don't think we do anywhere else? I.e we expect notifications to be fully built when given to us. Not a bad idea though.

zachdaniel:

lib/ash/action_input.ex
jimsynz: Nah man the code was already there, I just moved it into a private function to reuse it.

PR #2355, opened 2025-10-07
Mnesia DL: internal layout and in-mnesia Filter support

zentourist:
This was a fun one! Compound Primary Keys almost broke me 😆  I started thinking this would take a day or two. I was wrong...

What this does:

* Changes the internal format of the Ash.Resource records when they get stored in Mnesia so that we can better query against the values using Matchspecs
* Builds MatchSpecs from `Ash.Filter` so that we can do in-Mnesia filtering. Today this is just using guards. I would like to eventually add Pattern Matching in the MatchHead for even more efficient queries.

It does NOT support all Filter functionality and if we get into any issues it will simply fallback to runtime filtering for now.

I added comprehensive tests for the Matchspec items. The tests were largely generated with Sonnet 4.5, but I reviewed them in depth.

Initially, this broke a bunch of tests in the suite, but I was able to get everything working again without touching any of the existing testing so feature-wise it should be backward compatible. NOTE: The internal format does change so if there are any persisted tables out there **_THIS WILL BREAK THEM_**.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/data_layer/mnesia/matchspec.ex
+++ b/lib/ash/data_layer/mnesia/matchspec.ex
@@ -1,0 +1,253 @@
+defmodule Ash.DataLayer.Mnesia.MatchSpec do
+  @moduledoc """
+  Converts an `Ash.Filter` to Mnesia matchspecs for efficient querying.
+
+  ## Supported Operators
+
+  * Equality: `Eq`, `NotEq`, `IsNil`
+  * Comparison: `GreaterThan`, `GreaterThanOrEqual`, `LessThan`, `LessThanOrEqual`
+  * Logical: `And`, `Or`, `Not`
+  * Membership: `In`
+
+  ## Limitations and room for improvement:
+
+  * Right now we are building up a Guard and using wildcards for the MatchHead.
+    This will be much faster than the [Runtime
+    filters](lib/ash/filter/runtime.ex), but it could be improved upon by
+    determining when to use a simple MatchHead if the filters are sufficiently
+    simple.
+
+  * We are always returning the full record (:"$_"). We could optimize for
+    `select` queries by returning only the necessary fields.
+
+  * There is no support for [Has queries](lib/ash/query/operator/has.ex).
+  """
+
+  @typedoc """
+  A guard expression used in Mnesia matchspecs.
+  Guards are tuples representing conditional expressions.
+
+  The grammar for the MatchSpec Guards is available on the
+  [erts matchspec page](https://www.erlang.org/doc/apps/erts/match_spec).
+  """
+  @type guard :: tuple() | atom() | boolean()
+
+  @typedoc """
+  The MatchSpec uses positional arguments to represent the fields being matched.
+  This is a mapping of our `Ash.Resource` fields to matchspec variables (e.g.,
+  :"$1", :"$2").
+  """
+  @type field_map :: %{atom() => atom()}
+
+  @typedoc """
+  A complete Mnesia matchspec in the format [{match_head, [guards], [result]}].
+
+  For more information about the parts of a MatchSpec, see Mnesia's
+  [select/3](https://www.erlang.org/doc/apps/mnesia/mnesia.html#select/3).
+  """
+  @type matchspec :: [{tuple(), [guard()], [term()]}]
+
+  alias Ash.Query
+
+  alias Ash.Query.Operator.{
+    Eq,
+    GreaterThan,
+    GreaterThanOrEqual,
+    In,
+    IsNil,
+    LessThan,
+    LessThanOrEqual,
+    NotEq
+  }
+
+  @doc """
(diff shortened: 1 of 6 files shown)


zachdaniel:
🤔 NOTE: The internal format does change so if there are any persisted tables out there THIS WILL BREAK THEM.

We are going to need to figure something out here then, like making these new changes backwards compatible. Alternatively, we could break out a new data layer called `AshMnesia` for example, so that we can iterate on breaking changes.

lib/ash/data_layer/mnesia/transformers/define_records.ex
```diff
@@ -0,0 +1,258 @@
...
+
+        For compound primary keys, returns `[:_pkey | non_pk_fields]`.
+        For simple primary keys, returns all attribute names.
+
+        ## Examples
+
+            iex> #{inspect(__MODULE__)}.mnesia_record_info()
+            [:id, :name, :age, :email]
+        """
+        def mnesia_record_info do
```
zachdaniel, line 164: As opposed to defining functions inside the resource, we should use functions in `Ash.DataLayer.Mnesia.Info`, and use the transformer to put this information into the persisted data, i.e `Spark.Dsl.Transformer.persist`

lib/ash/data_layer/mnesia/matchspec.ex
```diff
@@ -0,0 +1,253 @@
+defmodule Ash.DataLayer.Mnesia.MatchSpec do
```
zachdaniel, line 1: This is really cool! I think this actually dovetails into something I'd really like to exist for this, however, which is something called `Ash.Filter.split`, which you pass in a filter, and a function that says "do you support this expression element" and it gives you back `{ands_that_you_support, ors_that_you_dont}`, that way you can use the first thing to build your own filter statement, and the second thing can be done with runtime filtering at the end.

Would you be open to this? This is *huge* BTW.

zentourist:
> 🤔 NOTE: The internal format does change so if there are any persisted tables out there THIS WILL BREAK THEM.
> 
> We are going to need to figure something out here then, like making these new changes backwards compatible. Alternatively, we could break out a new data layer called `AshMnesia` for example, so that we can iterate on breaking changes.

Yeah, I've been noodling on this. I don't hate the idea of breaking it out, but it is nice to have a good data_layer in the core for testing. I noticed a few tests use the mnesia DL. We could also just add a `V2` namespace. That would also allow for easy benchmarks between the two :)

I also had a thought about introducing migrations using [transform_table](https://www.erlang.org/doc/apps/mnesia/mnesia.html#transform_table/4), but I haven't spent any time digging into that yet.

lib/ash/data_layer/mnesia/transformers/define_records.ex
zentourist: 👍  - I'll see if I can figure this out. Are you talking about just this function or reorganizing them all this way? If so, I might need an example to point to.

lib/ash/data_layer/mnesia/matchspec.ex
zentourist: That sounds like a cool solution. 🆒 

I think that would definitely clean up the logic a bit and then we wouldn't have to do an all or nothing approach that we are here. My only concern is, will the filters split that way naturally. Or is there fallout if we do the split and run some in the datalayer and some at runtime?

zentourist:
I had one additional thought on the internal table format. I might be able to support both versions by introducing something like this:

```elixir
mnesia do
  table :table_name
  version 1
end
```

If someone uses v1 for the format, it would work in a backward compatible way and allow for deprecation. I'd have to give this a go, but I think it's possible.

lib/ash/data_layer/mnesia/matchspec.ex
zachdaniel: The concept with splitting is that you only split across `*and*` boundaries. i.e

```
a == 1 and (b == 2 or has(c, "value"))
```

You'd get back

`{expr(a == 1), expr(b == 2 or has(c, "value"))}`.

So the left filter can always be applied, and the right is applied afterwards.

zachdaniel:
Yep, I think versioning makes a lot of sense. We can potentially introduce a backwards compatibility configuration (see the docs on backwards compatibility configs) to make new apps default to the new better version, or make the option required (or warn when unset) so that users will have to configure it and can decide how to handle it.

zentourist:
Hey @zachdaniel - Just wanted to let you know that I'm not abandoning this. I was on vacation and am just getting back. I should be able to get back to this soon. ✌️

zentourist:
Sorry to have dropped the ball on this @zachdaniel. Life happened...

Hopefully I can bring this across the finish line at some point. ✌️

PR #2356, opened 2025-10-07
Improve selection of atomic validation attribute, including resource-level option

stevebrambilla:
The way atomic validations currently work is that it picks the first primary key field and uses that for the validation expression, assigning the current value to itself if the condition passes.

In our case, the selected primary key field is also the distribution column used for Citus, so we get this error:

```
** (Postgrex.Error) ERROR 0A000 (feature_not_supported) modifying the partition value of rows is not allowed
```

This PR makes two changes around this:
1. If there is already another atomic attribute in the changeset, prefer using it.
2. Added a new `atomic_validation_default_target_attribute` to `:resource` so resources can specify a compatible attribute. 

The atomic validation attribute now take the first of:
- First atomic attribute already in the changeset
- The resource's `atomic_validation_default_target_attribute`
- The resource's first primary key field (current behaviour)

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/actions/atomic_update_test.exs
+++ b/test/actions/atomic_update_test.exs
@@ -148,6 +148,7 @@ defmodule Ash.Test.Actions.AtomicUpdateTest do
       end
 
       update :validation_without_update do
+        accept [:title, :year]
         validate compare(:year, greater_than_or_equal_to: 2000)
       end
     end
@@ -279,6 +280,17 @@ defmodule Ash.Test.Actions.AtomicUpdateTest do
     # Validation passes for recent books, fails for older ones.
     assert {:ok, %Book{year: 2020}} = Book.validation_without_update(recent_book)
     assert {:error, %Ash.Error.Invalid{}} = Book.validation_without_update(older_book)
+  end
+
+  test "can update validation default target attribute" do
+    recent_book =
+      Book
+      |> Ash.Changeset.for_create(:create, %{title: "RecentBook", year: 2020})
+      |> Ash.create!()
+
+    updated_book = Book.validation_without_update!(recent_book, %{title: "UpdatedTitle"})
+
+    assert updated_book.title == "UpdatedTitle"
   end
 
   test "policies that require original data" do


test/actions/atomic_update_test.exs
```diff
@@ -188,6 +233,54 @@ defmodule Ash.Test.Actions.AtomicUpdateTest do
...
+      |> Ash.Changeset.for_create(:create, %{title: "RecentBook", year: 2020})
+      |> Ash.create!()
+
+    older_book =
+      Book
+      |> Ash.Changeset.for_create(:create, %{title: "OlderBook", year: 1960})
+      |> Ash.create!()
+
+    # Validation passes for recent books, fails for older ones.
+    assert {:ok, %Book{year: 2020}} = Book.validation_without_update(recent_book)
```
zachdaniel, line 281: We need to validate that this does not prevent the underlying change to this attribute. Shouldn't this be 2021?

stevebrambilla: There are two actions on `Book`, one of them does the `atomic_update` to increment to 2021 and the other (this one) doesn't.

As soon as you have another atomic attribute in the changeset, it gets chosen as the validation attribute, so the only way to exercise the default target attribute is not having any existing atomic attributes there.

Want me to add a separate test just validating that we can still update `:title`?

zachdaniel: That would be great yes, but probably other tests would have failed in retrospect. Can't hurt IMO

stevebrambilla: Sure, just pushed another commit with the extra test

zachdaniel:

zachdaniel:
@stevebrambilla just a note for you in case it matters, this logic has been slightly modified to actually choose the *last* changing attribute in the atomics list, not the first.

PR #2363, opened 2025-10-10
Fix: use is_nil in ets upsert filter if the value is nil

barnabasJ:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/ash/data_layer/ets_test.exs
+++ b/test/ash/data_layer/ets_test.exs
@@ -119,6 +119,21 @@ defmodule Ash.DataLayer.EtsTest do
     assert [{%{id: ^id}, %EtsTestUser{name: "Joe", id: ^id}}] = user_table()
   end
 
+  test "upsert with nil key value" do
+    %EtsTestUser{id: id} =
+      create_user(%{name: nil}, upsert?: true, upsert_identity: :unique_name)
+
+    assert [{%{id: ^id}, %EtsTestUser{id: ^id, name: nil}}] = user_table()
+
+    create_user(%{name: nil, age: 25},
+      upsert?: true,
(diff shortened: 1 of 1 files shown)


lib/ash/data_layer/ets/ets.ex
```diff
@@ -1449,9 +1449,16 @@ defmodule Ash.DataLayer.Ets do
...
         Enum.map(keys, fn key ->
+          value =
+            Ash.Changeset.get_attribute(changeset, key) || Map.get(changeset.params, key) ||
+              Map.get(changeset.params, to_string(key))
+
           {key,
-           Ash.Changeset.get_attribute(changeset, key) || Map.get(changeset.params, key) ||
-             Map.get(changeset.params, to_string(key))}
+           if is_nil(value) do
+             [is_nil: true]
```
zachdaniel, line 1458: Is this how postgres upserts work?

barnabasJ: Let me try that, but I assume so, otherwise the `nils_distinct?` flag wouldn't make sense to me. But let me add the same test to AshPostgres to make sure

zachdaniel: We'd have to look and see if the upsert identity is `nils_distinct?` in that case I think.

barnabasJ: I think that is handled by the if above.

zachdaniel: Right you are. Thanks for the explanation!

zachdaniel:

PR #2368, opened 2025-10-10
Update ash.install.ex - Change default sort order to alphabetical for…

nikfp:
… resource and domain sections

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

## Notes from contributor

I understand that this might not be wanted or needed, in which case feel free to close. For my own workflows, I feel that this would be a better starting order for looking through resource and domain definitions.


zachdaniel:
Thank you for the PR but its subjective so I have the benefit of keeping it how I like it 😆

PR #2379, opened 2025-10-13
test: add test to show missing nil check

barnabasJ:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/type/nil_apply_constraints_test.exs
+++ b/test/type/nil_apply_constraints_test.exs
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2020 Zach Daniel
+#
+# SPDX-License-Identifier: MIT
+
 defmodule Ash.Test.Type.NilApplyConstraintsTest do
   @moduledoc false
   use ExUnit.Case, async: true


zachdaniel:
Could you describe specifically what this is testing? Is the issue that we are passing in a `nil` value to `apply_constraints` in some cases when in others we aren't?

barnabasJ:
> Could you describe specifically what this is testing? Is the issue that we are passing in a `nil` value to `apply_constraints` in some cases when in others we aren't?

It's mostly around asking if this is expected behaviour or not. I saw that for arguments, we have a `nil` check in one of the code paths before calling apply_constraints.

setting arguments:
https://github.com/ash-project/ash/blob/daa69f1e537e106707164b9f3ab4f400abd1ba9e/lib/ash/changeset/changeset.ex#L6091-L6091
change attribute:
https://github.com/ash-project/ash/blob/daa69f1e537e106707164b9f3ab4f400abd1ba9e/lib/ash/changeset/changeset.ex#L6310-L6310
force changing attributes:
https://github.com/ash-project/ash/blob/daa69f1e537e106707164b9f3ab4f400abd1ba9e/lib/ash/changeset/changeset.ex#L6449-L6449

and then there is also this case 

https://github.com/ash-project/ash/blob/daa69f1e537e106707164b9f3ab4f400abd1ba9e/lib/ash/type/type.ex#L1223-L1223

Maybe we just miss the `nil` check in Ash.Type for non array types?

zachdaniel:
🤔 yeah its interesting. I guess it's an inconsistency and we should choose which is better? I think we should allow `nil` through to `apply_constraints`

zachdaniel:
Would you mind making that change?

barnabasJ:
> 🤔 yeah its interesting. I guess it's an inconsistency and we should choose which is better? I think we should allow `nil` through to `apply_constraints`

> Would you mind making that change?


Just to make sure, remove the extra guard from the set_arguments function and add the guard in Ash.Type?

zachdaniel:
> Just to make sure, remove the extra guard from the set_arguments function and add the guard in Ash.Type?

Remove the extra guard.  we should pass `nil` values into constraint application.

barnabasJ:
closed in favor of https://github.com/ash-project/ash/pull/2389

PR #2357, opened 2025-10-07
fix: extended context collision prevention to all bulk operation types

DGollings:
and added some convenience helpers

As requested here, https://github.com/ash-project/ash/pull/2353#issuecomment-3374737225

its a bit more extensive, added bulk_destroy whilst at it. But tried to keep it to a minimum with some extra tests

# Contributor checklist

Leave anything that you believe does not apply unchecked.

- [X] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [X] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [X] Refactoring
- [ ] Update dependencies

PR #2353, opened 2025-10-05
fix: improve nested bulk action notification handling

DGollings:
# Contributor checklist

Leave anything that you believe does not apply unchecked.

- [X] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [X] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [X] Refactoring
- [ ] Update dependencies

As discussed in https://github.com/ash-project/ash_graphql/issues/372

This PR fixes notifications being nil in nested operations by using improved tracking of their respective contexts

PR #2398, opened 2025-10-17
fix: key :resource not found in: nil

DGollings:
# Contributor checklist

- [X] Bug fixes ~~include regression tests~~

I thought that #2353 and #2357  solved the below issue, but after upgrading noticed it was still occurring
```elixir
** (Ash.Error.Unknown) 
Bread Crumbs:
  > Exception raised in bulk update: MyApp.foo
  > Exception raised in bulk update: MyApp.foo

Unknown Error

* ** (KeyError) key :resource not found in: nil

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
  (ash 3.6.2) lib/ash/actions/update/bulk.ex:3018: Ash.Actions.Update.Bulk.notification/3
  (ash 3.6.2) lib/ash/actions/update/bulk.ex:2744: anonymous fn/5 in Ash.Actions.Update.Bulk.process_results/10
  (elixir 1.18.4) lib/enum.ex:4442: Enum.flat_map_list/2
  (ash 3.6.2) lib/ash/actions/update/bulk.ex:2740: Ash.Actions.Update.Bulk.process_results/10
  (ash 3.6.2) lib/ash/actions/update/bulk.ex:1957: Ash.Actions.Update.Bulk.do_handle_batch/14
  (ash 3.6.2) lib/ash/data_layer/data_layer.ex:463: anonymous fn/3 in Ash.DataLayer.transaction/5
  (zelo 0.1.0) lib/zelo/repo.ex:50: anonymous fn/1 in MyApp.Repo."transaction (overridable 1)"/2
  (ecto 3.13.3) lib/ecto/repo/transaction.ex:7: anonymous fn/2 in Ecto.Repo.Transaction.transact/4
```

The problem I have is that the attached commit 8eaabba fixes it, or better, bypasses the issue of metadata being 'corrupted' when it roundtrips to/through (JSONB?) Postgres. 

My issue is solved and I'd like to submit a real PR, but I don't know how to reproduce whatever it is that triggers this issue. 
Have tried adding various Ash extensions and tried to reproduce our action in all its nested glory, but no luck yet. Does anyone have a suggestion how to prove that this code solves the above KeyError and whether this is an appropiate way to fix it?

We call this action through Ash_Graphql, have tenancy enabled and use Ash_Postgres (plus a bunch of extensions like Ash_Jason, Ash_Papertrail, Ash_JasonAPI

Might be related to #2345 / #2362

### DIFF
--- a/lib/ash/actions/update/bulk.ex
+++ b/lib/ash/actions/update/bulk.ex
@@ -2731,7 +2731,7 @@ defmodule Ash.Actions.Update.Bulk do
          domain,
          base_changeset
        ) do
-    run_bulk_after_changes(
+    results_after_changes = run_bulk_after_changes(
       changes,
       all_changes,
       batch,
@@ -2742,9 +2742,36 @@ defmodule Ash.Actions.Update.Bulk do
       base_changeset.resource,
       metadata_key
     )
-    |> Enum.flat_map(fn result ->
-      changeset = changesets_by_index[result.__metadata__[metadata_key]]
 
+    # Apply the proven AshPostgres pattern: extract metadata from changeset context and put it back onto results
+    results_with_reconstructed_metadata =
+      results_after_changes
+      |> Enum.with_index()
+      |> Enum.map(fn {result, index} ->
+        changeset = changesets_by_index[index]
+
+        if changeset do
+          case get_bulk_operation_metadata(changeset, :bulk_update) do
+            {index, metadata_key} ->
+              Ash.Resource.put_metadata(result, metadata_key, index)
+
+            nil ->
+              # Compatibility fallback - extract from context directly
+              context_metadata = changeset.context[:bulk_update]
+              if context_metadata && is_map(context_metadata) do
+                Ash.Resource.put_metadata(result, :bulk_update_index, context_metadata.index)
+              else
+                result
+              end
+          end
+        else
+          result
+        end
+      end)
+
+    # Zip results with changesets directly to avoid metadata lookup issues
+    results_with_reconstructed_metadata
+    |> Enum.zip_with(Map.values(changesets_by_index), fn result, changeset ->
       if opts[:notify?] || opts[:return_notifications?] do
         store_notification(ref, notification(changeset, result, opts), opts)
       end
@@ -2771,6 +2798,7 @@ defmodule Ash.Actions.Update.Bulk do
           []
       end
     end)
+    |> List.flatten()
     |> load_data(domain, resource, base_changeset, opts)
     |> case do
       {:ok, records} ->
@@ -2817,6 +2845,22 @@ defmodule Ash.Actions.Update.Bulk do
       other ->
         other
     end
+  end
+
+  defp get_bulk_operation_metadata(changeset, bulk_action_type) do
+    changeset.context
+    |> Enum.find_value(fn
+      # New format: {{:bulk_create, ref}, value} -> {index, metadata_key}
+      {{^bulk_action_type, ref}, value} ->
+        {value.index, {:"#{bulk_action_type}_index", ref}}
+
+      # Fallback for old format: {:bulk_create, value} -> {index, metadata_key}
+      {^bulk_action_type, value} when is_map(value) ->
+        {value.index, :"#{bulk_action_type}_index"}
+
+      _ ->
+        nil
+    end)
   end
 
   def run_bulk_after_changes(


zachdaniel:
Can you try with the latest version of `AshPostgres`? The code I'm thinking of may be in `main` and not yet released, I'd have to check. Does this perhaps happen for you on upserts specifically? There were some fixes in this area.

DGollings:
From my mix.exs

```elixir
{:ash, github: "ash-project/ash", ref: "e91fa3c2cc552f5081dfe34d7845197331797789", override: true}, # main
{:ash_postgres, github: "ash-project/ash_postgres", ref: "449ae3ff8715c3bb5598ff19d844e1ac728e25b7", override: true}, # main
```

still results in 
```elixir
** (Ash.Error.Unknown) 
Bread Crumbs:
  > Exception raised in bulk update: MyApp.foo
  > Exception raised in bulk update: MyApp.foo

Unknown Error

* ** (KeyError) key :resource not found in: nil

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
  (ash 3.7.3) lib/ash/actions/update/bulk.ex:3023: Ash.Actions.Update.Bulk.notification/3
  (ash 3.7.3) lib/ash/actions/update/bulk.ex:2749: anonymous fn/5 in Ash.Actions.Update.Bulk.process_results/10
  (elixir 1.18.4) lib/enum.ex:4442: Enum.flat_map_list/2
  (ash 3.7.3) lib/ash/actions/update/bulk.ex:2745: Ash.Actions.Update.Bulk.process_results/10
  (ash 3.7.3) lib/ash/actions/update/bulk.ex:1961: Ash.Actions.Update.Bulk.do_handle_batch/14
  (ash 3.7.3) lib/ash/data_layer/data_layer.ex:470: anonymous fn/3 in Ash.DataLayer.transaction/5
  (myapp 0.1.0) lib/myapp/repo.ex:50: anonymous fn/1 in MyApp.Repo."transaction (overridable 1)"/2
  (ecto 3.13.3) lib/ecto/repo/transaction.ex:7: anonymous fn/2 in Ecto.Repo.Transaction.transact/4
  (ecto_sql 3.13.2) lib/ecto/adapters/sql.ex:1458: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
  (db_connection 2.8.1) lib/db_connection.ex:1753: DBConnection.run_transaction/4
  (ash 3.7.3) lib/ash/actions/update/bulk.ex:1842: Ash.Actions.Update.Bulk.handle_batch/11
  (ash 3.7.3) lib/ash/actions/update/bulk.ex:1396: anonymous fn/13 in Ash.Actions.Update.Bulk.do_stream_batches/7
  (elixir 1.18.4) lib/stream.ex:626: anonymous fn/4 in Stream.map/2
  (elixir 1.18.4) lib/stream.ex:1761: anonymous fn/3 in Enumerable.Stream.reduce/3
  (elixir 1.18.4) lib/stream.ex:302: Stream.after_chunk_while/2
  (elixir 1.18.4) lib/stream.ex:1790: Enumerable.Stream.do_done/2
  (elixir 1.18.4) lib/stream.ex:1773: Enumerable.Stream.do_each/4
  (elixir 1.18.4) lib/stream.ex:956: Stream.do_transform/5
  (elixir 1.18.4) lib/enum.ex:4515: Enum.reverse/1
  (elixir 1.18.4) lib/enum.ex:3835: Enum.to_list/1

    (ash 3.7.3) lib/ash/actions/update/bulk.ex:3023: Ash.Actions.Update.Bulk.notification/3
    (ash 3.7.3) lib/ash/actions/update/bulk.ex:2749: anonymous fn/5 in Ash.Actions.Update.Bulk.process_results/10
    (elixir 1.18.4) lib/enum.ex:4442: Enum.flat_map_list/2
    (ash 3.7.3) lib/ash/actions/update/bulk.ex:2745: Ash.Actions.Update.Bulk.process_results/10
    (ash 3.7.3) lib/ash/actions/update/bulk.ex:1961: Ash.Actions.Update.Bulk.do_handle_batch/14
    (ash 3.7.3) lib/ash/data_layer/data_layer.ex:470: anonymous fn/3 in Ash.DataLayer.transaction/5
    (myapp 0.1.0) lib/myapp/repo.ex:50: anonymous fn/1 in MyApp.Repo."transaction (overridable 1)"/2
    (ecto 3.13.3) lib/ecto/repo/transaction.ex:7: anonymous fn/2 in Ecto.Repo.Transaction.transact/4
    (ecto_sql 3.13.2) lib/ecto/adapters/sql.ex:1458: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
    (db_connection 2.8.1) lib/db_connection.ex:1753: DBConnection.run_transaction/4
    (ash 3.7.3) lib/ash/actions/update/bulk.ex:1842: Ash.Actions.Update.Bulk.handle_batch/11
    (ash 3.7.3) lib/ash/actions/update/bulk.ex:1396: anonymous fn/13 in Ash.Actions.Update.Bulk.do_stream_batches/7
    (elixir 1.18.4) lib/stream.ex:626: anonymous fn/4 in Stream.map/2
    (elixir 1.18.4) lib/stream.ex:1761: anonymous fn/3 in Enumerable.Stream.reduce/3
    (elixir 1.18.4) lib/stream.ex:302: Stream.after_chunk_while/2
    (elixir 1.18.4) lib/stream.ex:1790: Enumerable.Stream.do_done/2
    (elixir 1.18.4) lib/stream.ex:1773: Enumerable.Stream.do_each/4
    (elixir 1.18.4) lib/stream.ex:956: Stream.do_transform/5
    (elixir 1.18.4) lib/enum.ex:4515: Enum.reverse/1
    (elixir 1.18.4) lib/enum.ex:3835: Enum.to_list/1
```

unfortunately

Managed to track it down to (I think) this
- Changeset indexed with: integer 0
- Result metadata contains: base64-encoded string "g2o=" (which represents integer 0)
- Lookup fails: because changesets_by_index[0] exists but changesets_by_index["g2o="] doesn't

And noticed that certain metadata is returned just fine (`:tenant` for example) but other metadata has to be readded (`:selected`)
This fix tries to do the same, but I did not spend enough time in the internals to be sure.

DGollings:
Also, not an upsert, at least not intentionally. Just (nested) bulk update(s) via GraphQL

zachdaniel:
What on earth would be encoding the index?? I think I'm going to need to see a failing test here to proceed. That failing test can be in ash_postgres if that's the only place you can reproduce it.

zachdaniel:
Also I did make a change to this logic before, I made it so that returned entries had their bulk index set without the reference because having a reference there broke the callers api.

DGollings:
The plot thickens, because I had so much trouble trying to reproduce this in ash(_postgres) I let Claude sniff around with `dbg`

# Ash papertrail

Now I need to understand the actual problem. The issue is that when bulk_create! is called in
AshPaperTrail (line 332), it returns notifications. These notifications have metadata that contains tuples
like {0, changeset}. But when these notifications flow back through the system, somehow the metadata gets
corrupted.

Let me check if there's something special about how AshPaperTrail stores the changes in JSONB:

Based on my investigation, I've found the root cause! Here's what's happening:

  1. AshPaperTrail's after_batch hook processes bulk updates and creates version records
  2. The notifications returned from bulk_create! (line 333 in create_new_version.ex) contain metadata
  3. The metadata has tuples like {index, changeset} that Ash uses to correlate bulk operations
  4. These tuples cannot be encoded to JSON and get corrupted when flowing through the system
  5. The corruption turns index 0 into "g2o=" (base64 for empty list [])

  The issue is that AshPaperTrail's bulk_create! returns notifications that have already gone through some
  transformation, possibly through JSONB storage, corrupting the metadata.

# Root Cause Analysis

  The bug occurs in AshPaperTrail's after_batch callback. Here's the issue:

  1. Line 333 in create_new_version.ex: |> Map.get(:notifications) - AshPaperTrail returns notifications
  from its internal bulk_create! call
  2. These notifications have metadata that's been corrupted through JSONB serialization
  3. The metadata contains tuples {index, changeset} that can't be properly encoded to JSON
  4. When returned to Ash's bulk update system, the corrupted metadata causes the KeyError

  The problem is that AshPaperTrail shouldn't be returning these notifications at all, or it should preserve
   the metadata correctly. The notifications it returns are for the version records it creates, not for the
  original resource updates.

  The Fix

  The issue is in
  deps/ash_paper_trail/lib/resource/changes/create_new_version.ex at line
   333. The bulk_create! method returns notifications, but these are notifications about the paper trail
  version records being created, not the original notifications. These notifications have corrupted metadata
   because they've gone through JSONB storage.

  The fix would be to either:
  1. Don't return the notifications from paper trail's bulk_create
  2. Return empty/no notifications from after_batch
  3. Fix the metadata preservation in the notifications

  Since AshPaperTrail is a dependency, we can:
  1. Report this bug to AshPaperTrail
  2. Temporarily disable paper_trail for the assign_vehicle action
  3. Apply a patch to the dependency

DGollings:
so, these ash paper trail notifications, bug or feature? :)

zachdaniel:
> These notifications have corrupted metadata
because they've gone through JSONB storage.

Where in AshPaperTrail do we send notifications through jsonb storage? We don't do that anywhere that I know of.

zachdaniel:
And yes, notifications should be created for all resource changes including version resources.

zachdaniel:
Looking back, I'm not sure I understand what the problem was with the *non* tuple versions of bulk indexes actually 🤔 I think there may have been a disconnect there, I was thinking that we must have been setting those values into the pdict (which we do for some things), but looking at it I don't see why we are and I don't see why the reference was needed in retrospect 😢

zachdaniel:
Yeah, so I think I'm going to temporarily revert the original changes. The reason for this is twofold:

1. I've had to make some changes to it to make it so that its possible to store embedded resources at compile time (i.e `default [%{a: 1}]` for an embedded resource. It was adding a `ref` to its metadata which can't be stored at compile time
2. The `bulk_create_index` was something that was being used in a few places I hadn't anticipated and changing it to a tuple is breaking that behavior
3. I need to understand more fully exactly what is going on here. I don't see any reason at all there would be JSON encoding of resource metadata.

zachdaniel:
Okay, so for now I'm going to close this particular suggestion. I've reverted the original changes and added back in the tests that were added to show that we don't yet have a reproduction of your original issue. Once we reproduce your original issue we can solve it from there.

DGollings:
>  I've reverted the original changes

did you remember to also revert the one in `ash_postgres`?

I'd agree that the tuple is breaking more than expected, the idea was to have a simple `{:bulk_update, ref}' similar to `{:ok, result}` and `{:error, reason}` but with whatever is creating this serialization issue is having trouble with tuples (in retrospect, more likely something erlang or ecto-y than jsonb)

My original issue was solved though, and I did have a reproduction of the issue in those commits you reverted. It was not until later, and a different action, where `ash_papertrail` somehow was also triggered.

So if it originally was `:bulk_update` and `{:bulk_update, ref}` is causing issues, what about something like `:bulk_update_#{index}` or `:bulk_update_#{uuid}`? Or even just `unique_int` and leave the bulk update part out of it. 

I'd be happy to submit another PR without tuples

zachdaniel:
@DGollings hmm....I adapted your tests and was not able to get a test failure. And yeah I reverted the ash_postgres one as well, thanks! Can you take a look at the latest version of the tests to see how what I was doing is different from what you were doing?

DGollings:
going to try to reproduce it again in ash_postgres with ash_papertrail (my current lead)

but I can confirm it still errors after updating ash and ash_postgres

Quick question, mostly out of curiosity or just in case: did the `"nested bulk operations"`not add any value? Noticed you removed those tests, yet thought they might be NTH nonetheless.

zachdaniel:
Oh, um...maybe I did it by accident, sorry 😓 Lets just start by getting the failing tests again and revisit this fix from the beginning (also sorry I'm sure that is inconvenient and frustrating)

DGollings:
I could submit a separate PR for those, but no worries, as frustrating as it is to exactly reproduce I'm glad to help figure it out. Just unhappy that the repro I thought I had turned not to be one. But we'll get there

PR #2399, opened 2025-10-18
fix: evaluate action_type earlier in the simplify phase to enable short-circuit evaluation

jechol:
`action_type` wasn’t being evaluated early enough. Even when the action was already known, the policy would still attempt to evaluate all checks — even when the action type didn’t match.

For example, with the following policy (which worked fine in 3.6.2 but broke afterward):

```elixir
policy action_type([:create, :update]) do
  authorize_if AshStateMachine.Checks.ValidNextState
end
```

When running a read action, `action_type` wouldn’t immediately evaluate to false. As a result, the expression `action_type([:create, :update]) and ValidNextState` remained as-is, and `ValidNextState` was still evaluated even though the action type didn’t match. Since `ValidNextState.filter/3` only works with changesets, this caused a FunctionClauseError when called with a query.

With this change, `action_type` evaluates to false immediately for read actions, causing the entire expression to become false, and `ValidNextState` is skipped.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/policy/check/action_type.ex
+++ b/lib/ash/policy/check/action_type.ex
@@ -6,8 +6,6 @@ defmodule Ash.Policy.Check.ActionType do
   @moduledoc "This check is true when the action type matches the provided type"
   use Ash.Policy.SimpleCheck
 
-  import Crux.Expression, only: [b: 1]
-
   @impl true
   def describe(options) do
     {operator, type} =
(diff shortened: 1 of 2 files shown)


zachdaniel:
Hmm... so I'm not sure that this is the best way to handle this one. The basic issue is that simplification logic is used elsewhere, and this feels like a missing step somewhere else, i.e "simplify -> check resulting expression" step is missing.

zachdaniel:
Actually @jechol could you try `main`? This may have been fixed by my most recent commit.

jechol:
Still reproducing at 4a3c61b

jechol:
Created a separate PR with a minimal reproduction test: #2401

The test shows a SimpleCheck that only handles changeset context crashing when called during a read action, even though the action_type([:create, :update]) condition is false.

maennchen:
@jechol The issue that causes this is the commutativity law in `crux`. For some more aggressive simplification, It ignore the order of variables in the expression.

I did a PR in `crux`, that gets rid of that: https://github.com/ash-project/crux/pull/4

maennchen:
@jechol That does not make your test pass, investigating...

maennchen:
@jechol There was a second issue. Second PR: https://github.com/ash-project/crux/pull/5

zachdaniel:
Latest release depends on the newest version of Crux. Thank you for the report and idea for the resolution 🙇

PR #2401, opened 2025-10-18
test: reproduce policy check called despite false condition

jechol:
When a policy condition evaluates to false, the checks inside the policy should not be called. Currently they are being evaluated anyway, which causes crashes when checks only handle specific contexts.

Example: action_type([:create, :update]) is false for read actions, but the filter check inside still gets called with query context, crashing if it only handles changeset context.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

zachdaniel:
This should be fixed in the latest release 🙇

PR #2411, opened 2025-10-27
fix: added bulk_action_ref for changeset lookups

DGollings:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

Otherwise as discussed @zachdaniel (I presume)

### DIFF
--- a/lib/ash/actions/update/bulk.ex
+++ b/lib/ash/actions/update/bulk.ex
@@ -2147,13 +2147,15 @@ defmodule Ash.Actions.Update.Bulk do
     end
   end
 
-  defp index_changesets_by_ref(batch, context_key) do
-    Enum.reduce(batch, %{}, fn changeset, changesets_by_ref ->
-      Map.put(
-        changesets_by_ref,
-        changeset.context[context_key].ref,
-        changeset
-      )
+  defp index_changesets(batch, context_key) do
+    Enum.reduce(batch, {%{}, %{}}, fn changeset, {by_ref, by_index} ->
+      ref = changeset.context[context_key].ref
+      index = changeset.context[context_key].index
+
+      {
+        Map.put(by_ref, ref, changeset),
+        Map.put(by_index, index, ref)
+      }
     end)
   end
 
@@ -2529,7 +2531,7 @@ defmodule Ash.Actions.Update.Bulk do
         end
       end)
 
-    changesets_by_ref = index_changesets_by_ref(batch, context_key)
+    {changesets_by_ref, changesets_by_index} = index_changesets(batch, context_key)
 
     batch =
       batch
@@ -2686,7 +2688,7 @@ defmodule Ash.Actions.Update.Bulk do
           end)
       end
 
-    {batch, changesets_by_ref}
+    {batch, changesets_by_ref, changesets_by_index}
   end
 
   defp manage_relationships(updated, domain, changeset, engine_opts) do
@@ -2704,7 +2706,7 @@ defmodule Ash.Actions.Update.Bulk do
   end
 
   defp run_after_action_hooks(
-         {batch_results, changesets_by_ref},
+         {batch_results, changesets_by_ref, changesets_by_index},
          opts,
          domain,
          ref,
@@ -2715,7 +2717,12 @@ defmodule Ash.Actions.Update.Bulk do
       Enum.flat_map(batch_results, fn result ->
         changeset =
           result
-          |> find_changeset(changesets_by_ref, metadata_key, ref_metadata_key)
+          |> find_changeset(
+            changesets_by_ref,
+            metadata_key,
+            ref_metadata_key,
+            changesets_by_index
+          )
           |> ensure_changeset!(result, metadata_key, ref_metadata_key)
 
         case manage_relationships(result, domain, changeset,
@@ -2759,11 +2766,11 @@ defmodule Ash.Actions.Update.Bulk do
         end
       end)
 
-    {results, changesets_by_ref}
+    {results, changesets_by_ref, changesets_by_index}
   end
 
   defp process_results(
-         {batch, changesets_by_ref},
+         {batch, changesets_by_ref, changesets_by_index},
          changes,
          all_changes,
          opts,
@@ -2780,6 +2787,7 @@ defmodule Ash.Actions.Update.Bulk do
       all_changes,
       batch,
       changesets_by_ref,
+      changesets_by_index,
       changesets,
       opts,
       ref,
@@ -2790,7 +2798,7 @@ defmodule Ash.Actions.Update.Bulk do
     |> Enum.flat_map(fn result ->
       changeset =
         result
-        |> find_changeset(changesets_by_ref, metadata_key, ref_metadata_key)
+        |> find_changeset(changesets_by_ref, metadata_key, ref_metadata_key, changesets_by_index)
         |> ensure_changeset!(result, metadata_key, ref_metadata_key)
 
       if opts[:notify?] || opts[:return_notifications?] do
@@ -2872,6 +2880,7 @@ defmodule Ash.Actions.Update.Bulk do
         all_changes,
         results,
         changesets_by_ref,
+        changesets_by_index,
         changesets,
         opts,
         ref,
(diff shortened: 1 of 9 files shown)


zachdaniel:
I believe the last thing we want to do is make sure that this key is cleared from any resulting record's metadata.

DGollings:
> I believe the last thing we want to do is make sure that this key is cleared from any resulting record's metadata.

done plus test

lib/ash/actions/update/bulk.ex
```diff
@@ -3453,4 +3516,84 @@ defmodule Ash.Actions.Update.Bulk do
...
+    ref_key = result.__metadata__[ref_metadata_key]
+
+    if ref_key do
+      changesets_by_ref[ref_key]
+    else
+      find_changeset_by_id_fallback(result, changesets_by_ref)
+    end
+  end
+
+  defp find_changeset_by_id_fallback(result, changesets_by_ref) do
```
zachdaniel: Let's remove this. Your usage of refs should authoritatively solve this. If no changeset can be found by ref we should just raise an error. And not all resources have an `id`.

DGollings: done, good call too as I immediately found some places where it was missing

lib/ash/actions/update/bulk.ex
```diff
@@ -3453,4 +3527,80 @@ defmodule Ash.Actions.Update.Bulk do
...
+    ref_key = result.__metadata__[ref_metadata_key]
+
+    if ref_key do
+      changesets_by_ref[ref_key]
+    else
+      raise "Missing ref metadata for record. The record should have had metadata key #{inspect(ref_metadata_key)} set during bulk update."
+    end
+  end
+
+  defp find_changeset_by_index(result, changesets_by_ref, metadata_key) do
```
zachdaniel: This seems incorrectly named, it should be `find_changeset_by_ref` right?

lib/ash/actions/update/bulk.ex
```diff
@@ -3453,4 +3527,80 @@ defmodule Ash.Actions.Update.Bulk do
...
+  defp find_changeset(result, changesets_by_ref, _metadata_key, ref_metadata_key)
+       when not is_nil(ref_metadata_key) do
+    find_changeset_by_ref(result, changesets_by_ref, ref_metadata_key)
+  end
+
+  defp find_changeset(result, changesets_by_ref, metadata_key, nil) do
+    find_changeset_by_index(result, changesets_by_ref, metadata_key)
+  end
+
+  defp add_metadata_to_results({:ok, results}, batch, metadata_key, ref_metadata_key, context_key) do
```
zachdaniel: Wait, this is wrong 😢 We shouldn't need to zip up after the fact. That is a new problem from this PR. We need to update the data layers that are returning this to return records with the proper `ref` attached, and then use that here. We can't assume that the data layer will return one record per input or that they will be in the order we gave them necessarily.

Likely what we need to do is this:
- check the first result to see if it has this new `ref` metadata.
  - if it does, then use `changesets_by_ref`
  - if it doesn't, then create a map of `index_to_ref`, and look up the ref in that map, then look up the changeset in `changesets_by_ref`.
- Then we need to update `ash_postgres`, `ash_sqlite`, and our built in data layers to return records with the new `ref` instead of this.

lib/ash/actions/update/bulk.ex
```diff
@@ -3453,4 +3527,80 @@ defmodule Ash.Actions.Update.Bulk do
...
+  defp result_matches_changes?(result, changes, changesets_by_ref, metadata_key, ref_metadata_key) do
+    if ref_metadata_key do
+      ref_key = result.__metadata__[ref_metadata_key]
+
+      if ref_key do
+        ref_key in List.wrap(changes)
+      else
+        # Fallback: find if this record ID has a changeset with ref in changes
+        Enum.any?(changesets_by_ref, fn {changeset_ref, changeset} ->
+          Map.get(changeset.data, :id) == result.id and
```
zachdaniel: Again, we cannot assume that records have an `id`. I don't understand why we would ever need to check this?

zachdaniel:
We're close! Thank you so much for bearing with me on this one.

lib/ash/actions/update/bulk.ex
DGollings: missed this whilst removing the fallbacks, my mistake

DGollings:
> We're close! Thank you so much for bearing with me on this one.

Appreciate (and understand) the thorough review :)
Resolved your comments in c2136df76a9fe9a43adaa0685855872052929fbb

Updated and also opened https://github.com/ash-project/ash_postgres/pull/645 and https://github.com/ash-project/ash_sqlite/pull/189 (~~the latter needs some work~~ tests should pass, but because I do not use ash_sqlite its harder to thoroughly test)

It passes my repro test, please review
(I'll check the failing CI jobs asap)

lib/ash/actions/update/bulk.ex
DGollings: you're right, I was overly trying to keep the PR small, but it should be the data layer setting this

lib/ash/actions/update/bulk.ex
```diff
@@ -1982,15 +2028,17 @@ defmodule Ash.Actions.Update.Bulk do
...
+    # Generate unique ref per changeset (not per batch)
+    ref = make_ref()
+
     record
     |> Ash.Changeset.new()
     |> Map.put(:domain, domain)
     |> Ash.Changeset.prepare_changeset_for_action(action, opts)
     |> Ash.Changeset.set_private_arguments_for_action(opts[:private_arguments] || %{})
-    |> Ash.Changeset.put_context(context_key, %{index: index})
+    |> Ash.Changeset.put_context(:bulk_update, %{index: index, ref: ref})
```
zachdaniel: We need to use the `context_key` variable here and make sure we aren't hardcoding it.

lib/ash/actions/update/bulk.ex
```diff
@@ -404,14 +404,9 @@ defmodule Ash.Actions.Update.Bulk do
...
 
-    {context_key, metadata_key} =
-      case action.type do
-        :update ->
-          {:bulk_update, :bulk_update_index}
-
-        :destroy ->
-          {:bulk_destroy, :bulk_destroy_index}
-      end
+    context_key = :bulk_update
```
zachdaniel: We still need to switch on `action.type` so that this uses the action type for these three values. For destroys the context key, metadata key, and ref_metadata_key need to be based on the action type.

lib/ash/actions/update/bulk.ex
DGollings: makes sense, changed to

```diff
-    |> Ash.Changeset.put_context(:bulk_update, %{index: index, ref: ref})
+    |> Ash.Changeset.put_context(context_key, %{index: index, ref: ref})
```

lib/ash/actions/update/bulk.ex
DGollings: this one I was less sure about, but I think you mean:

```diff
-    context_key = :bulk_update
-    metadata_key = :bulk_update_index
-    ref_metadata_key = :bulk_action_ref
+    {context_key, metadata_key, ref_metadata_key} =
+      case action.type do
+        :destroy ->
+          {:bulk_destroy, :bulk_destroy_index, :bulk_action_ref}
+
+        _ ->
+          {:bulk_update, :bulk_update_index, :bulk_action_ref}
+      end
```

Correct?

DGollings:
Updated @zachdaniel ,  all tests (still) pass (although now I'm wondering if we need an extra test or two?)

lib/ash/actions/update/bulk.ex
```diff
@@ -3453,4 +3545,55 @@ defmodule Ash.Actions.Update.Bulk do
...
+
+  defp ensure_changeset!(changeset, _result, _metadata_key, _ref_metadata_key) do
+    changeset
+  end
+
+  defp result_matches_changes?(result, changes, changesets_by_ref, metadata_key, ref_metadata_key) do
+    if ref_metadata_key do
+      ref_key = result.__metadata__[ref_metadata_key]
+
+      changesets_by_ref
```
zachdaniel: we shouldn't need this check.

zachdaniel:
Feel bad that I keep doing a drive by on this PR and finding small things. 

With the existing data layers currently returning indexes, I'm pretty sure this needs to handle that properly. i.e if we get back a record with a bulk index but not a ref, we should look up the ref by index in a map of index -> ref. Someday down the line we can undo that call. This could be tested at least in the interim by having the ETS data layer only return an index and seeing how it breaks.

DGollings:
And here I am impressed by your restraint, not simply making the necessary changes yourself and merging it in :)

I intentionally had zero backwards compatibility, but makes sense to have some, added in `ea31926c` so it can  be cleanly reverted in the future

zachdaniel:

DGollings:
thought we'd never get this merged :sweat_smile:

PR #2414, opened 2025-10-31
Add preserve_nil_values? to map and struct types

Malian:
This PR ensures `preserve_nil_values? ` is taken into account as a valid option for `constraints` in `Struct` type. It also add the same feature in the `Map` type.

# Doc is still not correct

As discussed with @Torkan in the discord channel, the PR keeps the current behavior by setting the option to `false` by default. That being said, the current doc is not correct (I copied it from the Struct type) because keys with `nil` value are always stored but the docs says the otherwise. 

### Current behavior:
```elixir
%{foo: nil}
# is stored as
{'foo': null}
# is casted to 
%{}
```

### Behavior with `preserve_nil_values?: true`
```elixir
%{foo: nil}
# is stored as
{'foo': null}
# is casted to 
%{foo: nil}
```

### Suggestion
Update the doc to says that, for not breaking change policy, the keys will `nil` value are stored as is, but are not taken into account when casting back to Elixir.

or

Fix the behavior according to the doc, introducing a breaking change in how the value are stored in the database. I'm not sure if this should be considered a breaking change, given that Ash's map or struct values casted from database will not change, only the values stored in the database will change. 

# How to consider data that have already been stored?

Ultimately, whatever we do, we may be introducing a breaking change. Indeed, people who already use map and struct have data with keys with `nil` value stored in the database. If those people add the `preserve_nil_values?: true` option, the data already in the database will now be taken into account when casting to a map or struct potentially overriding another value.

# Contributor checklist

- [x] Documentation changes

### DIFF
--- a/test/type/map_test.exs
+++ b/test/type/map_test.exs
@@ -76,8 +76,16 @@ defmodule Ash.Type.MapTest do
     assert {:ok, %{}} = Ash.Type.cast_stored(Ash.Type.Map, %{"foo" => nil}, constraints)
   end
 
+  test "cast_input casts keys with nil values by default" do
+    {:ok, constraints} = Ash.Type.init(Ash.Type.Map, fields: [foo: [type: :string]])
+
+    assert {:ok, %{"foo" => nil}} =
+             Ash.Type.cast_input(Ash.Type.Map, %{"foo" => nil}, constraints)
+  end
+
   test "cast_stored casts keys with nil values when preserve_nil_values? is true" do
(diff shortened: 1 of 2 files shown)


zachdaniel:
Mind running the formatter?

Malian:
@zachdaniel Done! I also fixed a test in the `type_test.exs` file that was failing because a new key had been added and the error message was no longer valid.

zachdaniel (approved):

zachdaniel:

Issue #2372, opened 2025-10-11
Support bypass multi tenancy for aggregates and calculations


shahryarjb:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

No Support for Revaluation or Bypassing Multi-Tenancy in Aggregates and Calculations

Sometimes we need to bypass `multitenancy`, even if we use `Ash.load` or add `read_action :read_any` and add a new tendency id to it, it dose not accept.
```elixir
 aggregates do
    custom :media_count, :media, :integer do
      public? true
      implementation MishkaCms.Runtime.Aggregates.MediaCountAggregate
      description "Returns total number of media files in this category using custom aggregate that bypasses multitenancy"
    end
  end
```

### Describe the solution you'd like

If it accepts from parent
```elixir
Ash.load!(category, [:media_count],
  tenant: media_query_tenant,
  authorize?: false
)
```
And something like `multitenancy :bypass`, fix the problem

### Refs
- https://discord.com/channels/711271361523351632/1426476980274200577

zachdaniel:
This would be a welcome change, PRs welcome!

shahryarjb:
@zachdaniel it should read `multitenancy` from its parent action? or it should have `multitenancy bypass` inside itself?

zachdaniel:
Well calculations can come from any action so they'd need to have their own config.

PR #2427, opened 2025-11-07
feat(aggregates): add multitenancy bypass option for aggregates

shahryarjb:
Based on #2372, it covers the aggregates for know, I will try for calculations

##### I covered these test cases

- [x] Action WITH bypass + Aggregate WITH bypass
- [x] Action WITHOUT bypass + Aggregates WITH and WITHOUT bypass
- [x] Mixed aggregates (bypass and non-bypass) with regular action
- [x] Action WITH bypass + Aggregate WITHOUT bypass
- [x] Without tenant in opts (return error `This is the limit of bypass actions; it should still error`)

> Please review carefully; I might have missed something. Many parts were involved, but I tried to keep codebase changes minimal. All **Ash tests** passed successfully. In my own project, cases where I previously had to use **raw queries** in calculations are now easily handled with **aggregates**. Also, many of my previous **duplications** are now completely removed.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/aggregate.ex
+++ b/lib/ash/actions/aggregate.ex
@@ -81,9 +81,6 @@ defmodule Ash.Actions.Aggregate do
                                     metadata do
             Ash.Tracer.set_metadata(opts[:tracer], :action, metadata)
 
-            # Preserve the original tenant from opts before handle_multitenancy might clear it
-            original_tenant = opts[:tenant] || query.tenant
-
             with {:ok, query} <- Ash.Actions.Read.handle_multitenancy(query),
                  {:ok, %{valid?: true} = query} <-
                    authorize_query(query, opts, agg_authorize?),
@@ -96,11 +93,11 @@ defmodule Ash.Actions.Aggregate do
               results =
                 Enum.reduce_while(
                   [
-                    {bypass_aggs, query.tenant || original_tenant,
+                    {bypass_aggs, query.tenant,
                      Map.merge(query.context || %{}, %{
                        shared: %{private: %{multitenancy: :bypass_all}}
                      })},
-                    {tenant_aggs, query.tenant || original_tenant, query.context}
+                    {tenant_aggs, query.tenant, query.context}
                   ],
                   {:ok, %{}},
                   fn


lib/ash/query/aggregate.ex
```diff
@@ -230,6 +239,19 @@ defmodule Ash.Query.Aggregate do
           build_opts -> build_query(target_resource, resource, build_opts)
         end
 
+      # Set bypass context IMMEDIATELY if multitenancy bypass is requested
+      query =
+        if opts.multitenancy == :bypass do
+          query
+          |> Ash.Query.set_tenant(nil)
```
zachdaniel: There is a separate discussion about not actually clearing the tenant in these scenarios, but instead just skipping actually using the tenant. That way you can see what tenant was bypassed and what wasn't. Lets try that here.

lib/ash/query/aggregate.ex
```diff
@@ -230,6 +239,19 @@ defmodule Ash.Query.Aggregate do
...
+          |> Ash.Query.set_context(%{
+            multitenancy: :bypass_all,
```
zachdaniel: no need to set this key, it will be copied from `shared`

zachdaniel:
This does look generally good, but I think we're going to have to test this in `ash_postgres` using `context` based multi-tenancy as well for resources that support global as well 😓 That is a bit harder, but it will have to affect the way that those join.

lib/ash/query/aggregate.ex
shahryarjb: Done 🙏🏻 i think i got it from read action

lib/ash/query/aggregate.ex
shahryarjb: @zachdaniel I deleted the `Ash.Query.set_tenant(nil)`, we do not need it, because i implemented  here to know what aggregations are bypass what are not

https://github.com/ash-project/ash/pull/2427/files#diff-43bd00d31554df0f454a25914f2291b8e66933c876900039fecfaf5d0e3345e2R91-R143

I just wanted to follow your codebase, but we do not need it here 🙏🏻

shahryarjb:
@zachdaniel 
> This does look generally good, but I think we're going to have to test this in `ash_postgres` using `context` based multi-tenancy as well for resources that support global as well 😓 That is a bit harder, but it will have to affect the way that those join.

Sure 🙏🏻 please tell me what must do, i will try it
By the way, I don’t have a context-based model in my codebase.
My CMS project has around 4k tests, most of which use `global: false`, as well as attributes.
I tested it there together with `AshJson`.
Could you please clarify exactly what I need to do and where I should do it?

I did some changes you told me please re-check it 🙏🏻♥️; By the way i tested the `ash_postgres` without adding any new tests with new Ash options i added and it has just 2 errors in `mix test`, but i think these are not about this update! 🤔

### For example
```
  1) test aggregate with parent filter and limited select FAILS when combining select() + limit() with aggregate using parent() in filter (AshSql.AggregateTest)
     test/aggregate_test.exs:1865
     ** (Ash.Error.Unknown)
     Bread Crumbs:
       > Error returned from: AshPostgres.Test.Chat.read

     Unknown Error

     * ** (Postgrex.Error) ERROR 42703 (undefined_column) column s0.last_read_message_id does not exist
 ```
 And
 ```
 2) test calculations that refer to aggregates can be authorized (AshPostgres.CalculationTest)
     test/calculation_test.exs:471
     match (=) failed
     code:  assert %{has_future_comment: true} =
              Post
              |> Ash.Query.load([:has_future_comment, :latest_comment_created_at])
              |> Ash.Query.for_read(:allow_any, %{})
              |> Ash.read_one!(authorize?: false)
     left:  %{has_future_comment: true}
     right: %AshPostgres.Test.Post{
              has_future_comment: false,
```
Thanks!

zachdaniel:
@shahryarjb there wouldn't be any failing tests yet, because `ash_postgres` tests wouldn't be using this option. A new test would need to be added to verify that this option works (I don't think it will).

shahryarjb:
Hi dear @zachdaniel i think my PR in the Ash Postgres needs so much attention, its required-database knowledge is more than me 🥲

PR: https://github.com/ash-project/ash_postgres/pull/649 

Thank you in advance

zachdaniel:
@shahryarjb that PR makes me think that we should just make it so that we only allow this for attribute multitenancy, and that is a change we can make here. WDYT?

shahryarjb:
Hi dear @zachdaniel i put a comment in the ash postgres, I agree with you as well. ♥️🙏🏻
https://github.com/ash-project/ash_postgres/pull/649#discussion_r2529652330

zachdaniel:
Excellent. So with that in mind, what we need to do then is validate when building an aggregate that *all resources along the relationship path* either have no multitenancy, or have attribute multitenancy. If there is context multitenancy anywhere then we can just raise an error.

shahryarjb:
> Excellent. So with that in mind, what we need to do then is validate when building an aggregate that *all resources along the relationship path* either have no multitenancy, or have attribute multitenancy. If there is context multitenancy anywhere then we can just raise an error.

@zachdaniel this condtion should be in this pr or should be ash postgres? Hamm it is an easy condition or some think you have in your mind?

zachdaniel:
@shahryarjb yeah it should be here in this PR 😄 can go somewhere around here: https://github.com/ash-project/ash/pull/2427/files#diff-b7b7a5528d3683c2b41b27b34431108ab641369372ec60ac1c908fa81beb04ffR2792

You should be able to look at each resource along the `aggregate.relationship_path` and check its multitenancy to ensure its not set to `:context`.

shahryarjb:
Hi dear @zachdaniel , could you please re-check my last 2 updates, if is there any consideration i should care please let me know to update!
By the way i decided to have `raise Ash.Error.Query.InvalidQuery` and force user not to use bypass for context even not using `!` in Ash function
Thank you in advance

lib/ash/actions/read/read.ex
```diff
@@ -2785,11 +2785,49 @@ defmodule Ash.Actions.Read do
...
+      end
+
+      aggregate_query =
+        if aggregate.multitenancy == :bypass do
+          aggregate.query
+          |> Ash.Query.set_context(%{
+            shared: %{multitenancy: :bypass_all}
+          })
+        else
+          Ash.Query.set_tenant(aggregate.query, aggregate.query.tenant || query.tenant)
```
zachdaniel: We should set the tenant in both cases, but just set the shared context only in the case we want to bypass multitenancy for.

Also, lets use `shared: %{private: %{multitenancy: :bypass_all}}` perhaps?

lib/ash/actions/read/read.ex
```diff
@@ -2785,11 +2785,49 @@ defmodule Ash.Actions.Read do
...
   end
 
+  # Validate context multitenancy and its relationships are not used with bypass
+  # Ref: https://github.com/ash-project/ash_postgres/pull/649#issuecomment-3536654583
   defp handle_aggregate_multitenancy(query) do
+    validate_context_strategy! = fn resource, relationship_name, aggregate_name ->
+      if Ash.Resource.Info.multitenancy_strategy(resource) == :context do
+        location = relationship_name && " in relationship `#{relationship_name}`"
+
+        raise Ash.Error.Query.InvalidQuery,
```
zachdaniel: No need to raise, lets just add this error to the query.

lib/ash/actions/read/read.ex
shahryarjb: @zachdaniel 
  Old behavior: When bypassing multitenancy, the tenant value was lost completely.

  New behavior: The tenant is always preserved. Bypassing just tells the system "don't filter by tenant" but still
  keeps the tenant value available for logging, auditing, or other uses.
  
  Based on this: https://github.com/ash-project/ash/pull/2427/commits/da6907d621934d174e7321e1464107b9a0cc52d2
  
  Hope it is as you told me

lib/ash/actions/read/read.ex
shahryarjb: @zachdaniel I refactored some lines of the code to support error tupple based on https://github.com/ash-project/ash/pull/2427/commits/cf19f84d1e0422d0505f5881c326daa2aff6855a
And i tested and all tests were okey

---

By the way I wish there were some coding-style guidelines, like how I should push code to the repo. There are many coding styles I prefer, but I think they might not be well-received. I'm not talking about performance, just coding style. but i tried to follow the repo style of coding

lib/ash/actions/aggregate.ex
```diff
@@ -81,32 +81,65 @@ defmodule Ash.Actions.Aggregate do
                                     metadata do
             Ash.Tracer.set_metadata(opts[:tracer], :action, metadata)
 
+            # Preserve the original tenant from opts before handle_multitenancy might clear it
+            original_tenant = opts[:tenant] || query.tenant
```
zachdaniel: I believe we updated `Ash.Actions.Read.handle_multitenancy` to not clear the tenant, so we shouldn't need to do this.

lib/ash/actions/read/read.ex
```diff
@@ -2785,19 +2785,70 @@ defmodule Ash.Actions.Read do
     end
   end
 
+  # Validate context multitenancy and its relationships are not used with bypass
+  # Ref: https://github.com/ash-project/ash_postgres/pull/649#issuecomment-3536654583
   defp handle_aggregate_multitenancy(query) do
+    validate_context_strategy = fn resource, relationship_name, aggregate_name ->
```
zachdaniel: This should just be a private function, no need to create an anonymous function here.

lib/ash/actions/read/read.ex
```diff
@@ -2785,19 +2785,70 @@ defmodule Ash.Actions.Read do
...
+          :ok
+        end
 
-        {:error, error} ->
-          {:halt, {:error, Ash.Error.set_path(error, aggregate.name)}}
+      aggregate_query =
+        aggregate.query
+        |> Ash.Query.set_tenant(aggregate.query.tenant || query.tenant)
+        |> then(
+          &if(aggregate.multitenancy == :bypass,
```
zachdaniel: Not a big fan of `then(&if`, I think we should clean this up by defining a private function.

lib/ash/actions/read/read.ex
```diff
@@ -2785,19 +2785,70 @@ defmodule Ash.Actions.Read do
...
+            do: Ash.Query.set_context(&1, %{shared: %{private: %{multitenancy: :bypass_all}}}),
+            else: &1
+          )
+        )
+
+      with :ok <- validation,
```
zachdaniel: lets put the logic for this in a private fun, so its `with :ok <- validate_multitenancy(....`

lib/ash/query/aggregate.ex
```diff
@@ -230,6 +239,17 @@ defmodule Ash.Query.Aggregate do
           build_opts -> build_query(target_resource, resource, build_opts)
         end
 
+      # Set bypass context IMMEDIATELY if multitenancy bypass is requested
```
zachdaniel: No need for the comment. I'd look at each comment to see if its truly necessary.

zachdaniel (changes requested):
See my comments. I think we're almost there 🥳

lib/ash/query/aggregate.ex
shahryarjb: Done ✅

lib/ash/actions/read/read.ex
shahryarjb: Done ✅

lib/ash/actions/read/read.ex
shahryarjb: Done ✅

lib/ash/actions/read/read.ex
shahryarjb: Done ✅

lib/ash/actions/aggregate.ex
shahryarjb: I re-checked and as you said it works, but it did not work before and force me to keep original and use it inside the code
By the way thank you
Done ✅

shahryarjb:
Hi dear @zachdaniel, I did all changes you told me! and hope they will be okey 🧙‍♂️
Thank you in advance for re-checking the PR and last changes ♥️🙏🏻
It passed my tests (in my project) and the Ash tests  ☃️

zachdaniel:

Issue #2433, opened 2025-11-12
Code interface documentation trims double newlines


Munksgaard:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Ash 3.8.0
  

### Operating system

NixOS

### Current Behavior

When writing documentation for my calculate functions (including code interfaces), I like to include a single-sentence first paragraph, followed by a more detailed description. However, when using code interfaces, paragraphs are collapsed, due to calls to `Ash.CodeInterface.trim_double_newlines/1` e.g. [here](https://github.com/ash-project/ash/blob/9c580b9693cd200a0927d24ab0ead2953c8eec2e/lib/ash/code_interface.ex#L534).

It means that the generated documentation looks worse than it should.

### Reproduction

Example to recreate:

```elixir
# in resource
calculations do
  calculate :foo, :string, expr(...) do
    description """
      A foo that frobnicates the resource.

      Here is a detailed description of how the frobnication works.
      """
  end
end

# in domain
resources do
  resource Foo do
    define_calculation :foo, args: [:_record]
  end
end
```

When inspecting the documentation by calling `mix docs`, the documentation is collapsed to:

> A foo that frobnicates the resource. Here is a detailed description of how the frobnication works.

### Expected Behavior

> A foo that frobnicates the resource. 
>
> Here is a detailed description of how the frobnication works.

Munksgaard:
Furthermore, removing double newlines breaks doctests.

zachdaniel:
I think we should actually just update it to remove *triple* newlines 😄

Munksgaard:
> I think we should actually just update it to remove _triple_ newlines 😄

Sounds reasonable! I've updated the PR :-)

zachdaniel:
Fixed here: 086e8004f0d3ad259ac749ce27abe7dd5e02ad0b

zachdaniel:
Thank you for your contribution!

PR #2434, opened 2025-11-12
fix: Trim triple newlines instead of double newlines

Munksgaard:
Removing double newlines breaks formatting in e.g. the generated browser docs,
and breaks doctests, since they expect the example code sections to be separated
by the rest of the comments by double newlines.

Instead, we should focus on collapsing triple newlines, which can occur for
legitimate reasons, e.g. a section of the generated documentation doesn't return
any text, but which shouldn't be showed in the resulting documentation.

Fixes: #2433

I tried adding a regression test to `code_interface_test.exs` by adding the following, but it didn't work because `Code.fetch_docs` returned `{:error, :module_not_found}`:

```
  defmodule CalculationDocumentation do
    use Ash.Resource,
      domain: Ash.Test.Domain

    attributes do
      attribute :foo, :string do
        public? true
      end
    end

    actions do
      defaults create: :*
    end

    calculations do
      calculate :foo_calculation, :boolean, expr(true) do
        description """
        Calculate a foo.

        This is hard to do.
        """
      end
    end

    code_interface do
      define_calculation :foo_calculation
    end
  end

  describe "documentation" do
    test "calculation code interface doc string doesn't strip double newlines" do
      generated_doc =
        Code.fetch_docs(CalculationDocumentation)
        |> dbg
        |> elem(6)
        |> Enum.find(&(elem(&1, 0) == {:function, :foo, 1}))
        |> elem(3)
        |> Map.fetch!("en")
        |> dbg

      assert String.contains?(generated_doc, """
             Calculate a foo.

             This is hard to do.
             """)
    end
  end
```

Any suggestions for a better way to test this would be greatly appreciated.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Edit: Updated to reflect this comment: https://github.com/ash-project/ash/issues/2433#issuecomment-3535480094

### DIFF
--- a/lib/ash/code_interface.ex
+++ b/lib/ash/code_interface.ex
@@ -465,7 +465,7 @@ defmodule Ash.CodeInterface do
 
              #{interface_options.docs()}
              """
-             |> Ash.CodeInterface.trim_double_newlines()
+             |> Ash.CodeInterface.trim_triple_newlines()
         @doc spark_opts: [
                {opts_location, interface_options.schema()}
              ]
@@ -531,7 +531,7 @@ defmodule Ash.CodeInterface do
 
              #{interface_options.docs()}
              """
-             |> Ash.CodeInterface.trim_double_newlines()
+             |> Ash.CodeInterface.trim_triple_newlines()
         @doc spark_opts: [
                {opts_location, interface_options.schema()}
              ]
@@ -1651,7 +1651,7 @@ defmodule Ash.CodeInterface do
 
              #{interface_options.docs()}
              """
-             |> Ash.CodeInterface.trim_double_newlines()
+             |> Ash.CodeInterface.trim_triple_newlines()
 
         @doc spark_opts: [
                {first_opts_location, interface_options.schema()},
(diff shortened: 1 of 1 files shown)


lib/ash/code_interface.ex
```diff
@@ -2299,13 +2292,6 @@ defmodule Ash.CodeInterface do
     end
   end
 
-  @doc false
-  def trim_double_newlines(str) do
```
zachdaniel, line 2303: We can't remove this right now because `AshPhoenix` is using it. Also I see a formatting error, is that from these changes or in main? Either way, will merge once this is added back in 😄

Munksgaard: I got to the same realization when I went to sleep last night, sorry about that!

The formatting changes are caused by my changes, though there was also a formatting error in `actions/destroy/bulk.ex` which I've fixed in a separate commit.

Munksgaard: See also my comment in https://github.com/ash-project/ash_phoenix/issues/439#issuecomment-3525943025

Munksgaard:
Huh, the formatter in the CI behaves differently from my local install, with regards to that bulk destroy change. I'll remove that commit again.

zachdaniel:
Thanks for pointing out the issues here! I'm just going to update the `trim_double_newlines` function to trim triple newlines. Its not a public function (`@doc false`) so it really doesn't matter how accurate the name is 😄

zachdaniel:
This will then automatically resolve the issue in `AshPhoenix` as well 😄

Munksgaard:
Thanks!

PR #2436, opened 2025-11-12
test: add test to show calculation bug with relationships

barnabasJ:
Adds tests to show problems with calculations.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/calculations/calculation_with_load_read_action_test.exs
+++ b/test/calculations/calculation_with_load_read_action_test.exs
@@ -1,0 +1,108 @@
+# SPDX-FileCopyrightText: 2019 ash contributors <https://github.com/ash-project/ash/graphs.contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Ash.Test.CalculationWithLoadReadActionTest do
+  @moduledoc false
+  use ExUnit.Case, async: true
+
+  defmodule Related do
+    use Ash.Resource,
+      data_layer: Ash.DataLayer.Ets,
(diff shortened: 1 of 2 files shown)


test/calculations/calculation_with_load_read_action_test.exs
```diff
@@ -0,0 +1,108 @@
...
+      uuid_primary_key(:id)
+    end
+
+    calculations do
+      calculate :calculated_common_name,
+                :string,
+                expr(related.name),
+                load: [
+                  related: Ash.Query.for_read(Related, :read_all)
+                ]
```
barnabasJ, line 68: Not 100% sure if this is a bug or a documentation issue

test/calculations/calculation_with_load_read_action_test.exs
```diff
@@ -0,0 +1,108 @@
...
+    attributes do
+      uuid_primary_key(:id)
+    end
+
+    calculations do
+      calculate :calculated_common_name,
+                :string,
+                expr(related.name),
+                load: [
+                  related: Ash.Query.for_read(Related, :read_all)
```
zachdaniel, line 67: Expression calculations don't use the `load` statement, it all happens in the database, so I wouldn't call this a bug. It is pretty confusing though, and perhaps for a calculation that *only* has an expression, we should consider adding a verifier that tells you what I just said.

PR #2462, opened 2025-12-02
fix: handle empty list in Ash.Filter.parse_and_join/3

DohanKim:
## Summary

  - Fix `FunctionClauseError` crash when using `Ash.Query.filter(or: [])` or `Ash.Query.filter(and: [])`
  - Add clause to handle empty lists by returning `{:ok, true}` (identity element for boolean operations)

  ## Root Cause

  `Ash.Filter.parse_and_join/3` didn't have a clause to handle empty lists, only:
  - Non-empty lists: `[statement | statements]`
  - Maps: `when is_map(statement) and not is_struct(statement)`

  ## Real-world Scenario

  This occurs when using `Ash.bulk_create/4` with AshPostgres data layer, `upsert?: true` and `return_skipped_upsert?: true`. When all
  records are successfully inserted (none are skipped due to upsert conditions), [`AshPostgres.DataLayer.bulk_create/3`](https://github.com/ash-project/ash_postgres/blob/cb2c8cbbeea32822c37f4ea7ba072ad081d48032/lib/data_layer.ex#L2065) builds a filter with `or: []` to fetch
   skipped records, which triggers the crash:

```
  ** (FunctionClauseError) no function clause matching in Ash.Filter.parse_and_join/3
      (ash 3.9.0) lib/ash/filter/filter.ex:4711: Ash.Filter.parse_and_join([], :or, %{...})
```

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -4708,6 +4708,8 @@ defmodule Ash.Filter do
 
   defp add_to_ref_path(other, _), do: other
 
+  defp parse_and_join([], _op, _context), do: {:ok, true}
+
   defp parse_and_join([statement | statements], op, context) do
     case parse_expression(statement, context) do
       {:ok, nested_expression} ->
--- a/test/filter/filter_test.exs
+++ b/test/filter/filter_test.exs
@@ -1213,4 +1213,40 @@ defmodule Ash.Test.Filter.FilterTest do
              |> Ash.Query.filter_input(%{embedded_bio: %{at_path: [:title], eq: "Dr."}})
              |> Ash.read!()
   end
+
+  describe "empty filter lists" do
+    setup do
+      post1 =
+        Post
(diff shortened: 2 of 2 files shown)


zachdaniel:
I don't think this is quite right. If I say `or: []` and that translates to `true` then that query in bulk create now reads out every row in the database.

I'd suggest that first we should just fix that code not to make a query at all if there are no things to look up, and 2 if we want to make this change, if it's `:or` then it should evaluate to `false`.

DohanKim:
You're right that the code in [AshPostgres.DataLayer.bulk_create/3](https://github.com/ash-project/ash_postgres/blob/cb2c8cbbeea32822c37f4ea7ba072ad081d48032/lib/data_layer.ex#L2065) should be changed to avoid the query entirely when there's nothing to look up.

I'll close this PR and submit a separate one for AshPostgres instead.

Thank you for the great project.

PR #2467, opened 2025-12-05
show that after transaction is not useable with atomics

barnabasJ:


zachdaniel:
@barnabasJ would you mind trying out removing this restriction to see if it works with AshPostgres? By my understanding it *should* be possible to do `after_transaction` calls with atomics, so it may be as easy as removing this restriction and adding some tests for it.

barnabasJ:
> @barnabasJ would you mind trying out removing this restriction to see if it works with AshPostgres? By my understanding it *should* be possible to do `after_transaction` calls with atomics, so it may be as easy as removing this restriction and adding some tests for it.

I'm working on something, I'm trying to understand the whole `atomic_after_action` thing the regular `Ash.Changeset.after_action` and the phases around that to see if we need Todo something like that as well

barnabasJ:
> @barnabasJ would you mind trying out removing this restriction to see if it works with AshPostgres? By my understanding it *should* be possible to do `after_transaction` calls with atomics, so it may be as easy as removing this restriction and adding some tests for it.

I'm working on something, I'm trying to understand the whole `atomic_after_action` thing the regular `Ash.Changeset.after_action` and the phases around that to see if we need Todo something like that as well

Issue #2468, opened 2025-12-10
Cannot use custom NewType in :union argument


zackattackz:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Please see this repo at this specific branch to reproduce the issue https://github.com/zackattackz/PetRockRental/tree/to-arg-union-wip

The versions are in mise.toml and mix.lock but for convenience are here:
```toml
[tools]
elixir = "1.18.4-otp-27"
erlang = "27"
```

```elixir
%{
  "ash": {:hex, :ash, "3.11.1", "9794620bffeb83d1803d92a64e7803f70b57372eb4addba5c12a24343cd04e1a", [:mix], [{:crux, ">= 0.1.2 and < 1.0.0-0", [hex: :crux, repo: "hexpm", optional: false]}, {:decimal, "~> 2.0", [hex: :decimal, repo: "hexpm", optional: false]}, {:ecto, "~> 3.7", [hex: :ecto, repo: "hexpm", optional: false]}, {:ets, "~> 0.8", [hex: :ets, repo: "hexpm", optional: false]}, {:igniter, ">= 0.6.29 and < 1.0.0-0", [hex: :igniter, repo: "hexpm", optional: true]}, {:jason, ">= 1.0.0", [hex: :jason, repo: "hexpm", optional: false]}, {:picosat_elixir, "~> 0.2", [hex: :picosat_elixir, repo: "hexpm", optional: true]}, {:plug, ">= 0.0.0", [hex: :plug, repo: "hexpm", optional: true]}, {:reactor, "~> 0.11", [hex: :reactor, repo: "hexpm", optional: false]}, {:simple_sat, ">= 0.1.1 and < 1.0.0-0", [hex: :simple_sat, repo: "hexpm", optional: true]}, {:spark, ">= 2.3.14 and < 3.0.0-0", [hex: :spark, repo: "hexpm", optional: false]}, {:splode, ">= 0.2.6 and < 1.0.0-0", [hex: :splode, repo: "hexpm", optional: false]}, {:stream_data, "~> 1.0", [hex: :stream_data, repo: "hexpm", optional: false]}, {:telemetry, "~> 1.1", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "e0074302bb88d667635fcbfdacbf8a641c53973a3902d0e744f567a49ec808fc"},
[...]

PR #2469, opened 2025-12-12
fix: dont validate constraints in union_types

zackattackz:
Remove call to `Spark.Options.validate!/2` inside `Ash.Type.Union.union_types/1`. It is unnecessary and it causes an issue when attempting to define unions using NewTypes whose parents have `required: true` in their constraints.

Add in two test cases in the union type's tests:

1. Checks that we can define and use a union with the mentioned definition.
2. Ensures that nothing was lost when we removed the call to `Spark.Options.validate!/2` by demonstrating that the constraints are still checked when defining a resource using a union with the mentioned definition.

These tests fail to compile/run without the fix.

Fixes https://github.com/ash-project/ash/issues/2468

--------

### Extra notes

I believe removing this call to `validate!` is fine. However it did used to modify the constraints by adding in default values, e.g if `type` was `Ash.Type.String`, the call to `validate!` would return constraints with `[trim?: true, allow_empty?: false]` before adding them to the final result from `union_types`. But this seems to not be necessary as when the type is fully initialized it will include those constraints anyways. (I could add a test case to show this if wanted)

The problem with keeping the `validate!` call, is that with NewTypes `validate!` is expecting to receive their applied constraints as well. I think we could do this by calling `Ash.Type.init/2` like so:

```elixir
schema = Ash.Type.constraints(type)

{:ok, constraints} = Ash.Type.init(type, config[:constraints] || [])
constraints = Spark.Options.validate!(constraints, schema)

{key, config |> Keyword.put(:constraints, constraints) |> Keyword.put(:type, type)}
```

But I think it'd just be more preferable to remove the `validate!` call altogether since it seems unnecessary.

I may be missing some considerations there though, so please scrutinize this!

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/union.ex
+++ b/lib/ash/type/union.ex
@@ -149,10 +149,7 @@ defmodule Ash.Type.Union do
          """
        end
 
-       schema = Ash.Type.constraints(type)
-       constraints = Spark.Options.validate!(config[:constraints] || [], schema)
-
-       {key, config |> Keyword.put(:constraints, constraints) |> Keyword.put(:type, type)}
+       {key, config |> Keyword.put(:type, type)}
      end)}
   end
 
--- a/test/type/union_test.exs
+++ b/test/type/union_test.exs
@@ -80,6 +80,89 @@ defmodule Ash.Test.Filter.UnionTest do
       ]
   end
 
+    defmodule DefaultedType do
+    @moduledoc "A type with a default constraint"
+    use Ash.Type
+
+    @constraints [
+      defaulted: [
+        type: :integer,
+        default: 9001
+      ]
+    ]
+
+    @impl true
+    def storage_type(_), do: :binary
+
+    @impl true
+    def constraints, do: @constraints
+
+    @impl true
+    def cast_input(value, _), do: {:ok, value}
+
+    @impl true
+    def cast_stored(value, _), do: {:ok, value}
+
+    @impl true
+    def matches_type?(_, _), do: true
+
+    @impl true
+    def dump_to_native(value, _), do: {:ok, value}
+  end
+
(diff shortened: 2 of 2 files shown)


zackattackz:
I may amend to add a regression test case for the part I mentioned in extra notes regarding the defaults being added to constraints

zachdaniel:
Could you try cloning ash_graphql down and pointing its ash dependency at this branch and running tests? If they pass then I think we're good here.

zackattackz:
Force pushed to rebase over latest main branch, also added in that extra regression test I mentioned before (check that a type with a `default` constraint gets the default value applied properly)

So then I ran the tests in `ash_graphql` and it seems to work.

For sanity check, I ran `mix check` from `ash_graphql` @ f86cd51 without any changes. It yields the logs attached in
[ash_graphql-mix-check-before.log](https://github.com/user-attachments/files/24316560/ash_graphql-mix-check-before.log)

Then in mix.exs I changed the ash dependency to point to my local repo, which was at this branch:
```elixir
 defp deps do
    [
      # {:ash, ash_version("~> 3.5 and >= 3.5.13")},
      {:ash, path: "../ash"},
      ...
```
And ran `mix deps.get`, producing a new 
[mix.lock](https://github.com/user-attachments/files/24316596/mix.lock.txt) (had to rename to .txt to attach)

Finally ran `mix check` again, yielding this successful output 
[ash_graphql-mix-check-after.log](https://github.com/user-attachments/files/24316613/ash_graphql-mix-check-after.log)

zachdaniel:

Issue #2420, opened 2025-11-03
More (user-) fault tolerant install script


janwirth:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

I tried out Ash for the first time, knowing little elixir.
I used the install script on the homepage.
It aborted when it could not connect to postgres.
I got postgres running, ran the migrations with `mix ectom.migrate` and started mix phx.server.
CSS and JS were missing.
Being new to the ecosystem I was unaware that I had to rerun `mix setup`.
I spent half a day debugging this, getting confused by JS and tailwind errors.

### Describe the solution you'd like

A better warning and instructions to rerun `mix setup`.
Better loggin as to which plugin is throwing the error.
Detection if assets were installed or not before running the app build.

### Describe alternatives you've considered

Doing the DB migration / connection last because it's the most 'moving' part. That means if the setup script bail out I can do the last mile easily.

zachdaniel:
The db migration/connection *should* be last in the current setup, but this all makes perfect sense. Making the install script attempt to first check connection to the database and provide a more friendly error makes sense to me, as does doing `mix setup` on a fresh install before that too.

I'll mark this as a good first issue for others to contribute, thanks!

celeste-wahlquist:
I'm planning on 
1. adding additional documentation to get-started.md 
2. creating a function which checks if postgres is running and either a) receives a postgres verification and continues to installation or b) does not receive postgres verification and alerts the user giving them a warning and option to bypass this warning and continue installation or abort installation

am I missing anything?

zachdaniel:
👋 I put some feedback on the PR, but unfortunately the postgres check isn't going to work as proposed. Could live in ash_postgres installation potentially, but even still I think the best answer is to figure out how to make ash and/or igniter tasks failure more gracefully.

For instance, when `mix ash.setup` fails, we can print an error explaining that it failed and should be rerun.
and when an igniter task fails, we can display which tasks didn't get run.

celeste-wahlquist:
2. the failure reporter on running tasks should be built directly into igniter. Could you PR that functionality there instead?
Igniter task solution has been moved to correct repository and pull request has been made 🫡

3. I'd rather not have the tests than add the entire test project into test fixtures TBH. Igniter has some examples in it for testing these things.
[...]

PR #2471, opened 2025-12-15
enhancement: on setup check db connectivity

ken-kost:
#2420 
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/mix/tasks/ash.setup.ex
+++ b/lib/mix/tasks/ash.setup.ex
@@ -12,9 +12,11 @@ defmodule Mix.Tasks.Ash.Setup do
   @doc @shortdoc
   def run(argv) do
     Mix.Task.run("compile")
+    extensions = Ash.Mix.Tasks.Helpers.extensions!(argv)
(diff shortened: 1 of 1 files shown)


zachdaniel:
If we do this it would be via allowing extensions to have their own `setup` function for example. See how we do it with `ash.migrate` that way this logic lives in the correct package :)

ken-kost:
> If we do this it would be via allowing extensions to have their own `setup` function for example. See how we do it with `ash.migrate` that way this logic lives in the correct package :)

```
  def setup(args) do
    # TODO: take args that we care about
    Mix.Task.run("ash_postgres.create", args)
    Mix.Task.run("ash_postgres.migrate", args)

    []
    |> AshPostgres.Mix.Helpers.repos!(args)
    |> Enum.all?(&(not has_tenant_migrations?(&1)))
    |> case do
      true ->
        :ok

      _ ->
        Mix.Task.run("ash_postgres.migrate", ["--tenant" | args])
    end
  end
```
Should I catch errors in create/migrate i.e. if those fail print error in shell? 🤔

PR #2486, opened 2025-12-24
fix: many-to-many relationships should inherit the sort/limit from their join relationship

sevenseacat:
This fixes an issue where a has_many relationship could have a limit applied, but the many-to-many relationship would ignore that limit and instead load the full set of related records

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -651,6 +651,15 @@ defmodule Ash.Actions.Read.Relationships do
             related_query.domain
           )
       )
+      |> Ash.Query.sort(join_relationship.sort)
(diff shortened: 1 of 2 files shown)


sevenseacat:
The current fix seems to only work for the ETS data layer - Postgres takes a route that applies lateral joins and it doesn't yet work that case. Investigating...

zachdaniel:
Yeah...the postgres route may be a bit more complicated 🤔 LMK if you run into any trouble there. Regardless, this PR can be merged as is since it fixes it for ETS.

zachdaniel:

PR #2487, opened 2025-12-24
fix: aggregates to_tenant support

tmpbeing:
In my project I was trying to do an `Ash.count!()` with a struct as `tenant` argument and it would fail with an error telling me the tenant struct doesn't implement `Strings.Chars` protocol, the issue being that `to_tenant` wasn't passed correctly, this fixes that.
I took the liberty to remove that `tenant` variable in the `reduce_while` as it seems it's always `query.tenant` and I couldn't understand why it exists.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/aggregate.ex
+++ b/lib/ash/actions/aggregate.ex
@@ -93,18 +93,18 @@ defmodule Ash.Actions.Aggregate do
               results =
                 Enum.reduce_while(
                   [
-                    {bypass_aggs, query.tenant,
+                    {bypass_aggs,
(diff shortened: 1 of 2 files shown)


tmpbeing:
@zachdaniel anything I can do to move this forward ? (sorry for the ping but I feel like this slipped through because I had the bad idea to push on christmas eve :smile: )

zachdaniel:

(sorry, lost track of this somehow)

PR #2495, opened 2025-12-29
improvement(create): Add multitenancy bypass options to create actions

### LINKED ISSUE #2380: Support bypass multi tenancy for update, create, destroy
### Code of Conduct

### AI Policy

There is a need to support bypassing multi-tenancy for the `update`, `create`, and `destroy` actions. Doing so will allow many generic actions to be removed. It’s important to mention that when tenancy is bypassed, the user can—for example—provide the `site_id` themselves.

Also note that this may not be necessary for `create`, because in `create` we don’t read any existing data and can manually set the `site_id`, for example as follows:

### Create works now like this:
```elixir

    create :create do
      primary? true
      accept List.delete(@admin_fields, :site_id)

      validate present([:name, :slug, :status, :language])
      validate {Validations.TenantContext, require_master: false}
      change &Ash.Changeset.force_change_attribute(&1, :site_id, &2.tenant) # we handle it and no problem unlike the other action
      change MishkaBlog.Resources.Changes.SetLanguageGroupId
      change slugify(:slug, into: :slug, ignore: [".", "!", "?"])

      description "Creates collection for tenant users - requires tenant context, auto-assigns site_id and language_group_id for master collections"
    end
```

#### For example for other actions
```elixir
    update :update do
      multitenancy :bypass
      primary? true
      require_atomic? false
      transaction? true
      accept List.delete(@admin_fields, :site_id)
     ....
    end
```

_No response_

### Additional context

For now `multitenancy :bypass` just active for `read` action
Thank you in advance

shahryarjb:
Hi, After merging this we will have `bypass` for 

- [x] `create`
- [x] `update`
- [x] `destroy` 
- [x] `Aggregates`

the only option should be done is `Calculations` (#2372) and it should be noted I added some tests for bulk operations

- [ ] Calculations #2372

**By the way in another PR, i want to move some duplicated functions to `ash/lib/ash/helpers.ex`, is it okey? or i should keep it like the current code??**

Fixed #2380 

If this PR has problem, please let me know to fix it, Thank you in advance 🙏🏻♥️

---

# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/policy/relates_to_actor_via_test.exs
+++ b/test/policy/relates_to_actor_via_test.exs
@@ -9,7 +9,10 @@ defmodule Ash.Test.Policy.RelatesToActorViaTest do
   alias Ash.Test.Domain, as: Domain
 
   defmodule Actor do
-    use Ash.Resource, domain: Domain, data_layer: Ash.DataLayer.Ets
+    use Ash.Resource,
+      domain: Domain,
+      data_layer: Ash.DataLayer.Ets,
+      authorizers: [Ash.Policy.Authorizer]
 
     actions do
       default_accept :*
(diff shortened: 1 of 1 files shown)


zachdaniel:

Issue #2493, opened 2025-12-29
Add support for SQL `IS DISTINCT FROM` operator


jechol:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

When filtering records, the standard equality operators (`==`, `!=`) treat `NULL` values specially - `NULL == NULL` returns `NULL` (not `true`). This makes it difficult to compare nullable fields where you want `NULL` to equal `NULL`.

For example, when trying to find orders where the shipping phone differs from the primary phone (where `NULL` means "not provided"), using `shipping_phone != primary_phone` won't match cases where both are `NULL`, even though they should be considered equal.

Currently, there's no direct way to use SQL's `IS DISTINCT FROM` operator in Ash filters.

### Describe the solution you'd like

Add `is_distinct_from` and `is_not_distinct_from` operators to Ash's filter DSL:

```
# Find orders where shipping phone differs from primary phone
# (NULL means "not provided", and NULL should equal NULL here)
Order
|> Ash.Query.filter(is_distinct_from(shipping_phone, primary_phone))
|> Ash.read!()
```

### Describe alternatives you've considered

The current workaround requires complex filter logic:

```elixir
# Equivalent to: is_distinct_from(shipping_phone, primary_phone)
filter(
  (is_nil(shipping_phone) and not is_nil(primary_phone)) or
  (not is_nil(shipping_phone) and is_nil(primary_phone)) or
  (shipping_phone != primary_phone)
)
```

### Additional context

This is part of the SQL standard and supported by most major databases including PostgreSQL and SQLite. 

I have a PR ready with implementation across ash, ash_sql, and ash_postgres if this use case makes sense for Ash.

zachdaniel:
I like it 👍 This would simplify a whole host of builtin Ash validations as well actually.

zachdaniel:
We may want to simplify at the expression level to `!=` when the right and left hand side can't return `NULL` though.

PR #2497, opened 2025-12-30
feat: add is_distinct_from and is_not_distinct_from functions

jechol:
Related to ash-project/ash#2493

## Summary

Add `is_distinct_from` and `is_not_distinct_from` functions for NULL-safe comparisons using SQL's `IS DISTINCT FROM` and `IS NOT DISTINCT FROM` operators.

## Use Case

When comparing nullable fields where NULL has semantic meaning (e.g., "not provided"), the standard `!=` operator doesn't match cases where both values are NULL:

#### Before
```elixir
Order
|> Ash.Query.filter(
  (is_nil(shipping_phone) and not is_nil(primary_phone)) or
  (not is_nil(shipping_phone) and is_nil(primary_phone)) or
  (shipping_phone != primary_phone)
)
```

#### After
```elixir
Order
|> Ash.Query.filter(is_distinct_from(shipping_phone, primary_phone))
```

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/query/operator/distinct_from.ex
```diff
@@ -0,0 +1,39 @@
...
+  @impl Ash.Query.Operator
+  def evaluate(%{left: left, right: right}) do
+    {:known, Comp.not_equal?(left, right)}
+  end
+
+  @impl Ash.Query.Operator
+  def evaluate_nil_inputs?, do: true
+
+  @impl Ash.Filter.Predicate
+  def simplify(%__MODULE__{left: left, right: right}) do
```
zachdaniel: This simplification logic is used in a different context (when comparing filters). What we'd want to do here is override `new` instead of `simplify`. The reason you're seeing eager evaluation working in your tests should be that we evaluate expressions automatically when the values are known.

jechol: Done! I've moved the optimization to new() and kept simplify() only in NotDistinctFrom to convert to Not(DistinctFrom).

lib/ash/query/operator/not_distinct_from.ex
```diff
@@ -0,0 +1,44 @@
...
+defmodule Ash.Query.Operator.NotDistinctFrom do
+  @moduledoc """
+  left is_not_distinct_from right
+
+  SQL's IS NOT DISTINCT FROM operator (NULL-safe equality).
+  Unlike `==`, this operator treats NULL as equal to NULL.
+
+  When both sides cannot return NULL, this simplifies to `==` for better performance.
+  """
+  use Ash.Query.Operator,
```
zachdaniel: ah, I didn't even really think about this, but this would be best modeled as an `Ash.Query.Function`, sorry 😓. There is a subtle difference in how we cast types etc. Operators can be "overloaded" whereas functions cannot be. Could we rework this to be a function?

zachdaniel: After that, this looks good to go to me!

jechol: Done! Converted to functions as suggested. Thanks for the review!

zachdaniel (approved):

zachdaniel:

This is a *great* addition 🥳

Issue #2506, opened 2026-01-03
Ash Reactor is wrapping error tuples in Ash.Error.Unknown.UnknownError


katipber:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Erlang/OTP 27 [erts-15.2.7.4] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]

Elixir 1.19.4 (compiled with Erlang/OTP 27)

### Operating system

macOS Sequoia

### Current Behavior

Error tuples within `Ash.Reactor` are getting wrapped in the message of `Ash.Error.Unknown.UnknownError`.

For example an error tuple returned from a dummy step like this one:
```elixir
defmodule Foo.Reactor.Atom do
  use Ash.Reactor

  step :first do
    run(fn _, _ ->
      {:error, :error_atom}
    end)
  end
end
```

Ends up wrapped like this:
```elixir
{:error,
 %Ash.Error.Unknown{
   errors: [
     %Ash.Error.Unknown.UnknownError{
[...]

PR #2507, opened 2026-01-04
fix: unwrap reactor errors

katipber:
This is to unwrap Reactor errors at the end of an action. Relates to #2506 

Before the change errors were wrapped both in `Ash.Error` and `Reactor.Error`, which seems redundant.
```elixir
{:error, %Ash.Error.Invalid{
   errors: [
     %Reactor.Error.Invalid{
       errors: [
         %Reactor.Error.Invalid.RunStepError{...}
      ]
    }
  ]
}}
```

After the change:
```elixir
{:error, %Ash.Error.Invalid{
  errors: [
    %Reactor.Error.Invalid.RunStepError{...}
  ]
}}
```

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/actions/generic_actions_test.exs
+++ b/test/actions/generic_actions_test.exs
@@ -822,6 +822,50 @@ defmodule Ash.Test.Actions.GenericActionsTest do
     test "it does not require setting optional inputs" do
       assert {:ok, "Marty"} = EchoResource.echo2("Marty")
     end
+
+    defmodule FailingReactor do
+      @moduledoc false
+      use Reactor
+
+      input :input
+
+      step :fail do
+        argument :input, input(:input)
+
+        run fn _ ->
+          {:error, "reactor step failed"}
+        end
+      end
+    end
+
+    defmodule FailingReactorResource do
+      @moduledoc false
+      use Ash.Resource, domain: Domain
+
+      actions do
+        action :fail, :string do
+          argument :input, :string, allow_nil?: false
+
+          run FailingReactor
+        end
+      end
+
+      code_interface do
+        define :fail, args: [:input]
+      end
+    end
+
+    test "reactor errors are unwrapped and not nested redundantly" do
+      assert {:error, error} = FailingReactorResource.fail("test")
+
+      # Verify erros are wrapped in Ash.Error.Invalid
+      assert %Ash.Error.Invalid{errors: errors} = error
+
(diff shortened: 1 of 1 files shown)


zachdaniel:
This shouldn't be necessary after the reactor update is released. Have you tried with `main` of reactor?

katipber:
> This shouldn't be necessary after the reactor update is released. Have you tried with `main` of reactor?

I tried both with `main` and `fix-error-classes` branches, the only difference is in `main` the outermost Ash error class is `Ash.Error.Unknown`. In both branches, without this change errors are wrapped twice.

With Reactor main:
```elixir
%Ash.Error.Unknown{
              errors: [%Reactor.Error.Invalid{errors: [%Reactor.Error.Invalid.RunStepError{...
```

I think the reason is Reactor already does wrapping, similar to Ash, then this wrapped error is passed to Ash error handlers. and wrapped again with `Ash.Error.Unknown` (`Ash.Error.Invalid` after the error class fixes).

Two more things to consider:
1. I also tried the errors with option `fail_on_not_found?` with unwraping reactor errors applied, and the error look like this:
```elixir
{:error,
 %Ash.Error.Invalid{
   errors: [
     %Reactor.Error.Invalid.RunStepError{
       error: %Ash.Error.Invalid{
         errors: [
           %Ash.Error.Query.NotFound{...}
         ]
       }...
```
Here you can see within `Reactor.Error.Invalid.RunStepError` there is another nesting, `Ash.Error.Invalid` -> `Ash.Error.Query.NotFound`. I'm not sure if this is the expected behaviour but maybe it would be easier for error handling if it was not wrapped with `Ash.Error.Invalid` within `Reactor.Error.Invalid.RunStepError`.

2. Reactor main breaks this test: `test generic actions wrapping a reactor it does not require setting optional inputs`:
```bash
     code:  assert {:ok, "Marty"} = EchoResource.echo2("Marty")
     left:  {:ok, "Marty"}
     right: {
              :error,
              %Ash.Error.Unknown{errors: [%Ash.Error.Unknown.UnknownError{error: "** (ArgumentError) Reactor is missing the following inputs; `:something_else`", field: nil, value: nil, splode: Ash.Error, bread_crumbs: [], vars: [], path: [], stacktrace: #Splode.Stacktrace<>, class: :unknown}]}
            }
     stacktrace:
       test/actions/generic_actions_test.exs:823: (test)
```

PR #2510, opened 2026-01-06
Atomic after transaction

barnabasJ:
# Contributor checklist

  Enable after_transaction hooks to work with atomic bulk operations (create, update, destroy).

  This PR ensures after_transaction hooks run correctly after the transaction commits, regardless of execution strategy (stream, atomic, atomic_batches).

  Key changes:
  - Add atomic_after_transaction field to changesets to preserve hooks during atomic upgrades
  - Refactor bulk operations to use inline error tuples instead of store_error pattern
  - Run after_transaction hooks outside transaction boundaries in process_results
  - Support hooks that convert errors to successes (and vice versa)

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/helpers/bulk.ex
+++ b/lib/ash/actions/helpers/bulk.ex
@@ -87,32 +87,32 @@ defmodule Ash.Actions.Helpers.Bulk do
 
   Uses pattern matching to extract index/ref from changeset context.
   """
-  def put_bulk_metadata(result, %{context: %{bulk_create: %{index: index, ref: ref}}}) do
+  def put_metadata(result, %{context: %{bulk_create: %{index: index, ref: ref}}}) do
     result
     |> Ash.Resource.put_metadata(:bulk_create_index, index)
     |> maybe_put_ref_metadata(ref, :bulk_action_ref)
   end
 
-  def put_bulk_metadata(result, %{context: %{bulk_destroy: %{index: index, ref: ref}}}) do
+  def put_metadata(result, %{context: %{bulk_destroy: %{index: index, ref: ref}}}) do
     result
     |> Ash.Resource.put_metadata(:bulk_destroy_index, index)
     |> maybe_put_ref_metadata(ref, :bulk_action_ref)
   end
 
-  def put_bulk_metadata(result, %{context: %{bulk_update: %{index: index, ref: ref}}}) do
+  def put_metadata(result, %{context: %{bulk_update: %{index: index, ref: ref}}}) do
     result
     |> Ash.Resource.put_metadata(:bulk_update_index, index)
     |> maybe_put_ref_metadata(ref, :bulk_action_ref)
   end
 
-  def put_bulk_metadata(result, _changeset), do: result
+  def put_metadata(result, _changeset), do: result
 
   @doc """
   Sets bulk operation metadata with explicit values (for batch operations).
 
   Used when index/ref are already extracted as separate variables.
   """
-  def put_bulk_metadata(result, index, ref, index_key) when is_integer(index) do
+  def put_metadata(result, index, ref, index_key) when is_integer(index) do
     result
     |> Ash.Resource.put_metadata(index_key, index)
     |> maybe_put_ref_metadata(ref, :bulk_action_ref)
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -43,4 +43,7 @@ shell.nix
 .elixir-tools
 
 # Personal notes and planning documents
-# notes/
+notes/
+
+# Scratch files
+test/scratch_test.exs
```
zachdaniel: Can we remove these (even the commented out one)? I think these should be in folks personal gitignore files.

lib/ash/actions/helpers/bulk.ex
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: 2019 ash contributors <https://github.com/ash-project/ash/graphs.contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Ash.Actions.Helpers.Bulk do
```
zachdaniel, line 5: Seems like these could probably just be in `actions/helpers.ex` with `bulk` in their name?

lib/ash/actions/update/bulk.ex
```diff
@@ -945,18 +896,46 @@ defmodule Ash.Actions.Update.Bulk do
...
+        {:error,
+         Ash.Error.to_error_class(error,
+           bread_crumbs: [
+             "Returned from bulk query update: #{inspect(atomic_changeset.resource)}.#{atomic_changeset.action.name}"
+           ]
+         )}
+    end
+  end
+
+  # Helper for notification dispatch (extracted from inline code for reuse)
```
zachdaniel: Seems like an unnecessary comment.

lib/ash/actions/update/bulk.ex
```diff
@@ -1473,13 +1459,17 @@ defmodule Ash.Actions.Update.Bulk do
...
       try do
-        records =
-          if opts[:return_records?] do
-            Enum.to_list(Stream.concat(changeset_stream))
-          else
-            Stream.run(changeset_stream)
-            nil
-          end
+        # Collect all results (records and {:error, error} tuples)
+        all_results = Enum.to_list(Stream.concat(changeset_stream))
```
zachdaniel: No need to `Enum.to_list`. That will materialize into a list and then we later split. Can just pass the stream into `Enum.split_with`.

lib/ash/actions/update/bulk.ex
```diff
@@ -2706,26 +2757,11 @@ defmodule Ash.Actions.Update.Bulk do
...
-          |> find_changeset(
-            changesets_by_ref,
-            metadata_key,
-            ref_metadata_key,
-            changesets_by_index
-          )
-          |> ensure_changeset!(result, metadata_key, ref_metadata_key)
-
+  defp run_after_action_hooks(batch_results, opts, domain, ref) do
+    # batch_results is now list of {:ok, result, changeset} | {:error, error, changeset}
```
zachdaniel: Seems like an unneeded comment? Or like related to the change, not the future state.

lib/ash/actions/update/bulk.ex
```diff
@@ -2758,85 +2793,166 @@ defmodule Ash.Actions.Update.Bulk do
...
-      base_changeset.resource,
-      metadata_key,
-      ref_metadata_key
-    )
-    |> Enum.flat_map(fn result ->
-      changeset =
-        result
-        |> find_changeset(changesets_by_ref, metadata_key, ref_metadata_key, changesets_by_index)
-        |> ensure_changeset!(result, metadata_key, ref_metadata_key)
+    # tagged_results is now list of {:ok, result, changeset} | {:error, error, changeset}
```
zachdaniel: Same here. The comment in general is alright but has a temporal bit.

lib/ash/bulk_result.ex
```diff
@@ -22,4 +22,28 @@ defmodule Ash.BulkResult do
...
   ]
+
+  @doc """
+  Recalculates the status of a bulk result based on its records and error_count.
+
+  - `:success` - when there are no errors (error_count == 0)
+  - `:error` - when there are only errors (records is empty or nil, error_count > 0)
+  - `:partial_success` - when there are both successful records and errors
+  """
+  @spec recalculate_status(t()) :: t()
```
zachdaniel: I don't think this is used is it?

zachdaniel:
Hot damn this is awesome. A couple small comments but that's all ❤️

lib/ash/actions/helpers/bulk.ex
barnabasJ: I personally prefer having the differentiator in the module name instead of the function name and the helper module already has 1000 lines and working on the code, there is more potential to extract bulk helpers, into the shared bulk helper module. 

But I can move it if you have a strong preference the other way

zachdaniel: Hmm... I would be down for this, but I think it would then be important to ensure that the *other* bulk related helpers are moved into this file 😄. Otherwise its a bit misleading placement.

barnabasJ: I did another pass and moved the bulk specific functions into the new helper module

zachdaniel:

PR #2516, opened 2026-01-13
perf: Re-enables regex as mfa to improve regex speed

sezaru:
This PR basically partially reverts (partially because some of the old code doesn't make sense to move to the new one) this commit: https://github.com/ash-project/ash/commit/d77537dd03f7e31dc7a127a415a0a935be15043c

The reason is so that spark will cache the regex (even if you don't use mfa and just use regex directly). Without this cache, `:re.import` will be called multiple times during a cast of a resource with regexes matches in its attributes, resulting in a very expressive performance loss.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/validation.ex
+++ b/lib/ash/resource/validation.ex
@@ -207,6 +207,17 @@ defmodule Ash.Resource.Validation do
 
   @doc false
   def transform(%{validation: {module, opts}} = validation) do
+    opts =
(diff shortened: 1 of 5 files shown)


zachdaniel:

PR #2517, opened 2026-01-13
fix: set accessing_from context when reading action during relationship creation

skanderm:
Fixes some missing `accessing_from` context when querying a relationship during a resource creation.

# Contributor checklist


zachdaniel:
This works this way intentionally. It's not related yet, so you shouldn't be able to see it in the circumstance that you are attempting to relate it, else you can relate things you can't see and then be able to see it moving forward.

PR #2519, opened 2026-01-14
bugfix: Fix aggregate type resolution when aggregate references a calculation

sevenseacat:
Came across this one while debugging a Cinder issue!

When an aggregate references a calculation (e.g., `first(:name, :rel, :calc_field)`), the type resolution was failing because `Ash.Resource.Info.attribute/2` was used instead of `Ash.Resource.Info.field/2`.

Since `attribute/2` only finds attributes (not calculations), it returned nil, causing crashes like `nil.embedded?/0` when sorting through a relationship path to such an aggregate.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/domain/info/diagram.ex
+++ b/lib/ash/domain/info/diagram.ex
@@ -77,7 +77,7 @@ defmodule Ash.Domain.Info.Diagram do
     attribute =
       if aggregate.field do
         related = Ash.Resource.Info.related(resource, aggregate.relationship_path)
-        Ash.Resource.Info.attribute(related, aggregate.field)
+        Ash.Resource.Info.field(related, aggregate.field)
       end
 
     attribute_type =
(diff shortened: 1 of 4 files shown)


zachdaniel:
Good find! I'm looking at this code and there is actually an issue with how this works(not newly introduced by you, it was already there) which is that aggregates can also aggregate *other aggregates*. Ideally what we would do here is centralize this into `Ash.Query.Aggregate.aggregate_type(aggregate, type)`, and then recursively check the type if the field is an aggregate.

I'm happy to merge this as is as it clearly fixes an issue, but are you interested in also addressing that it would be really cool. If not no worries I'll make a note of it and get to it when I can 🙇

sevenseacat:
I did actually come across that issue as well, aggregates not being able to aggregate other aggregates. I'll take a look!

(need a 👍 on this one before I can merge it)

zachdaniel (approved):

zachdaniel:

PR #2523, opened 2026-01-16
fix: Allow aggregates to reference other aggregates

sevenseacat:
Continuing on from #2519, allows aggregates to reference other aggregates (with type resolution).

* Renames the `EnsureAggregateFieldIsAttributeOrCalculation` verifier to just `ValidateAggregateField` (because `EnsureAggregateFieldIsAttributeOrCalculationOrAggregate` is a bit verbose) to allow aggregates
* Updates the type resolution in `Ash.Query.Aggregate.new` to check aggregate types both first and recursively, ahead of the generic case
* Makes that aggregate type resolution public so it can be reused in other places such as `Ash.Domain.Info.Diagram`

The tests look a bit odd because there's assertions inside a `capture_io` block but as I found out, defining a module inside a `capture_io` block in order to capture any generated warnings, doesn't make the module accessible outside the block 🫠 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/resource/aggregates_test.exs
+++ b/test/resource/aggregates_test.exs
@@ -200,7 +200,10 @@ defmodule Ash.Test.Resource.AggregatesTest do
           assert query.aggregates[:total_likes].type == Ash.Type.Integer
         end)
 
-      refute stderr =~ "DslError", "Verifier should allow aggregates as fields, got: #{stderr}"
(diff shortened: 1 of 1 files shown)


zachdaniel:

PR #2532, opened 2026-01-26
fix: ensure after_transaction hooks are called

barnabasJ:
Fix: Ensure after_transaction hooks run on bulk operation batch errors

  Problem

  When a batch transaction failed during bulk_create, bulk_update, or bulk_destroy, the after_transaction hooks were
  not being invoked. This happened because errors from batch transactions bypassed the process_results function that's
   responsible for running after_transaction callbacks.

  This was inconsistent with the expected behavior where after_transaction hooks should always run after the
  transaction completes (whether successfully or with an error), giving users the opportunity to:
  - Perform cleanup operations
  - Convert errors to successes (error recovery pattern)
  - Log or track operation outcomes

  Solution

  Route batch transaction errors through process_results so that after_transaction hooks receive {:error, error}
  tuples and can respond appropriately.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -57,7 +57,6 @@ config :ash, :compatible_foreign_key_types, [
 
 if config_env() == :test do
   config :elixir, :time_zone_database, Tz.TimeZoneDatabase
-  config :ash, :disable_async?, true
 
   config :ash, Ash.Type.UUIDv7, match_v4_uuids?: true
 end


config/config.exs
```diff
@@ -57,6 +57,7 @@ config :ash, :compatible_foreign_key_types, [
 
 if config_env() == :test do
   config :elixir, :time_zone_database, Tz.TimeZoneDatabase
+  config :ash, :disable_async?, true
```
zachdaniel: This is a pretty big change to the test suite. I think we should localize this to a specific test that needs it in a setup block and make that test async?: false so that change doesn't interfere with other testsz

barnabasJ: I mostly added it because it's recommended to set it in the Ash testing docs. But thinking about it now, it makes total sense not to default to it in the library itself. 

I ran the test-suite with --repeat-until-failure, and it seems the tests are stable without it.

zachdaniel:

PR #2534, opened 2026-01-27
fix: load before notify bulk update (WIP)

### LINKED ISSUE #2525: Regression for Loading Relations in actions
### Code of Conduct

### AI Policy

### Versions

ash: 3.13.1

### Operating system

macos

### Current Behavior

Suddenly getting test failures when upgrading ash from 3.12 to 3.13.1 saying that some relationships aren't loaded.

here's the code that seems to be the culprint on the "user" side:

```elixir
AdminMember
      |> Ash.Query.set_argument(:query, Map.get(params, "query", ""))
      |> Ash.Query.sort_input(params["sort"])
      |> Ash.Query.filter(user_id == ^current_user.id)
      |> Ash.read!(
        load: [:organization, :role],
        page: page_params,
        action: :search,
        actor: current_user
      )
```

But when using this on a stream, we get an error saying: 

```
   ** (EXIT from #PID<0.15503.0>) an exception was raised:
         ** (KeyError) key :name not found in:

         #Ash.NotLoaded<:relationship, field: :organization>
```

Upon further investigation, it seems it's related to a load statment no longer working:

```elixir
update :accept do
  change set_attribute(:status, :accepted)
  change load(:organization)
end
```

I run a load after an update, bu tmost of my updates automatically load the org, so i don'' re-load the org:

```elixir
def handle_info(
        %Phoenix.Socket.Broadcast{
          topic: "admin_members:updated:" <> _id,
          payload: %{data: admin_member}
        },
        %{assigns: %{current_user: current_user}} = socket
      ) do
    # force update of current_user to handle any changes to organizations
    admin_member = Ash.load!(admin_member, [:role], actor: current_user)

    {:noreply,
     socket
     |> stream_insert(:admin_members, admin_member)
     |> assign(:current_user, Ash.load!(current_user, [:organizations], actor: current_user))}
  end
```

Will try to reproduce in a repo, later.

### Reproduction

_No response_

### Expected Behavior

_No response_

zackattackz:
Fixes #2525

This is a WIP fix for the above issue.

For bulk updates, inside `process_results`, we will only call `store_notification` for the given record/changeset combo only AFTER `load_data` is called.

I am unsure of the grander implications of this change, @zachdaniel mentioned this in discord which is worth considering:
> The complexity is that clients can request additional loads and those can clobber whatever loads you need for your notifications, so it's not necessarily stable to depend on your action's final load statement. So you should have to load on the notifier or at the destination to ensure consistency.

Specifically the "clobbering" of loads I do not understand.

Regardless I will share this PR for further review. I've included two test cases which fail currently on main, but pass with this change in `lib/ash/actions/update/bulk.ex`.

If this change makes sense and we want to move forward, it may also make sense to apply the same logic to create and destroy.

Please let me know if anything is needed :)

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/update/bulk.ex
+++ b/lib/ash/actions/update/bulk.ex
@@ -2841,37 +2841,32 @@ defmodule Ash.Actions.Update.Bulk do
          domain,
          base_changeset
        ) do
-    results =
-      Enum.flat_map(tagged_results, fn
-        {:ok, result, changeset} ->
-          if opts[:notify?] || opts[:return_notifications?] do
-            store_notification(ref, notification(changeset, result, opts), opts)
-          end
+    process_result_any_success = fn result, changeset ->
+      Process.put({:any_success?, ref}, true)
 
+      metadata = %{tagged_result_changeset: changeset}
+      result = Ash.Resource.set_metadata(result, metadata)
+
+      if index = changeset.context[context_key][:index] do
+        Ash.Resource.put_metadata(result, metadata_key, index)
+      else
+        result
+      end
+    end
+
+    results =
+      Enum.map(tagged_results, fn
+        {:ok, result, changeset} ->
           try do
             case Ash.Changeset.run_after_transactions({:ok, result}, changeset) do
               {:ok, result} ->
-                Process.put({:any_success?, ref}, true)
-
-                if opts[:return_records?] do
-                  metadata =
-                    if index = changeset.context[context_key][:index] do
-                      %{metadata_key => index}
-                    else
-                      %{}
-                    end
-
-                  [Ash.Resource.set_metadata(result, metadata)]
-                else
-                  []
-                end
+                process_result_any_success.(result, changeset)
 
               {:error, error} ->
                 error
                 |> Ash.Actions.Helpers.Bulk.maybe_rollback(resource, opts)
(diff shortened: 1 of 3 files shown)


zackattackz:
Some conflicts with https://github.com/ash-project/ash/pull/2532 so should wait for that first

zachdaniel:
👋 so at the moment it is not desirable for us to apply loads before notifications. It creates a dependency that can't be relied on. For example, if your notifier needs `comments.author`, and so you apply a load like `comments: :author`, but the *caller* adds a load like `comments: Ash.Query.limit(Comment, 1)` to load only one comment per post, then your notifier now will only see one comment per post.

To solve for this, we should configure load dependencies on each `notify` call explaining what must be loaded, and then find a way to leverage the calculation system that we have to fetch those loads. It is a really complex topic in reality as we'd want to make sure they were deduplicated, so it would be something like:

- merge the attributes from all relevant notifiers with the selected attributes
- for all calculations, load calculations with something special in the name identifying them, i.e `calc1: [as: {:notifier, 1}, ...]`
- for all relationships, do the same as above but using the `LoadRelationship` calculation

zackattackz:
Makes sense, thank you for the thorough explanation

PR #2537, opened 2026-01-28
fix: pass full context to `Ash.load!` in cascade changes

### LINKED ISSUE #2536: cascade_destroy and cascade_update missing actor in Ash.load! call
## Description

The `Ash.load!` calls in `cascade_destroy` and `cascade_update` are missing the actor from the context, causing policy failures when policies require an actor.

## Location

- `lib/ash/resource/change/cascade_destroy.ex:347-350`
- `lib/ash/resource/change/cascade_update.ex:261-264`

Both have:

```elixir
|> Ash.load!(
  [{relationship.name, load_query}],
  tenant: tenant
)
```

## Root Cause

Looking at the git history:

1. **May 2024** (original): Used `authorize?: false` to skip authorization
2. **Feb 2025** (`f19a562`): Added `tenant: tenant` alongside `authorize?: false`
3. **Apr 2025** (`00818c1`): Removed `authorize?: false` but kept only `tenant: tenant`

When `authorize?: false` was removed, the actor should have been added. Without either, the load operation attempts to authorize but fails because there's no actor.

## Fix

Change `tenant: tenant` to `Ash.Context.to_opts(context)` to pass all relevant context options (actor, tenant, tracer, context, authorize?). This is consistent with how `context_opts` is already passed to `Ash.bulk_destroy!`/`Ash.bulk_update!` later in the same code path.

## Reported By

Ash Premium Support customer

jimsynz:
## Summary

- Fix `cascade_destroy` and `cascade_update` to pass full context (including actor) to `Ash.load!`
- Previously only `tenant: tenant` was passed, causing policy failures when policies require an actor

## Details

The `Ash.load!` calls in both cascade changes were missing the actor from the context. This was a regression introduced in #1948 when `authorize?: false` was removed but the actor wasn't added.

The fix uses `Ash.Context.to_opts(context)` to pass all relevant context options (actor, tenant, tracer, context, authorize?), consistent with how `context_opts` is already passed to `Ash.bulk_destroy!`/`Ash.bulk_update!` in the same code path.

Closes #2536

### DIFF
--- a/lib/ash/resource/change/cascade_destroy.ex
+++ b/lib/ash/resource/change/cascade_destroy.ex
@@ -346,8 +346,7 @@ defmodule Ash.Resource.Change.CascadeDestroy do
           |> List.wrap()
           |> Ash.load!(
             [{relationship.name, load_query}],
-            scope: context,
-            authorize?: false
+            scope: context
           )
           |> Enum.flat_map(fn record ->
             record
--- a/lib/ash/resource/change/cascade_update.ex
+++ b/lib/ash/resource/change/cascade_update.ex
@@ -260,8 +260,7 @@ defmodule Ash.Resource.Change.CascadeUpdate do
         |> List.wrap()
         |> Ash.load!(
           [{relationship.name, load_query}],
-          scope: context,
-          authorize?: false
+          scope: context
         )
         |> Enum.flat_map(fn record ->
           record


lib/ash/resource/change/cascade_destroy.ex
```diff
@@ -346,7 +346,7 @@ defmodule Ash.Resource.Change.CascadeDestroy do
           |> List.wrap()
           |> Ash.load!(
             [{relationship.name, load_query}],
-            tenant: tenant
+            Ash.Context.to_opts(context)
```
zachdaniel: You can just do:
`scope: context`, but we should also do `authorize?: false`, so `scope: context, authorize?: false` to retain original behavior.

lib/ash/resource/change/cascade_update.ex
```diff
@@ -260,7 +260,7 @@ defmodule Ash.Resource.Change.CascadeUpdate do
         |> List.wrap()
         |> Ash.load!(
           [{relationship.name, load_query}],
-          tenant: tenant
+          Ash.Context.to_opts(context)
```
zachdaniel: Same here.

barnabasJ (dismissed):

lib/ash/resource/change/cascade_destroy.ex
```diff
@@ -346,7 +346,8 @@ defmodule Ash.Resource.Change.CascadeDestroy do
           |> List.wrap()
           |> Ash.load!(
             [{relationship.name, load_query}],
-            tenant: tenant
+            scope: context,
```
stevebrambilla: Heads up that this function rebinds `context` above:

https://github.com/ash-project/ash/blob/32ee606081461b1afb2b444733986205b9e6cb81/lib/ash/resource/change/cascade_destroy.ex#L299-L303

So it's missing the `tenant` here now

lib/ash/resource/change/cascade_destroy.ex
barnabasJ: wasn't `authorized?: false` removed intentionally in favor of having `accessing_from` set and having policies? 

you didn't explicitly answer the question, but you merged the [PR](https://github.com/ash-project/ash/pull/1948#discussion_r2027491982)

lib/ash/resource/change/cascade_destroy.ex
zachdaniel: 🤯 good call!

zachdaniel: We should not overwrite `context` in the first place to avoid this confusion.

lib/ash/resource/change/cascade_destroy.ex
zachdaniel: Ah, yeah you're right 👍 maybe a comment here saying that this is meant to run authorization.

zachdaniel:
@jimsynz my bad, @barnabasJ is right, we shouldn't be doing `authorize?: false` here. Will merge once that's changed 🙇

PR #2544, opened 2026-02-02
fix: manage_relationship in atomics

barnabasJ:
Summary

  Fixes a regression introduced in commit 02b260a39 ("make ignored managed relationships not break atomicity") where
  manage_relationship changes were silently skipped during atomic bulk updates.

  Problem

  The atomic/3 callback was added unconditionally to ManageRelationship, making all manage_relationship changes claim
  to be atomic. However, the atomic bulk update path doesn't call manage_relationships() - only the stream path does.
  This caused relationship changes to be silently ignored.

  Solution

  Return {:not_atomic, reason} for non-ignored relationships, forcing the stream-based fallback path which properly
  executes relationship management. The ignore?: true optimization from the original commit is preserved.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/change/manage_relationship.ex
+++ b/lib/ash/resource/change/manage_relationship.ex
@@ -36,7 +36,12 @@ defmodule Ash.Resource.Change.ManageRelationship do
   end
 
   def atomic(changeset, opts, context) do
-    {:ok, change(changeset, opts, context)}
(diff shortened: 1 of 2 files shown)


zachdaniel:

Issue #1537, opened 2024-10-20
misleading error message with multiple "use Ash.Domain" macro invocations.


at7heb:
**Describe the bug**
I had two `use Ash.Domain` macro calls in my domain definition file. The error message 
```
    error: def can?/3 defines defaults multiple times. Elixir allows defaults to be declared once per definition. Instead of:
```
misled me, and the indicated line was the `defmodule` line:
```
    │
  1 │ defmodule Ogl.Example do
    │ ^^^^^^^^^^^^^^^^^^^^^^^^
    │
    └─ lib/ogl/example.ex:1
```

**To Reproduce**
```defmodule
Ogl.Example do
  use Ash.Domain

  # In your Domain(s)
  use Ash.Domain,
    extensions: [AshAdmin.Domain]
end

```
**Expected behavior**
The error message should have indicated that the second `use Ash.Domain` line was the problem.
**Runtime**
```
 Erlang/OTP 27 [erts-15.1] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit] [dtrace]

IEx 1.17.3 (compiled with Erlang/OTP 27)
```
 - macOS Sonoma 14.6.1
 - Ash   `"ash": {:hex, :ash, "3.4.11", "f96d33a2f63773262dec7dae6b7af3fa998b8fd1f1ac22dbed8e49475c92fbb3", ...`

**Additional context**
Low priority -- I won't do this again.

mudriyjo:
I can take it, if it's still relevant

zachdaniel:
@mudriyjo please do 🙇

yenrab:
Marked as closed in the comments but not actually closed.

PR #2545, opened 2026-02-02
Fix/1537 improve domain error message

TravelCurry02:
# Contributor checklist

- [X ] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [X ] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring

Fix misleading error message for duplicate use Ash.Domain (#1537)

## Problem
When `use Ash.Domain` was called multiple times in a module, the error message was misleading - it pointed to the `defmodule` line (line 1) instead of the duplicate `use Ash.Domain` invocation.

The original error message was:
```
    error: def can?/3 defines defaults multiple times. Elixir allows defaults to be declared once per definition. 
```
And it pointed to the `defmodule` line, making it very difficult to identify the actual problem.

## Solution
- Added detection for duplicate `use Ash.Domain` calls in `handle_opts/1`
- Error now correctly points to the second `use Ash.Domain` line using `__ENV__.line`
- Error message clearly indicates which line had the first call
- Applied same fix to `Ash.Domain.Interface` for consistency
- Added test case verifying error points to correct line (line 3, not line 1)

## Changes
- `lib/ash/domain/domain.ex`: Added duplicate detection with proper error location
- `lib/ash/domain/interface.ex`: Applied same fix for consistency
- `test/domain/domain_test.exs`: Updated test to verify error points to correct line

## Testing
- Added test in `test/domain/domain_test.exs` that verifies:
  - Error points to line 3 (second invocation), not line 1
  - Error message contains expected helpful text ("can only be called once", "Remove the duplicate")
- All existing tests pass

### DIFF
--- a/lib/ash/domain/domain.ex
+++ b/lib/ash/domain/domain.ex
@@ -51,7 +51,6 @@ defmodule Ash.Domain do
   @impl Spark.Dsl
   def handle_opts(opts) do
     quote do
-      # Check if `use Ash.Domain` has already been called
       if Module.get_attribute(__MODULE__, :__ash_domain_used__) do
         first_line = Module.get_attribute(__MODULE__, :__ash_domain_used__)
 


lib/ash/domain/interface.ex
```diff
@@ -8,18 +8,29 @@ defmodule Ash.Domain.Interface do
   defmacro __using__(_) do
     env = __CALLER__
 
-    ash_domain_already_using =
```
zachdaniel, line 11: We should be able to just remove it from here and only define this logic in `domain.ex`

lib/ash/domain/domain.ex
```diff
@@ -51,6 +51,24 @@ defmodule Ash.Domain do
   @impl Spark.Dsl
   def handle_opts(opts) do
     quote do
+      # Check if `use Ash.Domain` has already been called
```
zachdaniel: We can remove the unnecessary comments.

PR_DESCRIPTION.md
```diff
@@ -0,0 +1,37 @@
+# PR Description Template
```
zachdaniel: Looks like some generated files left in?

TravelCurry02: I thought I deleted those two files. They were just for documentation for the PR.

zachdaniel:
Still a few comments to address and a generated file left in the PR.

lib/ash/domain/domain.ex
TravelCurry02: I removed the comment at line 54

lib/ash/domain/interface.ex
TravelCurry02: I removed the duplicate logic and ran the tests again, and they all passed.

zachdaniel (approved):

zachdaniel:

PR #2546, opened 2026-02-03
feat: manual relationship via list

### LINKED ISSUE #2505: Manual relationships should support returning a list of relationship values
### Code of Conduct

### AI Policy

As discussed on [Discord](https://discord.com/channels/711271361523351632/1456872934638354432), manual relationships return a map of data, keyed by the primary key for the record that the value belongs to.

In my example, my primary key is a CiString, which does not implement simple equality so we end up in [this code branch](https://github.com/ash-project/ash/blob/main/lib/ash/actions/read/relationships.ex#L1039-L1066), using Enum.find_value for every record in the result set - slow for large data sets!

(In my case it takes ~12 seconds to do this mapping for a result set of 5500, half of which is spent in `String.Unicode.downcase`)

I thought that manual relationships would work like calculations - a list of records in, and a list of results out, each index of the input set mapping to the same record in the output set.

That would be the best solution I think - so you'd have something like this:

```elixir
def load([record1, record2, record3], _, _) do 
  # Stuff to load the data to relate

  {:ok, [results1, results2, results3]}
end
```

Refactoring my cobease to not use CiStrings as primary keys but they're so darn useful 😅 

### Additional context

_No response_

zackattackz:
Introduces the ability to return a list from the `Ash.Resource.ManualRelationship.load/3` callback.

It works similarly to the `Ash.Resource.Calculation.calculate/3` callback: Each result in the returned list is related to the record from the same position in the input records list.

The ability to return a map is maintained for backwards compatibility (Though it perhaps should be deprecated at some point?).

The main reason for this is to improve performance for cases where the loading resource uses a non-simple-equality primary key (See below issue).

Closes #2505

### Tests
The test cases included here are probably way too much, but they are very thorough in their coverage of the affected code to ensure full backwards compatibility. If you want to check the tests that pass on main, just remove `:list` from `@strategies_to_test` and run against main branch.

At least they prove correctness, but if they are too much I understand, and I can put together some simpler tests.

### Documentation
Also, this change should be documented - but I wasn't sure how exactly it should be communicated. Should we replace any mention of returning a map with instead returning a list? Should returning a map be mentioned as a legacy option with performance drawbacks? Either way it should definitely touch the "Manual Relationships" section of `documentation/topics/resources/relationships.md`. Just let me know any thoughts and I can draft the documentation.

### Implementation
The implementation works by first normalizing the result from the user's `load/3` callback:
- If it is a map, it is converted to a list in the "new format" which will end up having the same performance drawbacks as before this commit (O(n^2) for non-simple pkeys). Users can switch to returning a list instead for performance gains.
- If it is a list, it is untouched.

Then the related records are flattened, and loaded with the given query. Then they are regrouped back into a list efficiently by their index (:manual_key).

Finally, this list is passed into `do_attach_related_records`, which simply zips them together with the loading records.

Since the O(n^2) part is only happening in cases where a map is returned, users that instead use a list will not hit this bottleneck.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes (TODO)
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -505,12 +505,14 @@ defmodule Ash.Actions.Read.Relationships do
             tenant: related_query.tenant
           })
           |> case do
-            {:ok, records} ->
-              records
-              |> Enum.flat_map(fn {key, value} ->
+            {:ok, result_records} ->
+              result_records
+              |> normalize_manual_results(records, relationship)
+              |> Enum.with_index()
+              |> Enum.flat_map(fn {value, index} ->
                 value
                 |> List.wrap()
-                |> Enum.map(&Ash.Resource.put_metadata(&1, :manual_key, key))
+                |> Enum.map(&Ash.Resource.put_metadata(&1, :manual_key, index))
               end)
               |> Ash.load(related_query,
                 domain: related_query.domain,
@@ -520,7 +522,7 @@ defmodule Ash.Actions.Read.Relationships do
               )
               |> case do
                 {:ok, results} ->
-                  {:ok, regroup_manual_results(results, relationship)}
+                  {:ok, regroup_manual_results(results, relationship, Enum.count(records))}
 
                 {:error, error} ->
                   {:error, error}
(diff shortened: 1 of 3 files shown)


zachdaniel:
🔥 this is awesome.

zachdaniel:
> Also, this change should be documented - but I wasn't sure how exactly it should be communicated. Should we replace any mention of returning a map with instead returning a list? Should returning a map be mentioned as a legacy option with performance drawbacks? Either way it should definitely touch the "Manual Relationships" section of documentation/topics/resources/relationships.md. Just let me know any thoughts and I can draft the documentation.

Yep, I'd suggest updating any relevant docs to only show this newer strategy, with a footnote alluding to the deprecated methodology.

zachdaniel:
Looks good to me! I'll merge and docs can be done in a follow up 🥳

zachdaniel:

zackattackz:
Thank you! I'll follow up with the docs PR sometime soon.

PR #2547, opened 2026-02-04
bypass required attribute check for atomic_set & remove required attribute errors

rtorresware:
My clanker's justification of this change vs a bigger refactor:

> I opted to clear the specific Ash.Error.Changes.Required for the attribute when atomic_set records a create atomic/value, because for_create/3 validates required attrs before atomic_set runs. The test pattern (for_create → atomic_set) is supported today, so without cleanup the changeset stays invalid even though the atomic expression will populate the value at INSERT. This keeps behavior consistent across data layers without re-running validations or invalidating the whole changeset.
> Ideal fix: reorder validation so required checks see create_atomics (or update require_values/4 for :create to treat create_atomics as “set”), eliminating the need for post‑hoc error removal.

Opencode session: https://opncd.ai/share/pWJqbSRE

Regression tests live in ash_postgres PR, as this bug can only be replicated in sql backed data layers: https://github.com/ash-project/ash_postgres/pull/684

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1225,7 +1225,8 @@ defmodule Ash.Changeset do
          {:atomic, condition} <-
            atomic_condition(where, changeset, context),
          {{:atomic, modified_changeset?, new_changeset, atomic_changes, validations,
-           create_atomics}, condition} <-
+           create_atomics},
+          condition} <-
            {atomic_with_changeset(
               module.atomic(changeset, change_opts, struct(Ash.Resource.Change.Context, context)),
               changeset
@@ -2669,10 +2670,12 @@ defmodule Ash.Changeset do
                 set_error_field(value, attribute.name)
               end
 
-            %{
-              changeset
-              | create_atomics: Keyword.put(changeset.create_atomics, attribute.name, value)
-            }
+            changeset
+            |> Map.put(
+              :create_atomics,
+              Keyword.put(changeset.create_atomics, attribute.name, value)
+            )
+            |> clear_required_attribute_error(attribute.name)
 
           {:ok, value} ->
             allow_nil? =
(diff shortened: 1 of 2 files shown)


zachdaniel:
We have a better way that we do this for regular atomics, we just need to adopt a similar strategy for create_atomics 😄 We set the expression itself to `if is_nil(expr), do: expr, else: error(...)`

```elixir
  @doc false
  def handle_allow_nil_atomics(changeset, actor) do
    changeset.atomics
    |> Enum.reduce(changeset, fn {key, value}, changeset ->
      attribute = Ash.Resource.Info.attribute(changeset.resource, key)

      if attribute.primary_key? do
        changeset
      else
        allow_nil? =
          attribute.allow_nil? and attribute.name not in changeset.action.require_attributes

        if allow_nil? || not Ash.Expr.can_return_nil?(value) do
          value
        else
          if Ash.DataLayer.data_layer_can?(changeset.resource, :expr_error) do
            expr(
              if is_nil(type(^value, ^attribute.type, ^attribute.constraints)) do
                error(
                  ^Ash.Error.Changes.Required,
                  %{
                    field: ^attribute.name,
                    type: ^:attribute,
                    resource: ^changeset.resource
                  }
                )
              else
                ^value
              end
            )
          else
            {:not_atomic,
             "Failed to validate expression #{inspect(value)}: data layer `#{Ash.DataLayer.data_layer(changeset.resource)}` does not support the expr_error"}
          end
        end
        |> case do
          {:not_atomic, error} ->
            Ash.Changeset.add_error(changeset, error)

          value ->
            %{changeset | atomics: Keyword.put(changeset.atomics, key, value)}
        end
      end
    end)
    |> Ash.Changeset.hydrate_atomic_refs(actor, eager?: true)
    |> then(fn
      {:not_atomic, value} ->
        {:not_atomic, value}

      %Ash.Changeset{} = changeset ->
        if changeset.action.type == :update do
          attributes =
            changeset.attributes
            |> Map.keys()
            |> Enum.reject(&Ash.Resource.Info.attribute(changeset.resource, &1).allow_nil?)

          require_values(changeset, :update, false, attributes)
        else
          changeset
        end
    end)
  end
```

The correct answer here would be to adapt this.

PR #2551, opened 2026-02-05
demonstrate bug with mix of batch_change and change callbacks

jsw800:
I noticed this bug. These changes don't run in order when one has a `batch_change` and one has a `change` callback.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/change_test.exs
+++ b/test/change_test.exs
@@ -1,0 +1,106 @@
+defmodule Ash.Test.ChangeTest do
+  @moduledoc false
+  use ExUnit.Case, async: true
+
+  defmodule UpdateFirstPublishedAt do
+    @moduledoc false
+    use Ash.Resource.Change
+
+    @impl true
+    def batch_change(changesets, _opts, _context) do
+      Enum.map(changesets, fn changeset ->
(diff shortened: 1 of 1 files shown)


test/change_test.exs
```diff
@@ -0,0 +1,106 @@
+defmodule Ash.Test.ChangeTest do
+  @moduledoc false
+  use ExUnit.Case, async: true
+
+  defmodule UpdateFirstPublishedAt do
+    @moduledoc false
+    use Ash.Resource.Change
+
```
jsw800, line 8: adding these lines here:

```elixir
  @impl true
  def change(changeset, opts, context) do
    [changeset]
    |> batch_change(opts, context)
    |> List.first()
  end
```

causes both of these tests to pass, but they should pass as is.

zachdaniel:
Fixed here: 0cba7f34e56ed6dcb72537b299f8da21314d306f

jsw800:
Awesome, thanks!

PR #2557, opened 2026-02-09
test: Add failing test for calculations with expressions in filters

Munksgaard:
I believe filtering on a calculation should work, even though the calculation takes arguments. It works when using the macro-version of `Ash.Query.filter`, but not with the keyword version, and also not in the keyword version on code interface functions.

If the test does not represent desired behavior, please feel free to dismiss the PR.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/code_interface_test.exs
+++ b/test/code_interface_test.exs
@@ -273,6 +273,12 @@ defmodule Ash.Test.CodeInterfaceTest do
                   end)
                 end,
                 load: [:first_name, :last_name]
+
+      calculate :full_name_is, :boolean do
+        argument :required_name, :string, allow_nil?: false
+
+        calculation expr(full_name == ^arg(:required_name))
+      end
     end
 
     attributes do
@@ -412,6 +418,47 @@ defmodule Ash.Test.CodeInterfaceTest do
       assert {:ok, true} == User.can_get_by_id(nil, "some uuid")
       assert User.can_read_users?(nil)
       assert User.can_get_by_id?(nil, "some uuid")
+    end
+
+    test "can filter on calculations" do
+      require Ash.Query
+      user = User.create!("bob", %{last_name: "rob"}, load: :full_name)
+
+      # Can load the calculation
+      assert %User{full_name_is: true} =
+               Ash.load!(user, full_name_is: %{required_name: "bob rob"})
+
+      # Can filter on calculation using Ash.Query.filter macro
+      assert [%User{first_name: "bob", last_name: "rob"}] =
+               User
+               |> Ash.Query.filter(full_name_is(required_name: "bob rob"))
+               |> Ash.read!()
+
+      # Negative test: returns empty list when required name doesn't match
+      assert [] =
+               User
+               |> Ash.Query.filter(full_name_is(required_name: "rob bob"))
+               |> Ash.read!()
+
+      # Can filter using Ash.Query filter keyword
+      assert [%User{first_name: "bob", last_name: "rob"}] =
+               User
+               |> Ash.Query.filter(full_name_is: [required_name: "bob rob"])
+               |> Ash.read!()
+
+      # Negative test
+      assert [] =
+               User
+               |> Ash.Query.filter(full_name_is: [required_name: "rob bob"])
+               |> Ash.read!()
+
+      # Can filter on code interface
+      assert [%User{first_name: "bob", last_name: "rob"}] =
+               User.read_users!(query: [filter: [full_name_is: [required_name: "bob rob"]]])
+
+      # Negative test
+      assert [] =
+               User.read_users!(query: [filter: [full_name_is: [required_name: "rob bob"]]])
     end
   end
 


test/code_interface_test.exs
```diff
@@ -413,6 +419,27 @@ defmodule Ash.Test.CodeInterfaceTest do
...
+               |> Ash.Query.filter(full_name_is(required_name: "bob rob"))
+               |> Ash.read!()
+
+      assert [%User{first_name: "bob", last_name: "rob"}] =
+               User
+               |> Ash.Query.filter(full_name_is: %{required_name: "bob rob"})
+               |> Ash.read!()
+
+      assert [%User{first_name: "bob", last_name: "rob"}] =
+               User.read_users!(query: [filter: [full_name_is: %{required_name: "bob rob"}]])
```
zachdaniel: IIRC it is expecting the input to be a keyword list (which is admittedly a bit strange but our options are a bit limited here). Can you try that and confirm?

Munksgaard: I get the same error with a keyword list, but I'm happy to update the PR to contain either or both. What do you want?

Munksgaard: I've updated the PR with some negative tests and it is now using keyword lists instead of maps.

zachdaniel:
Awesome, thank you. I'll investigate later this week/this weekend 🙇 I think that we may just have a different format that we support for this, but I'll let you know.

zackattackz:
Doing it like this by passing an map with an "input" key seems to work

```elixir
      # Can filter using Ash.Query filter keyword
      assert [%User{first_name: "bob", last_name: "rob"}] =
               User
               |> Ash.Query.filter(full_name_is: %{input: [required_name: "bob rob"]})
               |> Ash.read!()

      # Negative test
      assert [] =
               User
               |> Ash.Query.filter(full_name_is: %{input: [required_name: "rob bob"]})
               |> Ash.read!()

      # Can filter on code interface
      assert [%User{first_name: "bob", last_name: "rob"}] =
               User.read_users!(
                 query: [filter: [full_name_is: %{input: [required_name: "bob rob"]}]]
               )

      # Negative test
      assert [] =
               User.read_users!(
                 query: [filter: [full_name_is: %{input: [required_name: "rob bob"]}]]
               )
```

Because there is this part in the filter code https://github.com/ash-project/ash/blob/d0fdf04510b3e091493be2c4f1664fd2197a9103/lib/ash/filter/filter.ex#L3314-L3325

Which extracts out the desired "input"


I think though there should be an extra case here like this to allow passing a keyword list like
```elixir
full_name_is: [input: [required_name: "bob rob"]]
```

Said case would look like this:

```elixir
{input, nested_statement} =
    case nested_statement do
      %{"input" => input} ->
        {input, Map.delete(nested_statement, "input")}

      %{input: input} ->
        {input, Map.delete(nested_statement, :input)}

     # > HERE <
      [input: input] ->
        {input, Map.new(Keyword.delete(nested_statement, :input))}

      _ ->
        {%{}, nested_statement}
    end
```

Thoughts?

zackattackz:
Actually the case for keyword list should probably be:

```elixir
  kw when is_list(kw) ->
    {input, rest} = Keyword.pop(kw, :input, %{})
    {input, Map.new(rest)}
```

To match more than just a list with `:input`

zachdaniel:
Down to add that, but we need to add a `Keyword.keyword?` check there before assuming a list is a keyword. I'd accept that PR 👍

Munksgaard:
Good find @zackattackz! Is the current behavior documented anywhere? I couldn't find it when reading the documentation. Whatever we do with the nested case, I think we should also improve the documentation a bit.

zackattackz:
Thank you both :) I could not find any documentation searching for :input, "input", and :input in .md files. So I'd think it could be added around here https://hexdocs.pm/ash/Ash.Filter.html#module-keyword-list-syntax ?

Right below that the docs also say
> Maps are also accepted, as are maps with string keys. Technically, a list of [{"string_key", value}] would also work.

But I think doing `[{"input", [...]}]` wouldn't work in this specific case either since it wouldn't match the case of `%{"input" => input}`

So we may also want a new case for that too? (not sure though would need @zachdaniel to weigh in)

Either way, if you want to add those to your PR for submission please feel to! @Munksgaard 

Otherwise if you can't I'd be glad to open a new one, it's up to you.

Munksgaard:
Sorry, I never got around to follow up on this. @zachdaniel: Did you merge something to "fix" this issue, or should I update this PR?

zachdaniel:
I force pushed something to main and it looks like it closed all PRs and broke the history somehow 😢

zachdaniel:
@ken-kost can you please reopen this? Sorry for the inconvenience 🙇

zachdaniel:
Sorry, I meant @Munksgaard. lol, I'm all over the place. @ken-kost you can reopen *your* PR 😓

Munksgaard:
No worries :-) It doesn't look like I can reopen the PR though. Can you do it or should I open a new PR?

Munksgaard:
Ping: Can you reopen this PR, or should I create a new one?

zachdaniel:
Sorry, I can't reopen this, it won't let me. The PR would need to be created fresh.

zachdaniel:
(and thank you for the ping 🙇)

PR #2558, opened 2026-02-09
Fix/add error messages

### LINKED ISSUE #365: Support error messages in `cast_stored` and `dump_to_native`.
A problem that we encounter when using `cast_stored` and `dump_to_native` on embeds is that it can be very difficult to determine where an error occurred, because there are no error messages for those callbacks. So you could have a 6-nested embedded resource and something deep in there could fail to cast to/from the database and all you would get is `:error`.

Support error messages in `cast_stored` and `dump_to_native`, i.e `{:error, "Error message"}`. This would be used in embeds to return which key specifically failed the cast.

mikaelweiss:
# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [x] Update dependencies

Resolves #365

### DIFF
--- a/lib/ash/embeddable_type.ex
+++ b/lib/ash/embeddable_type.ex
@@ -436,23 +436,8 @@ defmodule Ash.EmbeddableType do
             {:fetch, :error} ->
               {:cont, {:ok, struct}}
 
-            :error ->
-              {:halt, {:error, field: attr.name}}
-
-            {:error, keyword} ->
-              errors =
-                keyword
-                |> List.wrap()
-                |> Ash.Helpers.flatten_preserving_keywords()
-                |> Enum.map(fn
-                  string when is_binary(string) ->
-                    [message: string, field: attr.name]
-
-                  vars ->
-                    Keyword.put(vars, :field, attr.name)
-                end)
-
-              {:halt, {:error, errors}}
+            other ->
+              {:halt, Ash.Helpers.error_with_context(other, field: attr.name)}
           end
         end)
         |> case do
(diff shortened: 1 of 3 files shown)


zachdaniel:
Very nice! I think we'll need to do a few more things on this one.

Firstly, we'll need to make sure that the callbacks defined in the `EctoType` module for a given type still follow Ecto's expectations:

https://github.com/elixir-ecto/ecto/blob/v3.13.5/lib/ecto/type.ex#L294

And to really vet that, we'll want to make sure to test `ash_postgres` against these changes. A good way to do that will be to clone down `ash_postgres`, set `export ASH_VERSION=local` and then run the tests (typically `mix test --exclude migration`).

mikaelweiss:
The load/3 and dump/3 functions in the EctoType module (lib/ash/type/type.ex:1731 and lib/ash/type/type.ex:1744) both swallow errors with catch-all clauses, which follows Ecto's expectations:

```elixir
@impl true
def load(term, _, params) do
  parent = @parent

  case parent.cast_stored(term, params) do
    {:ok, value} ->
      {:ok, value}

    _ ->
      :error
  end
end

@impl true
def dump(term, _dumper, params) do
  parent = @parent

  case parent.dump_to_native(term, params) do
    {:ok, value} ->
      {:ok, value}

    _ ->
      :error
  end
end
```
I also ran `ASH_VERSION=local mix test` in `ash_postgres` (full test suite to be sure) and they passed.

Is there anything else you'd like me to test or check?

zachdaniel:
Ah, yeah so there is one more major thing to test. What we want to do is set up a resource that works against ETS, manipulate the ETS table to have a value that can't be loaded, and ensure that the error bubbles up in a way that is *not* exposed to the end user. It will likely need a custom error built to function that way.

Both `dump_to_native` and `cast_stored` should produce errors different from `InvalidAttribute` or `InvalidArgument` for example.

zachdaniel:
I accidentally force pushed a divergent history to the main branch and it closed all PRs 😢 @mikaelweiss can you please reopen this? Sorry for the inconvenience 🙇

PR #2559, opened 2026-02-09
fault tolerant install script #2420 fix

celeste-wahlquist:
# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [x] Update dependencies


celeste-wahlquist:
There are warnings relating to how the dialyzer is reading some arguments in relation to their opaqueness. There’s no dependency or shared code between the install task and the changeset.ex/action_input.ex module, and I have not modified changeset.ex or action_input.ex. It appears there were some pre-existing warnings in this file.

zachdaniel:
Unfortunately we really can't put postgres specific steps in here. These would need to go in the `ash_postgres` installer for example. `mix ash.setup` ends up calling the `setup` function in `AshPostgres.DataLayer` which is a place we could potentially modify these things.

With that said, *at install time* is not really a place where we want to enforce that these things are running. You may for example be about to run the project in a container. I think the only really doable thing here is in `mix ash.setup` to provide better output of what tasks are being run and what will be run, and when something fails, to print a message explaining that some set of things were not done and so you should fix them and rerun `mix ash.setup`.

PR #2562, opened 2026-02-10
fix: manual relationship implies no_attributes? true

zackattackz:
Adds a transformer on relationships.{has_one, has_many} DSLs to automatically set `no_attributes?` to `true` if `manual` is set.

Reworks logic in places that use `no_attributes?` and `manual` together to take this new implication into consideration. (read.ex, relationships.ex, and schema.ex)

Adds some tests to ensure that this implication holds. Also adds regression tests to ensure that `select` callback for manual relationships is respected.

### Notes

1. Should we allow overriding `no_attributes?: false` in manual relationships? (That is, change the transformer logic to NOT set `no_attributes?: true` if they've manually specified it as `false` in the DSL)
2. Do you think any other ash project repos may need similar logic changes like the ones made here? (e.g like in read.ex)

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


lib/ash/resource/relationships/shared_transformers.ex
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: 2019 ash contributors <https://github.com/ash-project/ash/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Ash.Resource.Relationships.SharedTransformers do
```
zachdaniel: There is a different shared module that we use for shared helpers across relationships, lets put it there instead of adding a new module: Ash.Resource.Relationships.SharedOptions

zackattackz:
ash_postgres may need some changes too, looking into it

zackattackz:
Yeah it seems like ash_postgres has some failures in certain tests with ash set to this branch. Will take a look tmmrw!

lib/ash/resource/relationships/shared_transformers.ex
zackattackz: Addressed in latest commit.

Also opened https://github.com/ash-project/ash_sql/pull/213 to fix the issue with this PR + ash_postgres.

zackattackz:
Basically, like the issue in ash_sql, any ash extensions / data layers or whatever that have code that tests something like:

```
if relationship.no_attributes?:
    ...something
else if relationship.manual:
    ...do something else
```
 
 Has the potential to break since now after this PR the no_attributes? branch will execute instead of the manual one

zackattackz:
^ So they can pretty much be fixed by just doing the manual case first. (like I did in the ash_sql PR)

```
if relationship.manual:
    ...do something else
else if relationship.no_attributes?:
    ...something
```

zachdaniel:

PR #2570, opened 2026-02-18
Add error messages

### LINKED ISSUE #365: Support error messages in `cast_stored` and `dump_to_native`.

mikaelweiss:
# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [x] Update dependencies

Resolves #365
Related: #2558

### DIFF
--- a/lib/ash/error/invalid/invalid_stored_data.ex
+++ b/lib/ash/error/invalid/invalid_stored_data.ex
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2019 ash contributors <https://github.com/ash-project/ash/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
 defmodule Ash.Error.Invalid.InvalidStoredData do
   @moduledoc "Used when stored data cannot be cast to its expected type when loading from the data layer."
 


mikaelweiss:
Current state, last thing to test:

https://github.com/ash-project/ash/pull/2558#issuecomment-3888208727

mikaelweiss:
@zachdaniel
I added tests and a custom error type as well

zachdaniel:
Nice, this looks great. They won't actually produce error messages in many cases due to the ecto type not actually supporting those error messages unfortunately 😢. But it can help with embedded resources, custom types, unions, etc. But to fix that is something we'd have to take up with Ecto itself ultimately 😄

mikaelweiss:
Gotcha. Do you want me to make any changes then, or is this all good?

zachdaniel:

(no changes necessary, just a limitation of the current implementation, but I'll advocate for enhancing ecto to support this)

Issue #2569, opened 2026-02-18
Inconsistent option `error?` vs `not_found_error?` for `Ash.get()` vs `get? true` in action?


StephanH90:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Ash: 3.15

### Operating system

ubuntu

### Current Behavior

I'm not sure it's a bug or by design but I noticed that the options for "don't return an error if no record is found" is different between `Ash.get` vs a read action with `get? true`:

```ex
# Using Ash.get():
Ash.get(SomeResource, error?: false) # <-- here its `error?`

# Using a read action in a resource:
defmodule MyResource do
  actions do
    read :some_action do
      get? true
    end
  end
end

MyResource.some_action(not_found_error?: false) # <-- here its `not_found_error?`
```

They seem to be doing the same thing, so shouldn't they use the same option?

zachdaniel:
Hmm....yeah I think that's a good point. Best bet would be to *add* a `not_found_error?` option to `Ash.get` for now, and add a note in the docs that `error?: false` will be removed in 4.0. Its only known at runtime when its used, so I'd rather not make it warn considering this would generate tons of logs for users.

RetroDev256:
<img width="1058" height="122" alt="Image" src="https://github.com/user-attachments/assets/37841bad-09d1-4dce-9cdb-4a9119f425d7" />

This issue no longer requires a PR, though it should probably remain open until 4.0 so we are reminded to remove the deprecated code.

zachdaniel:
Good call!

zachdaniel:
Good call!

PR #2572, opened 2026-02-18
feat: add `not_found_error?` to `Ash.get` to bring it in-line with actions

StephanH90:
Addresses this issue: https://github.com/ash-project/ash/issues/2569

While I was implementing it I also realised that the tests in `ash_test` don't actually work with a real data layer, so I slightly changed them where necessary.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -230,7 +230,14 @@ defmodule Ash do
                     )
 
   @get_opts_schema [
+                     # This will be removed in 4.0. Use `not_found_error?` instead.
                      error?: [
+                       type: :boolean,
+                       default: true,
+                       doc:
+                         "Deprecated: Use not_found_error? instead. Whether or not an error should be returned or raised when the record is not found. If set to false, `nil` will be returned. Deprecated: This will be removed in 4.0."
+                     ],
+                     not_found_error?: [
                        type: :boolean,
                        default: true,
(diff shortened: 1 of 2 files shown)


lib/ash.ex
```diff
@@ -2146,7 +2154,8 @@ defmodule Ash do
         {:ok, single_result}
 
       {:ok, []} ->
-        if opts[:error?] do
+        # deprecated: opts[:error?] will be removed in 4.0 and then opts[:not_found_error?] can be used directly here
+        if Enum.all?([opts[:error?], opts[:not_found_error?]]) do
```
zachdaniel, line 2158: Shouldn't this be `if Enum.any?` ?

StephanH90: both options default to `true`. Using `Enum.any?` would mean that you have to pass both `error?: false` and `not_found_error?: false` which is not what we want.

Using `Enum.all?` means that if you pass `false` to `error?` or `not_found_error?` will have the desired outcome of disabling the error.

I assume that's what you want.

zachdaniel: Ah, I forgot that the default was `true` 👍

zachdaniel:

PR #2576, opened 2026-02-19
feat: add support for data layers with partial success in bulk_create

rbino:
This is necessary to support databases that can encapsulate multiple "transactions" that can fail or succeed independently inside a single batch request (e.g. TigerBeetle: https://docs.tigerbeetle.com/coding/linked-events/).

To do so:
- Add a new possible result to the bulk_create callback, {:partial_success, errors, results}
- Add a new data layer capability :bulk_create_with_partial_success. This is useful to avoid stopping on bulk create errors for this kind of data layers. Otherwise, the user would be required to always pass `stop_on_error?: false`, and if they forget to do so they risk losing track of the successful results that were actually persisted

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/helpers/bulk.ex
+++ b/lib/ash/actions/helpers/bulk.ex
@@ -90,6 +90,11 @@ defmodule Ash.Actions.Helpers.Bulk do
   errors (instead of at the first individual error). This allows us to return
   all results from that batch, including successfully persisted records.
 
+  When stopping, notifications accumulated in `{:bulk_notifications, ref}` are
+  handled via `build_batch_partial_success_result/3`, which calls
+  `return_notifications_or_notify/2` to either return them or send them
+  immediately depending on `notify?` and `return_notifications?` options.
+
   Designed for use with `Stream.transform/3`. Returns `{[batch_results_list],
   acc ++ batch_results_list}` to continue processing, or throws
   `{:batch_partial_success, accumulated ++ batch_results_list, ref}` to signal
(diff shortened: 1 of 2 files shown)


zachdaniel:
🤔 this is nice! I do think that we still need to stop on error here, its just that we need to collect any errors that were generated even if its more than one.

rbino:
Makes sense, although I think that rather than collecting only errors we also have to return the successful results (since there might be some, and they are persisted).

So I think I have to leave the code as is (since we still don't want to throw with a single error if the Data Layer supports partial success) _and_ I have to find another insertion point (that has access to all error and results) where we check `stop_on_error?` and possibly early return there.

@zachdaniel does this sound right?

zachdaniel:
Hmm...my read of this though is that it would continue on to the next batch for example now. So yes we need to collect all the successes, but we then need to ensure that `stop_on_error` is honored by stopping after we do that. Am I misreading that?

rbino:
Yes, that's the high level idea: if the Data Layer can `:bulk_create_with_partial_success`, instead of stopping at the first _error_ (returning only that), we stop at the first _batch that contains an error_, and we return all results (both records and errors) up to that point

zachdaniel:
Is there a test that shows that the next batch won't be processed? Because without the `throw` in there, I'm not understanding how it doesn't just move on and process the next batch when there is an error in the first one for example.

rbino:
Sorry, I think I wasn't clear: the current code _does not_ work as described yet, I was asking for confirmation on the approach. I'll ping you as soon as it's implemented.

zachdaniel:
Ohhhhh yes. That is how it should work in this case 👌

rbino:
@zachdaniel should be ready for review now, the only part I'm not sure about is Notifications. I've more or less copied what the other `error` clause does (the single error one), but since I _could_ also have some successfull records should I call `notify` here (or return notifications) instead?

zachdaniel:
We should notify for any successes and/or return notifications, yes.  We should make sure we have a test that covers that scenario as well.

rbino:
@zachdaniel done, have a look at the last 2 commits to see the isolated changes

zachdaniel:

PR #2578, opened 2026-02-20
Enhance Ash.Type.NewType custom constraints to support arrays

meeq:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Context

Follow-up to https://github.com/ash-project/ash/pull/2575 which fixes support for custom constraints added on Ash.Type.NewType arrays.

Also enhances the test cases to ensure correctness when used with Ash.Resource

Sorry for the churn here, I found this gap after the changes from the PR had already been merged.

### DIFF
--- a/lib/ash/type/new_type.ex
+++ b/lib/ash/type/new_type.ex
@@ -132,7 +132,7 @@ defmodule Ash.Type.NewType do
         Code.ensure_compiled!(subtype_of)
       end
 
-      @compile {:inline, get_constraints: 1}
+      @compile {:inline, custom_constraint_keys: 0, custom_constraints: 1, subtype_constraints: 1}
 
       if is_nil(subtype_of) do
         raise "Must supply `:subtype_of` option when using #{mod}"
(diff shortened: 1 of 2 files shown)


lib/ash/type/new_type.ex
```diff
@@ -428,14 +430,25 @@ defmodule Ash.Type.NewType do
...
+          value,
+          get_constraints(constraints)
+        )
       end
 
       @impl Ash.Type
       def apply_constraints_array(value, constraints) do
-        constraints = get_constraints(constraints)
-        Ash.Type.apply_constraints({:array, unquote(subtype_of)}, value, items: constraints)
+        with {:ok, results} <-
```
zachdaniel: I don't think we can approach it this way, because the underlying type has to have its `apply_constraints_array` called. `apply_constraints_array` isn't necessarily an optimization, it could be used to write a type that is like " a list of maximum 10 items" for example, so it needs to see all of the items. So we'd probably want to apply the array constraints first, then apply the custom ones.

zachdaniel:

PR #2583, opened 2026-02-23
#2420 ash.setup error log

celeste-wahlquist:
# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [x] Update dependencies

### DIFF
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -1514,6 +1514,15 @@ defmodule Ash.Actions.Create.Bulk do
             end
 
           {:error, error} ->
+            if opts[:transaction] && opts[:rollback_on_error?] do
+              if Ash.DataLayer.in_transaction?(changeset.resource) do
+                Ash.DataLayer.rollback(
+                  changeset.resource,
+                  error
+                )
+              end
+            end
+
             [{:error, error, changeset}]
         end
 
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -2197,6 +2197,15 @@ defmodule Ash.Actions.Destroy.Bulk do
             end
 
           {:error, error} ->
+            if opts[:transaction] && opts[:rollback_on_error?] do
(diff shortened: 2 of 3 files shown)


celeste-wahlquist:
Solutions begins with an empty list, with each task compilation a callback is made to see if error list is empty or if it contains a failure. If failure arises, information regarding the error raised, location of error, and additional context is provided. 

Example of error log output: 

[ash.setup] lib/app.ex:10 MyApp.foo/0
  context: %{argv: [], step: :compile}
  error: ** (CompileError) lib/app.ex:10: undefined function foo/0
  
  See ash.setup_text.ex for testing of solution

lib/mix/tasks/ash.setup.ex
```diff
@@ -5,28 +5,197 @@
...
+
+  defp get_extension_name(extension) do
+    if function_exported?(extension, :name, 0) do
+      extension.name()
+    else
+      inspect(extension)
+    end
+  end
+
+  defp record_extension_failure(extension, argv, error, stacktrace) do
```
zachdaniel: Extracting information like this from stack traces is really brittle. I think it's much better to simply show which extension failed (we know because we're running it), and its associated error/stack trace. Code often moves around, functions get refactored or renamed etc. and this type of logic often becomes more of a liability than a benefit.

celeste-wahlquist: Thank you! I'll get right on that 🫡

celeste-wahlquist:
Return variables resolved: 
        Enum.each(records, &print_failure_record/1)
        Mix.raise("ash.setup failed with #{length(records)} error(s). See above.")
       
Function will return file that did not run and mix error only :3

zachdaniel:
You've included a whole bunch of unrelated files in this PR :)

celeste-wahlquist:
My apologies for the clutter! I accidentally included some local scratch/helper files in that last push. I'm going to clean up the branch, ensure it's rebased against the latest main, and force-push a clean version with only the relevant changes shortly.

PR #2584, opened 2026-02-24
feat: Add offset option to has_one and has_many relationships

### LINKED ISSUE #1740: Add offet / limit to has_many and has_one relationships

I was hoping to create a relationship that would grab the _second newest_ record in a list without needing to create a Manual Relationship. My attempts thus far have failed because the `has_one` relationship DSL lacks an `offset` function. 

I'd like it if `offset` was included in the `has_many` and `has_one` relationships, and if `limit` was included in the `has_many` relationship.

I tried using `read_action`, pointed at a read action that used a query preparation to add sort, limit, and offset. Alas it seems to have gotten stripped out because it's just not intended to work that way. In this case, since it was a `has_one` relationship and the final query didn't include the `limit` clause, Ash returned an error saying it received multiple results when it expected only one. 

My next approach will be to use a Manual Relationship, as suggested by Zach! But I'd love to have it in the DSL because as I understand it that means I can compose it more easily with other data layer operations, plus it just looks nice and is easier to do.

```elixir
    has_one :previous_log, Log do
      sort inserted_at: :desc
      offset 1
    end

    has_many :ten_previous_logs, Log do
      sort inserted_at: :desc
      offset 1
      limit 10
    end
```

nallwhy:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary
- Add `offset` option to `has_one` and `has_many` relationship DSL
- `offset` on `has_one` implies `from_many?: true` to ensure lateral joins

Closes #1740

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -56,7 +56,7 @@ To find the right module to edit:
 
 When making DSL documentation improvements, make sure to:
 1. Edit the appropriate DSL definition module (not generated docs)
-3. Test that your changes generate correctly by running `mix docs`
+2. Test that your changes generate correctly by running `mix docs`
 
 ## Rules
 


zachdaniel:
@nallwhy this looks good, but we will also need to add this option to `ash_sql` for joining and for `ash_postgres` in its lateral join code.

PR #2587, opened 2026-02-25
feat: add pre_flight_authorization() builtin validation for conditional checks during Ash.can?

emadshaaban92:
-  Set pre_flight_authorization? in context earlier so it's available during validation, rather than setting it after subject construction.
- Use that private context to implement pre_flight_authorization()

Btw is there a reason context is ignored here? 

```elixir
%Ash.Changeset{} = changeset ->
    changeset
    |> Ash.Changeset.set_tenant(opts[:tenant] || changeset.tenant)
    |> Ash.Changeset.set_context(%{private: %{actor: actor}})
```

Also for %Ash.ActionInput{} and %Ash.Query{}

I know the caller can add whatever context they want in the subject itself when they call Ash.can like this:

```elixir
Ash.can?(changeset |> Ash.Changeset.set_context(...), actor)
```

But the context is ignored if sent like this, which isn't what I expected since context almost always get merged in other places.
```elixir 
Ash.can?(changeset, actor, context: ...)
```

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/validation/builtins.ex
+++ b/lib/ash/resource/validation/builtins.ex
@@ -404,4 +404,22 @@ defmodule Ash.Resource.Validation.Builtins do
   def argument_in(argument, list) do
     {Validation.ArgumentIn, argument: argument, list: list}
   end
+
+  @doc """
+  Validates that the action is being run in a pre-flight authorization context (i.e. `Ash.can?/3`).
+
+  Primarily meant for use in `where` to skip or enable validations during authorization checks.
+
+  ## Examples
+
+      # Skip an expensive validation during can/can? checks
+      validate expensive_check(), where: [negate(pre_flight_authorization())]
+
+      # Only run a validation during can/can? checks
+      validate some_check(), where: [pre_flight_authorization()]
+  """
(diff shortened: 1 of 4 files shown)


lib/ash/resource/validation/builtins.ex
```diff
@@ -404,4 +404,22 @@ defmodule Ash.Resource.Validation.Builtins do
...
+  ## Examples
+
+      # Skip an expensive validation during can/can? checks
+      validate expensive_check(), where: [negate(pre_flight())]
+
+      # Only run a validation during can/can? checks
+      validate some_check(), where: [pre_flight()]
+  """
+  @spec pre_flight() :: Validation.ref()
+  def pre_flight do
```
zachdaniel: Let's call this `pre_flight_authorization` since `pre_flight` is a bit of a vague name.

emadshaaban92: Done

zachdaniel:

PR #2590, opened 2026-02-25
feat: add touch_update_defaults? option to bulk_create

mbaertschi:
Allows callers to skip updating update_default fields (like updated_at) when a bulk_create upsert results in an update. Supported in ETS and Mnesia data layers.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/data_layer/ets/ets.ex
+++ b/lib/ash/data_layer/ets/ets.ex
@@ -1441,13 +1441,17 @@ defmodule Ash.DataLayer.Ets do
 
   @doc false
   @impl true
-  def upsert(resource, changeset, keys, identity, opts \\ [from_bulk_create?: false]) do
+  def upsert(resource, changeset, keys, identity \\ nil) do
+    do_upsert(resource, changeset, keys, identity)
+  end
+
+  defp do_upsert(resource, changeset, keys, identity, from_bulk_create? \\ false) do
     pkey = Ash.Resource.Info.primary_key(resource)
     keys = keys || pkey
 
     if (is_nil(identity) || !identity.nils_distinct?) &&
          Enum.any?(keys, &is_nil(Ash.Changeset.get_attribute(changeset, &1))) do
-      create(resource, changeset, opts[:from_bulk_create?])
+      create(resource, changeset, from_bulk_create?)
     else
       key_filters =
         Enum.map(keys, fn key ->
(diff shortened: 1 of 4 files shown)


lib/ash/data_layer/mnesia/mnesia.ex
```diff
@@ -652,6 +661,25 @@ defmodule Ash.DataLayer.Mnesia do
...
+  # them as already set and skips them.
+  defp apply_upsert_update_defaults(to_set, resource, existing_record, opts) do
+    if opts[:touch_update_defaults?] == false || to_set == [] do
+      update_default_attrs =
+        resource
+        |> Ash.Resource.Info.attributes()
+        |> Enum.filter(& &1.update_default)
+
+      Enum.reduce(update_default_attrs, to_set, fn attr, acc ->
+        Keyword.put(acc, attr.name, Map.get(existing_record, attr.name))
```
zachdaniel: It's possible to be updating those attributes manually though, so this needs to account for that which I'm not sure this does? i.e if I did `Ash.Changeset.force_change_attribute(changeset, :updated_at, <some_value>)`

lib/ash/data_layer/ets/ets.ex
```diff
@@ -1519,6 +1522,35 @@ defmodule Ash.DataLayer.Ets do
...
+          acc
+        else
+          value =
+            case attr.update_default do
+              function when is_function(function) -> function.()
+              {m, f, a} when is_atom(m) and is_atom(f) and is_list(a) -> apply(m, f, a)
+              value -> value
+            end
+
+          Keyword.put(acc, attr.name, value)
```
zachdaniel, line 1560: Similar to the amnesia comment, we'd want to include any explicitly changed fields here.

lib/ash/data_layer/mnesia/mnesia.ex
mbaertschi: you are right, this should be fixed now

lib/ash/data_layer/ets/ets.ex
mbaertschi: you are right, this should be fixed now

mbaertschi:
hei @zachdaniel besides the two issues you mentioned, I also added this feature to the non-bulk upsert path. Do you agree? I have to backport this to the ash_postgres and ash_sqlite data layers as well

mbaertschi:
https://github.com/ash-project/ash_sqlite/pull/208 and https://github.com/ash-project/ash_postgres/pull/701 now both respect the non-bulk path as well

zachdaniel:
I like the addition of the feature, but for *non* bulk I'd prefer it to be communicated via the changeset context. Adding an `opts` arg for arbitrary stuff isn't something I'd prefer to do in the data layer behavior if it can be avoided.

zachdaniel (approved):

zachdaniel:

PR #2591, opened 2026-02-25
feat: add generate_rest? option to seed_generator

robesris:
# Contributor checklist

- [ ] I accept the AI Policy, or AI was not used in the creation of this PR.
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Problem

`seed_generator` in its tuple form `{Resource, %{attrs}}` generates random values for all writable, non-generated attributes, even those not listed in the attributes map. This makes it easy to accidentally introduce nondeterminism into test generators — any attribute not explicitly pinned will receive a random value, which can cause flaky tests that are difficult to trace back to the generator.

## Solution

Adds a `generate_rest?` option to `seed_generator/2`. When `false`, only attributes explicitly listed in the attributes map (and overrides) receive generated values. Unlisted attributes receive their resource-level default or `nil` (if `allow_nil?`). Raises with a clear message if a required attribute (`allow_nil?: false`, no default) is not provided, catching missing generators early.

```elixir
seed_generator(
  {MyApp.Document, %{name: StreamData.string(:alphanumeric)}},
  generate_rest?: false
)
```

Defaults to `true` to preserve existing behavior.

### DIFF
--- a/lib/ash/generator/generator.ex
+++ b/lib/ash/generator/generator.ex
@@ -401,6 +401,10 @@ defmodule Ash.Generator do
   * `:context` - Passed through to the changeset
   * `:after_action` - A one argument function that takes the result and returns
     a new result to run after the record is created.
+  * `:generate_rest?` - If `false`, only generates values for attributes explicitly provided
+    in the attributes map (and overrides). Unlisted attributes receive their resource-level
+    default or `nil` (if `allow_nil?`). Raises if a required attribute with no default is
+    missing. Only applies to the tuple form `{Resource, attrs}`. Defaults to `true`.
   """
   @spec seed_generator(
           Ash.Resource.record()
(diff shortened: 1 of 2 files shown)


lib/ash/generator/generator.ex
```diff
@@ -447,21 +451,28 @@ defmodule Ash.Generator do
...
-                  resource
-                )
-              ),
-              false,
-              :create,
-              []
-            )
+            merged = Map.merge(attributes, Map.new(opts[:overrides] || %{}))
+
+            if opts[:generate_rest?] == false do
```
zachdaniel: I'd flip the conditional, and use `if Keyword.get(opts, :generate_rest?, true) do` to make the access default to `true`.

lib/ash/generator/generator.ex
```diff
@@ -502,15 +513,22 @@ defmodule Ash.Generator do
...
-          |> Enum.reject(&(!&1.writable? && &1.generated?))
-          |> generate_attributes(
-            to_generators(Map.merge(attributes, Map.new(opts[:overrides] || %{}))),
-            false,
-            :create,
-            []
-          )
+          merged = Map.merge(attributes, Map.new(opts[:overrides] || %{}))
+
+          if opts[:generate_rest?] == false do
```
zachdaniel: Same here.

lib/ash/generator/generator.ex
```diff
@@ -1123,6 +1141,23 @@ defmodule Ash.Generator do
...
 
+  defp validate_required_attributes(resource, generators, action_type) do
+    resource
+    |> Ash.Resource.Info.attributes()
+    |> Enum.reject(&(!&1.writable? && &1.generated?))
+    |> Enum.each(fn attribute ->
+      default =
+        if action_type == :create, do: attribute.default, else: attribute.update_default
+
+      if !attribute.allow_nil? && is_nil(default) && !Map.has_key?(generators, attribute.name) do
```
zachdaniel: There is a bit more to check here.  If the attribute is a create or update action, then we have to check if it's in `allow_nil_input` (overrides required attributes) or `require_attributes` (makes an input required that would not normally be required). Otherwise this looks great!

lib/ash/generator/generator.ex
robesris: Addressed 95b7b79b9

lib/ash/generator/generator.ex
robesris: Addressed 7aeced878

lib/ash/generator/generator.ex
robesris: Addressed 95b7b79b9

PR #2603, opened 2026-03-02
fix: limit struct field inference to embedded resources

jechol:
## Summary
- Restrict `Ash.Type.Struct` field inference to `instance_of` values that are Ash embedded resources.
- Keep non-embedded resources on explicit field definitions instead of inferring from resource attributes.
- Add regression tests for both embedded inference and non-embedded non-inference behavior.

## Verification
- `mix test test/type/struct_test.exs test/type/struct_inference_test.exs`
- `mix test`

### DIFF
--- a/lib/ash/type/struct.ex
+++ b/lib/ash/type/struct.ex
@@ -160,13 +160,27 @@ defmodule Ash.Type.Struct do
         fields
 
       :error ->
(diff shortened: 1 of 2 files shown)


zachdaniel:
🤔 did you experience a bug due to the old behavior?

PR #2606, opened 2026-03-03
chore: add failing test for sum on count

StephanH90:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Description

An aggregate of an aggregate bypasses authorization as shown in failing tests. Aggregates of aggregates work correctly when using the new resource based approach, but they are broken when using relationships. 

This isn't such a strange use-case if you imagine something like this:

```ex
defmodule Document do
  relationships do
    belongs_to :category, Category
  end

  policies do
    policy action(:read) do 
      # imagine a filter policy here where I can only see 2 of 4 documents in a given category or child category
    end
  end
end

defmodule Category do
  relationships do
    belongs_to :parent_category, Category
    has_many :child_categories, Category
    has_many :documents, Document
  end

  aggregates do
    count :count_of_documents, :documents

    # This aggregate fails because it skips policies
    count :sum_of_count_of_documents_of_child_categories, :child_categories, :count_of_documents # I know the naming is a bit weird
  end
end
```

I added tests that re-use the existing fixtures in the `aggregate_tests.exs` file.

### DIFF
--- a/test/actions/aggregate_test.exs
+++ b/test/actions/aggregate_test.exs
@@ -64,12 +64,28 @@ defmodule Ash.Test.Actions.AggregateTest do
       belongs_to :post, Ash.Test.Actions.AggregateTest.Post do
         public?(true)
       end
+
+      belongs_to :parent, Ash.Test.Actions.AggregateTest.Comment do
+        public? true
+      end
+
+      has_many :children, Ash.Test.Actions.AggregateTest.Comment do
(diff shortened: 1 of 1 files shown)


zachdaniel:
Thank you, I've ported and fixed to `main` 🙇

PR #2608, opened 2026-03-04
fix: redact sensitive values in built-in validation errors

### LINKED ISSUE #1879: Redact sensitive values in built in validation error
**Describe the bug**
At the moment, using `Ash.Resource.Validation.Builtins.confirm` includes the comparison value in `value` when returning an error
It does this even if the attribute is set as `sensitive: true`

**To Reproduce**
A minimal set of resource definitions and calls that can reproduce the bug.
```elixir
defmodule MyApp.MyResource do
  use Ash.Resource,
    otp_app: :my_app,
    data_layer: AshPostgres.DataLayer

actions do
update :confirm_phone_number do
      require_atomic? true

      argument :input_phone_number_confirmation_code, :string, allow_nil?: false, sensitive?: true

      validate confirm(:input_phone_number_confirmation_code, :phone_number_confirmation_code)
end
end

  attributes do
    attribute :phone_number_confirmation_code, :string do
      allow_nil? true
      sensitive? true
    end
end
```

Will get a changeset like:
```elixir
    arguments: %{phone_number_confirmation_code: "incorrect"},
    errors: [
      %Ash.Error.Changes.InvalidAttribute{
        field: :phone_number_confirmation_code,
        message: "confirmation did not match value",
        private_vars: nil,
        value: "correct",
        splode: nil,
        bread_crumbs: [],
        vars: [],
        path: [],
        class: :invalid
      }
    ],
```

**Expected behavior**
Something like:
```elixir
    arguments: %{phone_number_confirmation_code: "incorrect"},
    errors: [
      %Ash.Error.Changes.InvalidAttribute{
        field: :phone_number_confirmation_code,
        message: "confirmation did not match value",
        private_vars: nil,
        value: "REDACTED",
        splode: nil,
        bread_crumbs: [],
        vars: [],
        path: [],
        class: :invalid
      }
    ],
```

**Runtime**
 - Elixir version: 18.2
 - Erlang version: 27.3
 - OS: Ubuntu
 - Ash version: 3.4.68

**Additional context**

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary
- Built-in validations leaked actual field values in error structs even when the field was marked `sensitive?: true`
- Added `sensitive?/2` and `maybe_redact/3` shared helpers to `Ash.Resource.Validation` for checking sensitivity and redacting values
- Applied redaction to all 11 built-in validations that include values in errors
- Both non-atomic (`validate/3`) and atomic (`atomic/3`) code paths are covered

Closes #1879

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -43,6 +43,7 @@ config :ash, :custom_expressions, [Ash.Test.Expressions.JaroDistance]
 config :ash, :keep_read_action_loads_when_loading?, false
 config :ash, :read_action_after_action_hooks_in_order?, true
 config :ash, :bulk_actions_default_to_errors?, true
+config :ash, :redact_sensitive_values_in_errors?, true
 
 config :crux, :sat_testing, true
 config :ash, :no_join_mnesia_ets, :dynamic
--- a/documentation/topics/advanced/manual-installation.md
+++ b/documentation/topics/advanced/manual-installation.md
@@ -147,8 +147,9 @@ Update `config/config.exs`:
 +   default_actions_require_atomic?: true,
 +   read_action_after_action_hooks_in_order?: true,
 +   bulk_actions_default_to_errors?: true,
-+   transaction_rollback_on_error?: true
-+ 
++   transaction_rollback_on_error?: true,
++   redact_sensitive_values_in_errors?: true
++
   config :spark,
     formatter: [
 ...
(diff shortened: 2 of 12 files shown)


zachdaniel:
@nallwhy We're going to put this behind a backwards compatibility flag because the values being hidden here can change the type of a value that people may be relying on from the error message.  Otherwise: this is an awesome change :sunglasses

nallwhy:
Thanks for the feedback!

Contributing to Ash has become remarkably smooth with AI assistance. I'm planning to keep picking up issues like this going forward.

zachdaniel:
LGTM, just need to adjust the flag in test 😄

zachdaniel:

PR #2610, opened 2026-03-04
fix: use individually overridden callbacks within array callbacks

jsw800:
I have a few types that override `cast_stored` and `dump_to_native` to override how things are stored in my DB layer (for backwards compatibility reasons, I need some things to exist in the DB one way while being a bit different in elixir/API layer).

For individual records, this works great. But, when serializing an array of records (like in an attribute that is type `{:array, MyTypeWithOverriddenCallbacks}`), the records in the arrays weren't being dumped right, they were being dumped/loaded the default way.

It turns out that the default `*_array` callback implementations for `NewType`s use the `subtype_of` value for the individual callbacks, which is why things aren't being dumped right in this case.

I wrote this up with Claude's help, which I say not to abdicate responsibility for the code, but rather just to be transparent + point out that, while I'm fairly sure this does fix the problem, I'm not 100% sure this is the best way to fix the problem—this essentially replicates the entire body of the Ash.Type functions, which definitely feels a bit janky. Very open to feedback on better ways to accomplish this.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/new_type.ex
+++ b/lib/ash/type/new_type.ex
@@ -214,15 +214,48 @@ defmodule Ash.Type.NewType do
       end
 
       @impl Ash.Type
-      def cast_input_array(value, constraints) do
-        subtype_constraints = subtype_constraints(constraints)
+      def cast_input_array(nil, _), do: {:ok, nil}
 
-        with {:ok, value} <- unquote(subtype_of).cast_input_array(value, subtype_constraints) do
-          Ash.Type.apply_constraints(
-            {:array, unquote(subtype_of)},
-            value,
-            items: subtype_constraints
-          )
+      def cast_input_array(term, constraints) do
+        term
+        |> Stream.with_index()
+        |> Enum.reduce_while({:ok, []}, fn {item, index}, {:ok, casted} ->
+          case cast_input(item, constraints) do
+            :error ->
+              {:halt, {:error, message: "invalid value at %{index}", index: index, path: [index]}}
+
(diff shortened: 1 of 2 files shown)


lib/ash/type/new_type.ex
```diff
@@ -276,9 +309,38 @@ defmodule Ash.Type.NewType do
       end
 
       @impl Ash.Type
-      def cast_stored_array(value, constraints) do
-        constraints = subtype_constraints(constraints)
-        Ash.Type.cast_stored({:array, unquote(subtype_of)}, value, items: constraints)
+      def cast_stored_array(term, constraints) do
```
zachdaniel, line 312: The benefit of this: `Ash.Type.cast_stored({:array, unquote(subtype_of)}, value, items: constraints)` is that it ensures if the underlying type has its own `cast_stored_array` that it will be called. I think we may need to make this a configuration or additional callback indicating that the `cast_stored_array` of the underlying type should be ignored in favor of iteratively calling the custom `cast_stored`? Or maybe we need to detect that `cast_stored` was overridden and only do this behavior when that happens?

jsw800: When you say "underlying type" are you referring to the type that the new type is a subtype of? e.g. Ash.Type.Struct, etc? I'm not sure I'm totally following

zachdaniel:
nvm after re-reading this looks right to me 😄

PR #2611, opened 2026-03-04
feat: add mix ash.ecto_compat task to detect Ecto schema compatibilit…

### LINKED ISSUE #769: Actor option ignored/overwritten.
It seems that when you provide the `actor` option to code interfaces or API action functions it is ignored, and only setting it with `Ash.Changeset.for_(create|read|update)` works.  Maybe not always?

See https://elixirforum.com/t/relate-actor-results-in-could-not-relate-to-actor-as-no-actor-was-found-and-allow-nil-is-false/59534/3

RylandBangerter85:
## Summary

- Adds `mix ash.ecto_compat` diagnostic task that detects compatibility gaps between Ash resources and their generated Ecto schemas
- Detects Ash timestamps (`create_timestamp`/`update_timestamp`) not registered as Ecto autogenerate fields
- Detects static Ash attribute defaults not present on the Ecto struct
- Supports checking a single resource (`--resource`), all resources in a domain (`--domain`), or auto-discovery from configured domains

Closes #769

## What it does

Ash resources generate Ecto schemas automatically, but two things don't translate:

1. **Timestamps**: `create_timestamp`/`update_timestamp` are not registered in `__schema__(:autogenerate_fields)`, so `Repo.insert/1` won't auto-populate them
2. **Static defaults**: `default 1` is applied by Ash's changeset pipeline, not baked into the Ecto struct, so `%MyResource{}` shows `nil`

This task surfaces those gaps so developers mixing Ash API calls with direct Repo calls can make informed decisions.

## Files

- `lib/ash/ecto_compat.ex` — Core detection logic
- `lib/mix/tasks/ash.ecto_compat.ex` — Mix task CLI
- `test/resource/ecto_compat_test.exs` — 7 tests (all passing)
- `documentation/topics/development/development-utilities.md` — User-facing docs

## Test plan

- [x] 7 unit tests covering: no-issue baseline, timestamp detection, default detection, error handling for non-Ash modules, error handling for unloaded modules, output formatting
- [x] Full test suite passes (148 failures are all pre-existing Picosat-related, unrelated to this change)

# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/topics/development/development-utilities.md
+++ b/documentation/topics/development/development-utilities.md
@@ -22,6 +22,62 @@ All Ash packages that ship with extensions provide exports in their `.formatter.
 
 Ash uses [Spark](https://hexdocs.pm/spark) to build all of our DSLs (like `Ash.Resource` and `Ash.Domain`) and to validate options lists to functions. `Spark` ships with an extension that is automatically picked up by ElixirLS to provide autocomplete for all of our DSLs, and options list. You don't need to do anything to enable this, but it only works with ElixirLS (not other language server tools).
 
+## Ecto Compatibility Checker
+
+Ash resources generate Ecto schemas automatically, but there can be differences between how Ash handles defaults and timestamps versus how Ecto handles them. The `mix ash.ecto_compat` task helps you identify these differences.
+
+### Usage
+
+Check all resources in your configured domains:
+
+```bash
+mix ash.ecto_compat
+```
+
+Check a specific resource:
+
+```bash
+mix ash.ecto_compat --resource MyApp.User
+```
+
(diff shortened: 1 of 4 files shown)


zachdaniel:
Hey @RylandBangerter85 this does indeed seem quite useful, but I'm not convinced it's the right call for this to be built into core. It's a useful utility, but I'd prefer to find some way to build towards ecto compatibility, contribute functional defaults to ecto etc. If what we're talking about is a net new task that can be helpful for those that find it, I think a separate hex package would be more appropriate in this instance. Thank you!

PR #2612, opened 2026-03-05
fix: use Enum.member? instead of in-macro to mitigate 1.20-rc.2 compiler warnings

Torkan:
The compiler in Elixir 1.20-rc.2 infers that `@all_attributes` is an empty list when doing some inference checks, and concludes that it must always be false, and as a consequence gives a compile warning.

I suspect this is due to the in macro expanding before Spark has had the opportunity to hydrate the module attribute, but I'm far from sure.

Using `Enum.member?` instead however, stops the compiler warnings.

# Contributor checklist


zachdaniel (approved):

zachdaniel:

PR #2613, opened 2026-03-05
fix: remove unreachable is_list guard in soft destroy path

djgoku:
Dialyzer flagged that `is_list(notifications)` in the `{:ok, notifications}` branch can never succeed because `Update.run` always returns a struct (record) in its 2-tuple form, never a list. The remaining branches already handle all return type conversions correctly.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #2615, opened 2026-03-07
feat: add mix ash.gettext.extract for i18n support

### LINKED ISSUE #1771: Generate gettext file for various error messages
Being able to translate error messages into different languages is quite important as most of the world speaks another language, natively.

Generate a gettext file that can possibly be updated with a mix task, that will create a POT file so we can generate the translation for the error messages.

Maybe even for field names, we can also generate a "humanized" name.

Would need to manually create the file OR scan the deps of ash and generate the file.

For example

```bash
 mix ash.codegen.gettext
```

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

Closes #1771 .

- Add `Ash.Gettext` module with `error_message/1` macro — compiles to the bare string, serves as a marker for gettext extraction
- Migrate all 16 validation modules from hardcoded strings to `error_message("...")` calls
- Add `mix ash.gettext.extract` task that walks source AST to extract marked strings into `priv/gettext/ash.pot` with source references
- Include `priv/` in hex package files so the `.pot` is available as a dependency artifact

## How users translate Ash error messages

1. `mix ash.gen.gettext` — copies `ash.pot` from the Ash dependency into your app's `priv/gettext/`
2. `mix gettext.merge priv/gettext` — creates `priv/gettext/<locales>/LC_MESSAGES/ash.po`
3. Fill in `msgstr` values in the generated `.po` file
4. In `translate_error/1`:
```elixir
Gettext.dgettext(MyApp.Gettext, "ash", msg, vars)
```

## CI consideration

Ash CI should verify the `.pot` stays in sync with source:
```bash
mix ash.gettext.extract
git diff --exit-code priv/gettext/ash.pot
```

### DIFF
--- a/lib/ash/resource/validation.ex
+++ b/lib/ash/resource/validation.ex
@@ -156,7 +156,6 @@ defmodule Ash.Resource.Validation do
       @before_compile Ash.Resource.Validation
 
       import Ash.Expr
-      import Ash.Gettext
       require Ash.Query
 
       @impl true
--- a/lib/ash/resource/validation/action_is.ex
+++ b/lib/ash/resource/validation/action_is.ex
@@ -5,6 +5,7 @@
 defmodule Ash.Resource.Validation.ActionIs do
   @moduledoc "Validates that the action is the specified action."
   use Ash.Resource.Validation
+  import Ash.Gettext
 
   @opt_schema [
     action: [
--- a/lib/ash/resource/validation/argument_does_not_equal.ex
+++ b/lib/ash/resource/validation/argument_does_not_equal.ex
@@ -6,6 +6,7 @@ defmodule Ash.Resource.Validation.ArgumentDoesNotEqual do
   @moduledoc false
 
   use Ash.Resource.Validation
+  import Ash.Gettext
 
   alias Ash.Error.Changes.InvalidArgument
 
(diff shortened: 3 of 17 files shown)


nallwhy:
The REUSE check is failing because `priv/gettext/ash.pot` has no copyright/license header. This file is auto-generated by `mix ash.gettext.extract` — it's a gettext template, not source code.
What's the best way to handle this?

lib/ash/resource/validation.ex
```diff
@@ -156,6 +156,7 @@ defmodule Ash.Resource.Validation do
       @before_compile Ash.Resource.Validation
 
       import Ash.Expr
+      import Ash.Gettext
```
zachdaniel: I don't think we can should assume all hand-rolled validations don't already have an `error_message/1` call. We should import it explicitly in each builtin validation.

lib/mix/tasks/install/ash.install.ex
```diff
@@ -234,6 +234,12 @@ if Code.ensure_loaded?(Igniter) do
               [:transaction_rollback_on_error?],
               true
             )
+            |> Igniter.Project.Config.configure(
```
zachdaniel, line 237: Should be fixed in `main`.

zachdaniel:
@nallwhy typically what you'd do is add an assistive "reuse" file next to the file in question. For example, you can see `.tool-versions.license`

zachdaniel:

PR #2616, opened 2026-03-07
fix: soft destroy with return_notifications? loses notifications

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

When a soft destroy action is called with `return_notifications?: true` (e.g., from `managed_relationships.ex` cleaning up related records), the notifications are silently lost, causing a `CaseClauseError`.

```elixir
Unknown Error

 * ** (CaseClauseError) no case clause matching:

     :ok

   (ash 3.19.3) lib/ash/actions/update/update.ex:277: Ash.Actions.Update.run/4
   (ash 3.19.3) lib/ash.ex:3674: Ash.update/3
```

Regression introduced in f33977d9e ("fix: remove unreachable is_list guard in soft destroy path (#2613)"). That commit removed a `when is_list(notifications)` guard from `destroy.ex` that was compensating for `update.ex` returning an incorrect tuple shape. Removing the guard exposed the underlying bug.

### Root Cause

`update.ex`'s `add_notifications/7` had destroy-specific return format handling that bypassed the normal `{:ok, result, notifications}` shape:

```elixir
# update.ex - before fix
if changeset.action_type == :destroy && !return_destroyed? do
  {:ok, notifications}  # 2-tuple: {:ok, list}
end
```

`destroy.ex`'s soft destroy handler then pattern-matched this as `{:ok, record}`, binding `notifications` to `record`, and returned `:ok` — discarding the notifications entirely:

```elixir
# destroy.ex (unchanged)
case Ash.Actions.Update.run(...) do
  {:ok, record} ->           # {:ok, notifications_list} matched here
    if opts[:return_destroyed?], do: {:ok, record}, else: :ok  # notifications lost
```

Callers like `managed_relationships.ex` expected `{:ok, notifications}` but received `:ok`, triggering `CaseClauseError`.

### Fix

Remove the `action_type == :destroy` special-casing from `update.ex`'s `add_notifications/7`. The function now always returns `{:ok, result}` or `{:ok, result, notifications}`, and `destroy.ex`'s existing soft destroy handler correctly decides the return format based on `return_destroyed?` / `return_notifications?`.

This is the right layer for the fix because:
- **Separation of concerns**: `update.ex` should not need to know about destroy return format conventions
- **Single responsibility**: `destroy.ex` already has the correct branching logic for destroy return values

### DIFF
--- a/lib/ash/actions/update/update.ex
+++ b/lib/ash/actions/update/update.ex
@@ -335,10 +335,8 @@ defmodule Ash.Actions.Update do
         changeset.resource,
         result,
         changeset.action,
-        changeset,
         instructions,
-        opts[:return_notifications?],
-        opts[:return_destroyed?]
+        opts[:return_notifications?]
       )
     end
     |> case do
@@ -403,10 +401,8 @@ defmodule Ash.Actions.Update do
          resource,
          result,
          action,
-         _changeset,
          instructions,
-         return_notifications?,
-         _return_destroyed?
+         return_notifications?
        ) do
     if return_notifications? do
       {:ok, result, Map.get(instructions, :notifications, [])}


lib/ash/actions/update/update.ex
```diff
@@ -403,25 +401,15 @@ defmodule Ash.Actions.Update do
...
          action,
-         changeset,
          instructions,
-         return_notifications?,
-         return_destroyed?
+         return_notifications?
        ) do
     if return_notifications? do
-      if changeset.action_type == :destroy && !return_destroyed? do
-        {:ok, Map.get(instructions, :notifications, [])}
```
zachdaniel, line 413: 🤔 don't we still need to support this return type? soft destroy actions should return data in the same shape as regular destroy actions

nallwhy: `Destroy.run/4` already handles `return_destroyed?` on lines 33-46 by transforming whatever `Update.run/4` returns. The old code in `add_notifications` *also* tried to handle it, and the two interfered with each other — that's exactly the bug this PR fixes.

Specifically, when `return_notifications?: true, return_destroyed?: false`:

```
Old Update.add_notifications:
  action_type == :destroy && !return_destroyed? → {:ok, notifications}

Destroy.run receives {:ok, notifications}:
  matches {:ok, record} (where record = notifications list)
  !return_destroyed? → :ok
  → notifications lost ❌
```

By removing `return_destroyed?` from Update, each layer has a single responsibility: Update always returns `{:ok, result, notifications}`, and Destroy.run transforms it based on `return_destroyed?`. No double-handling, no interference.

zachdaniel:

PR #2618, opened 2026-03-08
improvement: raise error when authorizer/notifier is added to extensions instead of proper option

### LINKED ISSUE #716: Compiler doesn't warn when using Ash.Policy.Authorizer in wrong place.
Sometimes when sleep deprived / not paying attention you may add `Ash.Policy.Authorizer` to the `extensions` list on your resource declaration (instead of in the `authorizers` list), you will then proceed to be flummoxed for the next few hours as your policies don't work - yet your resource has the policy DSL available in it.

Ideally this would throw an error.

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

Closes #716

- This catches a common mistake where the DSL becomes available (since it's a valid extension) but the actual authorization/notification logic silently doesn't run
- Adds a new verifier `VerifyExtensionsAreNotMisplaced` that raises a compile-time error when a module implementing `Ash.Authorizer` or `Ash.Notifier` is added to `extensions` instead of `authorizers`/`notifiers`

### DIFF
--- a/test/resource/verify_extensions_are_not_misplaced_test.exs
+++ b/test/resource/verify_extensions_are_not_misplaced_test.exs
@@ -1,0 +1,115 @@
+# SPDX-FileCopyrightText: 2019 ash contributors <https://github.com/ash-project/ash/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Ash.Test.Resource.VerifyExtensionsAreNotMisplacedTest do
+  @moduledoc false
+  use ExUnit.Case, async: true
+
+  defmodule PlainExtension do
+    @moduledoc false
+    use Spark.Dsl.Extension
+  end
+
+  test "raises error when an authorizer is added to extensions instead of authorizers" do
+    output =
+      ExUnit.CaptureIO.capture_io(:stderr, fn ->
+        defmodule Module.concat(["rand#{System.unique_integer([:positive])}", Post]) do
+          @moduledoc false
+
+          use Ash.Resource,
+            domain: Ash.Test.Domain,
+            data_layer: Ash.DataLayer.Ets,
+            extensions: [Ash.Policy.Authorizer]
+
+          attributes do
+            uuid_primary_key :id
+          end
+        end
+      end)
+
+    assert output =~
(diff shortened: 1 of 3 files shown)


test/resource/verify_extensions_are_not_misplaced_test.exs
```diff
@@ -0,0 +1,115 @@
...
+  use ExUnit.Case, async: true
+
+  defmodule PlainExtension do
+    @moduledoc false
+    use Spark.Dsl.Extension
+  end
+
+  test "raises error when an authorizer is added to extensions instead of authorizers" do
+    output =
+      ExUnit.CaptureIO.capture_io(:stderr, fn ->
```
zachdaniel, line 16: Could you double check that this one wasn't already solved? I could have sworn we added a warning or an error for this case before. Regardless its a good addition 😄

nallwhy: I wasn't aware that `Ash.Policy.Authorizer` already checks this via `Ash.Policy.Authorizer.Verifiers.VerifyInAuthorizers`. However, since that verifier lives inside the `Ash.Policy.Authorizer` extension itself, third-party authorizers that don't include a similar self-check verifier would silently fail.

Rather than placing the burden on each authorizer/notifier to handle its own verification, it seems better to handle this at the framework level — which is what this PR does. Should I remove `VerifyInAuthorizers` and let the new `VerifyExtensionsAreNotMisplaced` handle it for all cases?

zachdaniel: Hmm...we also have a conundrum here. When placed in a domain, the policy authorizer is placed under `extensions` IIRC. So this transformer wouldn't correctly work there.

I'm not sure we can state 100% that all extensions that *are* notifiers or authorizers can *only* be authorizers or notifiers 🤔

nallwhy: Ah, that's a good point — I didn't consider the Domain case where `extensions: [Ash.Policy.Authorizer]` is the intended usage. That breaks the assumption this verifier is based on. I'll close this.

Issue #2619, opened 2026-03-08
`Ash.can?/3` does not pass `scope`


joangavelan:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Erlang/OTP 27
Elixir 1.18.3
Ash  3.5.36

### Operating system

MacOS

### Current Behavior

I have the following policy for my `:create_establishment` action

```elixir
defmodule Lamashka.PolicyChecks.ActorIsOrgOwner do
  use Ash.Policy.SimpleCheck

  @impl true
  def describe(_options) do
    "Actor is the owner of the organization"
  end

  @impl true
  def match?(
        actor,
        %{resource: Lamashka.Establishments.Establishment, subject: %Ash.Changeset{} = changeset},
        _opts
      ) do
    changeset.tenant.organization.owner_id == actor.id
  end

  @impl true
  def match?(_actor, _context, _options) do
    false
  end
end
```

When calling `Ash.can?/3` with a `scope:` option, the scope's tenant data is `nil` causing the function call to raise, while the equivalent domain-generated `can_*?` function works correctly with the same scope.

```elixir
# This raises a KeyError because changeset.tenant is nil inside the policy check
Ash.can?({Establishment, :create_establishment}, actor, scope: scope)

# This works correctly
Establishments.can_create_establishment?(actor, scope: scope)
```

Full discussion on Discord: https://discord.com/channels/711271361523351632/1019647368196534283

PR #2621, opened 2026-03-08
fix: merge scope with actor when doing can? check

StephanH90:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Fixes https://github.com/ash-project/ash/issues/2619

I added tests, but I'm not sure if this is how you are supposed to fix it this, to be honest.

### DIFF
--- a/lib/ash/can.ex
+++ b/lib/ash/can.ex
@@ -76,6 +76,8 @@ defmodule Ash.Can do
     opts = Keyword.put_new(opts, :filter_with, :filter)
     pre_flight? = Keyword.get(opts, :pre_flight?, true)
 
+    actor_or_scope = Ash.Helpers.deep_merge_maps(actor_or_scope, opts[:scope])
+
     context =
       Ash.Helpers.deep_merge_maps(opts[:context] || %{}, %{
         private: %{pre_flight_authorization?: pre_flight?}
--- a/test/code_interface_test.exs
+++ b/test/code_interface_test.exs
@@ -3,6 +3,18 @@
 # SPDX-License-Identifier: MIT
 
 defmodule Ash.Test.CodeInterfaceTest do
+  defmodule Scope do
+    defstruct [:actor, :tenant, :context]
+
+    defimpl Ash.Scope.ToOpts do
+      def get_actor(%{actor: actor}), do: {:ok, actor}
+      def get_tenant(%{tenant: tenant}), do: {:ok, tenant}
+      def get_context(%{context: context}), do: {:ok, context}
+      def get_tracer(_), do: :error
+      def get_authorize?(_), do: :error
+    end
+  end
+
   @moduledoc false
   use ExUnit.Case, async: true
 
@@ -680,6 +692,10 @@ defmodule Ash.Test.CodeInterfaceTest do
 
     assert User.can_update_by_id?(user, user.id, %{})
     assert User.can_destroy_by_id?(user, user.id, %{})
+
+    assert User.can_destroy_by_id?(user, %{},
+             scope: %Ash.Test.CodeInterfaceTest.Scope{actor: user}
+           )
   end
 
   test "optional arguments are optional" do
@@ -735,18 +751,6 @@ defmodule Ash.Test.CodeInterfaceTest do
     # Test that function-based defaults can still be overridden
     assert "Hello, Override Actor." =
              User.hello_actor_with_function_default!(actor: %{name: "Override Actor"})
-  end
-
-  defmodule Scope do
-    defstruct [:actor, :tenant, :context]
-
-    defimpl Ash.Scope.ToOpts do
-      def get_actor(%{actor: actor}), do: {:ok, actor}
-      def get_tenant(%{tenant: tenant}), do: {:ok, tenant}
-      def get_context(%{context: context}), do: {:ok, context}
-      def get_tracer(_), do: :error
-      def get_authorize?(_), do: :error
-    end
   end
 
   defmodule ContextScope do
--- a/test/policy/simple_test.exs
+++ b/test/policy/simple_test.exs
@@ -478,6 +478,7 @@ defmodule Ash.Test.Policy.SimpleTest do
   test "Ash.can? honors a provided scope", %{admin: admin, user: user} do
     tweet = Ash.create!(Ash.Changeset.for_create(Tweet, :create), authorize?: false)
     assert Ash.can?({tweet, :read}, %Scope{actor: admin})
+    assert Ash.can?({tweet, :read}, %{}, scope: %Scope{actor: admin})
     refute Ash.can?({tweet, :read}, %Scope{actor: user})
   end
 


zachdaniel:
Hm... so looking at the code, this should be handled by this logic here:

https://github.com/ash-project/ash/pull/2621/changes#diff-6cb9a467291bbba24696c98cb231577406c0ca6569e93761e158d567356edccaL84

Can you confirm with a test only that this isn't working so I can see that test fail?

StephanH90:
I hope I understand your question correctly but if `actor = %{}` then:

```ex
 {actor, opts} =
      if is_struct(actor_or_scope) and Ash.Scope.ToOpts.impl_for(actor_or_scope) do
        opts
        |> Keyword.put(:scope, actor_or_scope)
        |> Ash.Actions.Helpers.apply_scope_to_opts()
        |> Keyword.pop(:actor)
      else
        {actor_or_scope, opts}
      end
```
will simply return `{%{}, opts}` since `is_struct(%{}) == false`.

If I uncomment my code, I get this failing test:

```ex

 1) test Ash.can? honors a provided scope (Ash.Test.Policy.SimpleTest)
     test/policy/simple_test.exs:478
     Expected truthy, got false
     code: assert Ash.can?({tweet, :read}, %{}, scope: %Scope{actor: admin})
     arguments:

         # 1
         {%Ash.Test.Support.PolicySimple.Tweet{
            id: "193653d4-7b44-41ce-9079-c4c04129252d",
            user_id: nil,
            is_foo: #Ash.NotLoaded<:calculation, field: :is_foo>,
            user: #Ash.NotLoaded<:relationship, field: :user>,
            __meta__: #Ecto.Schema.Metadata<:loaded>
          }, :read}

         # 2
         %{}

         # 3
         [
           scope: %Ash.Test.Policy.SimpleTest.Scope{
             actor: %Ash.Test.Support.PolicySimple.User{
               id: "7c54ec75-bc41-4c0f-8eea-43f94475839f",
               admin: true,
               manager: nil,
               organization_id: nil,
               most_recent_car: #Ash.NotLoaded<:aggregate, field: :most_recent_car>,
               restricted_from_driving: #Ash.NotLoaded<:calculation, field: :restricted_from_driving>,
               has_car: #Ash.NotLoaded<:calculation, field: :has_car>,
               post_texts: #Ash.NotLoaded<:calculation, field: :post_texts>,
               cars_join_assoc: #Ash.NotLoaded<:relationship, field: :cars_join_assoc>,
               organization: #Ash.NotLoaded<:relationship, field: :organization>,
               posts: #Ash.NotLoaded<:relationship, field: :posts>,
               cars: #Ash.NotLoaded<:relationship, field: :cars>,
               __meta__: #Ecto.Schema.Metadata<:loaded>
             },
             tenant: nil,
             context: nil
           }
         ]

     stacktrace:
       test/policy/simple_test.exs:481: (test)
```
where the corresponding test is:

```ex
  test "Ash.can? honors a provided scope", %{admin: admin, user: user} do
    tweet = Ash.create!(Ash.Changeset.for_create(Tweet, :create), authorize?: false)
    assert Ash.can?({tweet, :read}, %Scope{actor: admin})
    assert Ash.can?({tweet, :read}, %{}, scope: %Scope{actor: admin}) # <- this fails
    refute Ash.can?({tweet, :read}, %Scope{actor: user})
  end
```
which IMO should pass?

zachdaniel:
Gotcha. I think we should just remove that `is_struct` check 🤔

StephanH90:
But if you have this code (ie the `is_struct` check removed):

```ex
if actor_or_scope and Ash.Scope.ToOpts.impl_for(actor_or_scope) do
  ...
end
```
then this will still fail:
```ex
assert Ash.can?({tweet, :read}, %{}, scope: %Scope{actor: admin})
```

since we provide the scope with `scope: %Scope{}` and not through the `actor_or_scope` argument.

zachdaniel:
Oh, you're totally right 😄 . 

All that is missing in that case is this line after the conditional:

```elixir
opts = Ash.Actions.Helpers.apply_scope_to_opts(opts)
```

that should take any `scope: scope` and ensure its handled.

StephanH90:
That will still not work because the rest of the code only uses `actor` as far as I can tell.

If we not merge the scope with the actor on line 85 or so then its "too late". The existing code already does this (IF we pass a struct as actor):

```ex
{actor, opts} =
  if is_struct(actor_or_scope) ... other stuff ... do
    opts
    |> Keyword.put(:scope, actor_or_scope) # opts has :scope now
    |> Ash.Actions.Helpers.apply_scope_to_opts() # populate the :scope
    |> Keyword.pop(:actor)  # <- here we pop the keyword "actor" from the `opts[:scope]`
  else
    ...
  end
```

Applying scope to opts after this will still result in `{nil, opts}`. The `Keyword.pop(:actor)` actually ends up assigning the `actor`. If we don't do this for other cases where actor is not a struct (or can be passed through) any actor in scope is disregarded.

I also tried your suggestion and the test still fails.

I'm sorry for the pushback :cry:

zachdaniel:
No need to apologize, if I'm wrong I'm wrong 😄 I've been busy and not able to dive super deeply into this so been doing some drive bys. 

With that said, I'm realizing what the issue actually is 😄 . The reason this works the way that it works is because that second argument is meant to be the actor and the rest of the opts are used for other things. So this:

```elixir
assert Ash.can?({tweet, :read}, %{}, scope: %Scope{actor: admin})
```

doesn't really make a whole lot of sense as a concept because the actor is supposedly `%{}`. Technically it should be 

```elixir
assert Ash.can?({tweet, :read}, admin, opts)
```

And it looks like we tried to account for that via supporting passing a scope instead of the actor:

```elixir
assert Ash.can?({tweet, :read}, %Scope{actor: admin}, ...opts)
```

The thing that has me pushing back is that ultimately we cannot do this:

`actor_or_scope = Ash.Helpers.deep_merge_maps(actor_or_scope, opts[:scope])`

because we never merge actors or anything like that. That was the "code smell" that made me realize something wasn't quite right here. So we can theoretically support this:

```elixir
assert Ash.can?({tweet, :read}, %{}, scope: %Scope{actor: admin})
```

but that `%{}` that is being passed in as the actor would just be thrown away/ignored and that would be pretty confusing. I'm thinking that the only real change needed here is to document these thing and/or warn when a `scope` option is provided as an opt instead of as the second argument.

emadshaaban92:
I faced the same issue (slightly differently) and only noticed your PR and the issue it fixes while in the middle of writing my fix here #2631 

I think the problem in this PR is the test case? .. it tests if scope overrides actor which like Zach said would be confusing (although it's what the docs says, we might need to change the docs)

While the problem in the original issue #2619 was about tenant not being passed through scope (and context for my case) .. both cases made it clear that the problem happened when we pass both actor through `actor_or_scope` and scope through `opts`  .. so the problem is clearly in the else branch not the if.

zachdaniel:
Thanks @emadshaaban92! Appreciate you @StephanH90 for isolating the issue 🙇

PR #2622, opened 2026-03-09
fix: register Ash timestamps with Ecto's autogenerate system

### LINKED ISSUE #769: Actor option ignored/overwritten.

RylandBangerter85:
Ash resources generate Ecto schemas automatically, but create_timestamp/update_timestamp attributes were not registered with Ecto's autogenerate system. This meant __schema__(:autogenerate_fields) returned [] and Repo.insert/1 left timestamp fields as nil.

This wires Ash's timestamp attributes into Ecto's :ecto_autogenerate and :ecto_autoupdate module attributes at compile time, so that Repo.insert/1 and Repo.update/1 auto-populate them — matching what Ecto's native timestamps() macro does.

Timestamps sharing the same storage type are grouped into a single registration entry so they receive the same generated value, preserving Ash's match_other_defaults? semantics.

Closes #769

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/schema.ex
+++ b/lib/ash/resource/schema.ex
@@ -61,6 +61,8 @@ defmodule Ash.Schema do
               )
             end
 
+            Ash.Schema.register_ecto_autogenerate(__MODULE__)
+
             field(:aggregates, :map, virtual: true, default: %{})
             field(:calculations, :map, virtual: true, default: %{})
             field(:__metadata__, :map, virtual: true, default: %{}, redact: true)
@@ -197,6 +199,8 @@ defmodule Ash.Schema do
                 )
               )
             end
+
+            Ash.Schema.register_ecto_autogenerate(__MODULE__)
 
             field(:aggregates, :map, virtual: true, default: %{})
             field(:calculations, :map, virtual: true, default: %{})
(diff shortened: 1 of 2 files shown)


lib/ash/resource/schema.ex
```diff
@@ -383,4 +387,75 @@ defmodule Ash.Schema do
...
+  end
+
+  @doc false
+  def timestamp_attribute?(attr) do
+    storage_type = resolve_storage_type(attr)
+
+    ecto_datetime_type?(storage_type) &&
+      attr.writable? == false &&
+      (is_function(attr.default) || match?({_, _, _}, attr.default)) &&
+      attr.match_other_defaults? == true &&
```
zachdaniel: What we would want to do is not reject things that `match_other_defaults? == false`, but just include those as a group on their own, that way they would be included.

lib/ash/resource/schema.ex
```diff
@@ -383,4 +387,75 @@ defmodule Ash.Schema do
...
+      attr.allow_nil? == false
+  end
+
+  # Resolves the underlying Ecto storage type atom for an attribute by
+  # initializing its type constraints first. This is necessary because
+  # Ash NewTypes (like UtcDatetimeUsec) need their subtype constraints
+  # merged before storage_type returns the correct atom.
+  defp resolve_storage_type(attr) do
+    case Ash.Type.init(attr.type, attr.constraints) do
```
zachdaniel: hmm....this *should* already be done I think?

RylandBangerter85:
Okay I should have fixed those two things now

lib/ash/resource/schema.ex
```diff
@@ -383,4 +387,86 @@ defmodule Ash.Schema do
...
+        {[attr.name], {Ecto.Schema, :__timestamps__, [storage_type]}}
+      )
+    end)
+  end
+
+  @doc false
+  def timestamp_attribute?(attr) do
+    ecto_datetime_type?(Ash.Type.storage_type(attr.type, attr.constraints)) &&
+      attr.writable? == false &&
+      (is_function(attr.default) || match?({_, _, _}, attr.default)) &&
```
zachdaniel: The attr.default would need to be specifically `&DateTime.utc_now/0` to apply here.

RylandBangerter85:
Okay! The timestamp_attribute? check now specifically requires attr.default == &DateTime.utc_now/0 instead of accepting any function or MFA default.

zachdaniel:

PR #2623, opened 2026-03-11
fix: bulk action global validation ordering and before_action? support

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

- Fix global validation/change ordering in `bulk_create`, `bulk_update`, and `bulk_destroy` to match non-bulk behavior
- Add `before_action?` support for validations in the bulk create and bulk update pipelines

## Problem

Two bugs across all bulk action modules (`Ash.Actions.Create.Bulk`, `Ash.Actions.Update.Bulk`, `Ash.Actions.Destroy.Bulk`):

1. **Reversed ordering**: Global validations were concatenated before global changes in `pre_template_all_changes`.
   The non-bulk path runs action changes + global changes first, then global validations,
   but the bulk path was concatenating in action changes → global validations → global changes order.

2. **`before_action?` ignored**: The `before_action?` flag on validations was completely ignored in the bulk path, causing them to run inline.
   The non-bulk path checks this flag and defers via `Ash.Changeset.before_action/2`,
   but the bulk `run_action_changes` never read this field.

### `lib/ash/actions/create/bulk.ex`

- `pre_template_all_changes`: Reorder concat so global changes come before global validations
- `run_action_changes`: Check `before_action?` on validations; when true, register via `Ash.Changeset.before_action/2` instead of running inline
- Extract inline validation logic into `run_bulk_validation/6`
- Simplify `validation.message` error handling to reduce duplication

### `lib/ash/actions/update/bulk.ex`

- `pre_template_all_changes`: Same ordering fix as bulk_create
- `run_action_changes`: Add `before_action?` support for validations
- Simplify `validation.message` error handling in `validate_batch_non_atomically`

### `lib/ash/actions/destroy/bulk.ex`

- `pre_template_all_changes`: Same ordering fix as bulk_create
- `before_action?` support inherited via delegation to `Ash.Actions.Update.Bulk.run_action_changes`

### DIFF
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -1442,6 +1442,7 @@ defmodule Ash.Actions.Destroy.Bulk do
 
   defp pre_template_all_changes(action, resource, :destroy, base, actor, tenant) do
     action.changes
+    |> Enum.concat(Ash.Resource.Info.changes(resource, action.type))
     |> then(fn changes ->
       if action.skip_global_validations? do
         changes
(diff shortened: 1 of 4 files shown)


zachdaniel:
Can we separate these two changes? The reason for that is that `before_action` can't necessarily be supported in bulk_update actions depending on how they are running and it would be nice to consider these separately. Thanks!

PR #2627, opened 2026-03-11
feat: Add attribute_always_select? option to belongs_to relationships

emadshaaban92:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Problem
When reading records with a limited select, foreign key attributes from belongs_to relationships get masked as `Ash.NotLoaded`. If you then call an update action on that record, any change with a `where [present(:foreign_key)]` guard will silently skip — even though the value exists in the database.

# Example
Consider a Task resource that optionally belongs to a Project:

```elixir
belongs_to :project, MyApp.PM.Project, allow_nil?: true
```
When a task is completed, we want to check if the entire project is now done and update its status — but only for project-linked tasks:

```elixir
update :complete do
  change transition_state(:completed)

  change MyApp.PM.Task.Changes.AutoCompleteProject do
    where [present(:project_id)]
  end
end
```

In a dashboard LiveView, tasks are loaded with a limited select for performance:

```elixir
Task
|> Ash.Query.select([:id, :state, :title, :due_at])
|> Ash.read!()
```

When a user marks a task as complete, the action runs on the record as-is. Since `project_id` wasn't in the select, it's `Ash.NotLoaded`, `present?` returns `false`, and the `AutoCompleteProject` change is silently skipped — even for tasks that do belong to a project.

The current workaround is to either:

- Always remember to include the FK in every query's select (error-prone)
- Define the attribute manually with `always_select? true` and set `define_attribute? false` (verbose)

# Solution
Adds attribute_always_select? to belongs_to, consistent with existing attribute_writable? and attribute_public? options:

```elixir
belongs_to :project, MyApp.PM.Project,
  allow_nil?: true,
  attribute_always_select?: true
```

# Alternative / Future Consideration

## ensure_loaded or preload on actions
A complementary approach would be an `ensure_loaded` option that loads specific fields on the record before running validations and changes. Unlike `attribute_always_select?` (which is attribute-scoped and global), `ensure_loaded` would generalize to calculations, aggregates, and relationships, and could be scoped to specific actions.

```elixir
update :complete do
  ensure_loaded [:project_id]

  change transition_state(:completed)

  change MyApp.PM.Task.Changes.AutoCompleteProject do
    where [present(:project_id)]
  end
end
```

Note: Code in this PR is hand-written. The PR description was generated with AI assistance.


zachdaniel:
There is an option called `action_select` which you can use to ensure certain attributes are selected on the base record. Have you tried that? This change is useful anyway though 😄

zachdaniel:
Please make sure to run `mix spark.cheat_sheets` locally.

emadshaaban92:
> There is an option called `action_select` which you can use to ensure certain attributes are selected on the base record. Have you tried that? This change is useful anyway though 😄

I tried it first and it didn't work .. then I understood it's for the returning clause of the update query itself .. so it makes the attribute available for `after_action` hooks and not for validations or changes .. Did I get this wrong? :thinking: 



> Please make sure to run `mix spark.cheat_sheets` locally.

Done

zachdaniel:
Yep, that's a good point 😄. We can use a change for that though as opposed to a DSL item. i.e `change ensure_loaded(:foo)`

zachdaniel:

Issue #1271, opened 2024-07-01
Allow marking actions as internal to prevent exposing them via APIs


smt116:
In complex systems, there are often actions that suppose to only be used internally. For example, a way of user lookup by some custom key or creating a record. Currently, they can be added to policies with `authorize_if always()` or used only with `authorized?: false`.

Using a policy introduces a risk of exposing them in the API at some point. On the other hand, `authorized?: false` breaks the authorization circuit by not passing the current action context down the line (e.g., to nested actions).

It would be great if Ash could allow for something like `internal? true` on action definition. That property would be used when building APIs' actions, by raising a compilation error.

zachdaniel:
To align with what that means generally, we'd say something like `private? true` in the action. Although to align with 3.0 it would be something like `public? false` (and if I had thought of it before 3.0 it would default to false 😢). This seems like a good idea to me. It would be up to extensions to check that they are only using public actions, but thats how everything else already is so that should be fine.

zachdaniel:
I've labeled this as a "good first issue", as implementing it in core will be easy. Once that is done, issues will need to be opened in `AshGraphql`, `AshJsonApi`, and `AshPhoenix` (not creating forms for private actions? maybe, maybe not) to track the usage of this metadata there.

smt116:
> Once that is done, issues will need to be opened in `AshGraphql`, `AshJsonApi`, and `AshPhoenix` (not creating forms for private actions? maybe, maybe not) 

I would say that a form is still kind of an application interface. IMHO, non-public actions should never be used outside of the system. Unfortunately, it means no console usage too (without `authorized?: false`), but better be consistent than having exceptions in the logic. It should be impossible to interact with non-public actions using any kind of "human accessible interface".

zachdaniel:
Console usage would be quite hard to detect unfortunately. I'm not sure if it's reasonable.

zachdaniel:
Because if it uses process dictionary for example internal action calls by the action would think they are in the console too

smt116:
Right 🤔. Good enough

Dgoz101:
Is this still something that needs done?

zachdaniel:
Yep! We can simply add a `public?` option on actions, defaulting to `true`. Then we adjust other extensions like `AshGraphql`, `AshTypescript` and `AshJsonApi` to raise errors if you try to expose a private action.

PR #2629, opened 2026-03-12
Allow marking actions as internal to prevent exposing them via APIs #1271

Dgoz101:
# Contributor checklist

- [ X] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Chores
- [ ] Documentation changes
- [X ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

Added public? option for actions so that they can be marked as internal. Actions marked with public?: false are treated as internal and can be safely bypassed by policies while being excluded from public API generation through helpers. Default for public? is true, so any action that needs to be internal only will need to be changed to false.


zachdaniel:
Please run `mix spark.cheat_sheets` locally and commit the results 😄

Dgoz101:
> Please run `mix spark.cheat_sheets` locally and commit the results 😄

Sorry it took me a while to get to this. Got it done now.

PR #2634, opened 2026-03-13
fix: support before_action? validations in bulk_create and bulk_update

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

before_action?` validations were not supported in bulk create and bulk update paths. They ran immediately instead of being deferred to `before_action` hooks, causing them to execute before changes were applied.

This PR fixes that by:

- Wrapping `before_action?` validations in `Ash.Changeset.before_action/2` hooks in both bulk create and bulk update stream paths, consistent with non-bulk actions
- Returning `{:not_atomic, ...}` from `run_atomic_validation/3` for `before_action?` validations, so bulk update with `strategy: [:atomic, :stream]` automatically falls back to the stream strategy
- Extracting inline validation logic into `run_bulk_validation/6` helper in bulk create for clarity

## Behavior

- **bulk_create**: `before_action?` validations are wrapped in `before_action` hooks and run at the correct point in the lifecycle
- **bulk_update with `strategy: :stream`**: same as above
- **bulk_update with `strategy: [:atomic, :stream]`**: atomic path detects `before_action?` validation as `{:not_atomic, ...}` and falls back to stream strategy automatically
- **bulk_update with `strategy: :atomic` only**: fails with a clear message indicating that `before_action?` validations cannot run atomically

## ~~Current scope~~

This change only affects the **stream** (non-atomic) execution path. The `before_action?` check lives inside the `has_validate?()` branch, which is the non-atomic validation path. In the atomic path, validations go through `validate_batch_atomically` / atomic changeset expressions, which is a completely separate mechanism.

This means:
- **bulk_create**: always works (creates always process individual changesets)
- **bulk_update with `strategy: :stream`**: works
- **bulk_update with `strategy: :atomic`**: `before_action?` flag is effectively ignored — the validation runs atomically as a DB expression regardless

## ~~What should happen in the atomic path?~~

There are a few options for how to handle `before_action?` validations when bulk_update runs atomically:

**Option A: Stream-only support (current implementation)**
The atomic path doesn't touch our code at all, so `before_action?` is silently a no-op there. This is the simplest approach but could be surprising — a user might set `before_action?: true` expecting deferred execution and not realize it's ignored in atomic mode.

**Option B: Raise an error when atomic + before_action?**
Add a check in `validate_batch_atomically` that raises if `validation.before_action?` is true, with a message like _"before_action? validations cannot run atomically, use `require_atomic? false` or `strategy: [:stream]`"_. This is the most explicit approach — consistent with how Ash already handles atomic incompatibilities (e.g. `require_atomic?`).

**Option C: Auto-fallback to stream**
Detect `before_action?` validations before attempting atomic execution and automatically fall back to stream strategy. Convenient, but changes performance characteristics in a way the user might not expect.

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1130,16 +1130,22 @@ defmodule Ash.Changeset do
 
   @doc false
   def run_atomic_validation(changeset, %{where: where} = validation, context) do
-    with {:atomic, condition} <- atomic_condition(where, changeset, context) do
-      case condition do
-        false ->
-          changeset
+    if validation.before_action? do
+      {:not_atomic,
+       "before_action? validation `#{inspect(elem(validation.validation, 0))}` cannot be run atomically. " <>
+         "To use before_action? validations, set `require_atomic? false` or use `strategy: [:stream]`."}
+    else
+      with {:atomic, condition} <- atomic_condition(where, changeset, context) do
+        case condition do
+          false ->
+            changeset
 
-        true ->
-          do_run_atomic_validation(changeset, validation, context)
+          true ->
+            do_run_atomic_validation(changeset, validation, context)
 
-        where_condition ->
-          do_run_atomic_validation(changeset, validation, context, where_condition)
+          where_condition ->
+            do_run_atomic_validation(changeset, validation, context, where_condition)
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


zachdaniel:
Yeah, I think we should disallow before_action, your option B

zachdaniel:

PR #2635, opened 2026-03-13
fix: support atomic-only changes in bulk_create

barnabasJ:
## Summary

- Refactors `batch_change` in `Ash.Actions.Create.Bulk` to support change modules that only implement `atomic/3` (no `change/3` or `batch_change/3`)
- Previously, such modules would crash in `batch_change` because the fallback always called `change/3`, which doesn't exist on atomic-only modules
- Adds a `cond`-based dispatch (`has_batch_change?` → `has_change?` → `atomic?`) replacing the old 2-branch `if/else`
- Extracts `batch_atomic_change/6` helper to avoid duplication across the templated and non-templated code paths

## Test plan

- [x] Added test for `atomic_set` with per-record arguments in bulk_create (multi-record and single-record)
- [x] Added test for atomic-only change module (no `change/3`) exercising the new `atomic?` fallback path
- [x] Added test for `{:not_atomic, reason}` error handling in bulk context
- [x] All 63 existing bulk_create tests pass
- [x] `mix compile --warnings-as-errors` clean
- [x] `mix format --check-formatted` clean

### DIFF
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -2059,32 +2059,7 @@ defmodule Ash.Actions.Create.Bulk do
             end)
 
           module.atomic?() ->
-            Enum.map(batch, fn changeset ->
-              change = %{
-                change
-                | change:
-                    {module,
-                     templated_opts(
-                       change_opts,
-                       actor,
-                       changeset.to_tenant,
-                       changeset.arguments,
-                       changeset.context,
-                       changeset
-                     )}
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:
This is cool, I'm actually not even sure if atomic only changes properly work in creates? I added support for atomics in creates but I don't recall how I handled that 🤔

zachdaniel:

PR #2636, opened 2026-03-16
Implemented required_error from postgress to ash core

dallingson:
Closes Issue in postgres:
https://github.com/ash-project/ash_postgres/issues/261
More info can be found in the following PR: 
https://github.com/ash-project/ash_postgres/pull/707

Changes:
- Add core query function Ash.Query.Function.RequiredError implementing ash_required!/2, which returns the value when present or an Ash.Error.Changes.Required when it is nil.
- Add tests in test/query/required_error_test.exs covering nil/non‑nil values, string vs atom name keys, missing :resource / :name, new/1 argument validation, and metadata helpers (can_return_nil?/1, evaluate_nil_inputs?/0).
- Document the function with a short entry in documentation/topics/reference/expressions.md describing typical usage: expr(ash_required!(^value, ^attribute)) where attribute has at least :name and :resource.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1225,8 +1225,7 @@ defmodule Ash.Changeset do
          {:atomic, condition} <-
            atomic_condition(where, changeset, context),
          {{:atomic, modified_changeset?, new_changeset, atomic_changes, validations,
-           create_atomics},
-          condition} <-
+           create_atomics}, condition} <-
            {atomic_with_changeset(
               module.atomic(changeset, change_opts, struct(Ash.Resource.Change.Context, context)),
               changeset


documentation/topics/reference/expressions.md
```diff
@@ -87,6 +87,7 @@ The following functions are built in:
 - String interpolation | `"#{first_name} #{last_name}"`, is remapped to the equivalent usage of `<>`
 - `fragment/*` | Creates a fragment of a data layer expression. See the section on fragments below.
 - `error/2` | Raises an error with a given exception module and parameters. See the section on error expressions below.
+- `ash_required!/2` | Returns the given value when present, or produces a required-attribute error when it is `nil`. Typically used as `expr(ash_required!(^value, ^attribute))` where `attribute` has at least `:name` and `:resource`.
```
zachdaniel: Sorry I keep requesting changes on you! I think the function inside of Ash core itself should just be called `required!`. The one that `ash_postgres` defines inside your database should be called `ash_required` to match how we use the `ash_` prefix elsewhere.

zachdaniel:
What we would then want to do is modify the code in `Ash.Changeset` that adds atomic required errors to use this function if the data layer supports it. We can use a capability like `:required_error` and check the data layer if it supports that capability.

dallingson:
Did I get this implemented correctly?

documentation/topics/reference/expressions.md
dallingson: im just seeing this comment now... I will work on this right now

zachdaniel:
Looks like some failing checks to address 😄

zachdaniel:
Also worth checking (or having claude check) if there are any other expression patterns that use that `if is_nil(...)` that should also use `required` when the data layer supports it 😄

dallingson:
I fixed the formatting issues, and I used Claude to check to see if anything else should use ```required```. As far as I can tell, I think we are good. Is there anything else I need to do for this issue?

zachdaniel:

PR #2638, opened 2026-03-18
fix: on_match :destroy for many_to_many now destroys both join and destination

### LINKED ISSUE #859: Many to many and on match destroy in `manage_relationship`
I think `on_match: :destroy` in `manage_relationship` is incorrect for `many_to_many`.

Right now it destroys a join resource. Instead it should destroy both a join and a destination like other types of relationships do.

If one wants to destroy only a join resource then they should use `:unrelate`.

Destroy is implemented correctly for `on_missing` where `:destroy` destroys both and `:unrelate` only join one. (Though there is a mistake in documentation that states that it is possible to pass `[:join, :keys]` with a long form.) That's the only other case with support for `:destroy` so it is strange that they are inconsistent.

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

- Fixes `on_match: :destroy` for `many_to_many` relationships to destroy both the join record and the destination record, consistent with how `on_missing: :destroy` already works
- Adds a new 3-tuple form `{:destroy, :dest_action, :join_action}` for explicit control over both actions
- Adds `many_to_many_destroy_destination_on_match?` backward compatibility config:
  - `false` (default): `{:destroy, action}` 2-tuple only destroys the join record (old behavior)
  - `true` (set by installer for new projects): 2-tuple also destroys both join and destination
  - Will be removed in 4.0

Closes #859

### DIFF
--- a/documentation/topics/development/backwards-compatibility-config.md
+++ b/documentation/topics/development/backwards-compatibility-config.md
@@ -252,3 +252,23 @@ When a field is marked `sensitive?: true`, its value in validation error structs
 replaced with a redacted placeholder via `Ash.Helpers.redact/1`. This applies to both
 the non-atomic (`validate/3`) and atomic (`atomic/3`) code paths across all built-in
 validations.
+
+## many_to_many_destroy_destination_on_match?
+
+```elixir
+config :ash, many_to_many_destroy_destination_on_match?: true
+```
+
+### Old Behavior
+
+When using `on_match: {:destroy, :action_name}` (2-tuple form) with a `many_to_many`
+relationship, only the join record was destroyed using the given action. The
+destination record was left intact, making it effectively the same as `:unrelate`.
+
+### New Behavior
+
+The 2-tuple `{:destroy, :action_name}` for `many_to_many` now destroys both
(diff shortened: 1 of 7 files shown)


zachdaniel:
@nallwhy I don't believe that we should retain backwards compatibility on the `{:delete, :action}` format necessarily. If we want to keep it backwards compatible we can add a backwards compatibility configuration for this?

nallwhy:
Since we don't need to maintain backward compatibility for the `{:destroy, action}` format, I went ahead and simplified the implementation — the code is much cleaner this way.

zachdaniel:
Well, I think we may need to maintain backwards compatibility via a config of some kind, but I wouldn't want to have `{:destroy, :action}` destroy only the join, and `{:destroy, :action, :action}` destroy both for users moving forward. We can use a backwards compatibility configuration for this (documented in hex and you can also add it to the installer so it automatically uses the new nice behavior for new users). We'll strip that out in 4.0

zachdaniel:

PR #2645, opened 2026-03-23
Module.Callback Refactor

celeste-wahlquist:
# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [x] Refactoring
- [x] Update dependencies

### DIFF
--- a/.github/workflows/ash-ci.yml
+++ b/.github/workflows/ash-ci.yml
@@ -8,11 +8,19 @@ on:
     secrets:
       HEX_API_KEY:
         required: false
+      OBAN_PRO_LICENSE_KEY:
+        required: false
     inputs:
+      oban-pro:
+        type: boolean
+        default: false
       release:
         type: boolean
         default: true
       publish-docs:
+        type: boolean
+        default: true
+      unused-deps:
         type: boolean
         default: true
       spark-formatter:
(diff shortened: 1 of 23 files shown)


.github/workflows/ash-ci.yml
```diff
@@ -102,8 +102,8 @@ jobs:
   report_mix_deps:
     name: "Report Mix Dependencies"
     runs-on: ubuntu-latest
-    # Only run this job when we're on the main branch, not for PRs
-    if: github.ref == 'refs/heads/main'
+    # Only run this job for upstream main (forks often lack dependency submission permissions)
+    if: github.ref == 'refs/heads/main' && github.repository == 'ash-project/ash'
```
zachdaniel: We can't add this restriction because many different repositories use this same CI. We'd need to check maybe just the organization, i.e `ash-project`

test/ash_test.exs
```diff
@@ -416,6 +416,7 @@ defmodule Ash.Test.AshTest do
 
     def notify(notification) do
       send(self(), {:notification, notification})
+      :ok
```
zachdaniel, line 419: Given that we weren't requiring this before, I think we should just change the callback to `any` TBH. Because we've always ignored these results this could break a bunch of people's apps. Alternatively we can add a `warn?: true` option that will *warn* on an invalid return and then we can fix it in 4.0 to be more restriction.

.github/workflows/ash-ci.yml
celeste-wahlquist: Okay, undid the new 105-106 and left as original

test/ash_test.exs
celeste-wahlquist: changed the callback contract/wrapper to accept any() and removed strict runtime return-shape enforcement there

  @callback notify(Ash.Notifier.Notification.t()) :: any()
  @spec notify(module(), Ash.Notifier.Notification.t()) :: any()

.github/workflows/test-subprojects.yml
```diff
@@ -14,6 +14,7 @@ jobs:
   test-subprojects:
     runs-on: ubuntu-latest
     name: Subproject ${{matrix.project.org}}/${{matrix.project.name}} - OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}}
+    if: ${{ github.repository == 'ash-project/ash' || matrix.project.name != 'ash_oban' }}
```
zachdaniel, line 17: Why not run this for `ash_oban`?

celeste-wahlquist: Thanks for catching this! This isn’t “disable fork CI” wholesale — it only skips the ash_oban subproject matrix job on forks. On ash-project/ash, we still run it. The motivation is that ash_oban’s mix deps.get can pull oban_pro / private Hex packages that forks may not have configured, so the matrix cell would fail for reasons not related to the individuals code. Happy to revisit if we want a different approach 👍

zachdaniel: Ah, okay. Makes sense.

zachdaniel:
Looks like some conflicts (sorry) and then we can get it merged! Great change 🔥

zachdaniel:

Issue #1625, opened 2024-11-27
Support pipelines, to declare subsections of action logic that can be referenced from many other places


zachdaniel:
For example:

```elixir
update :open do
  pipe_through [:change_state]
  change set_attribute(:state, :open)
end

update :close do
  pipe_through [:change_state], where: [expr(^actor(:role) == :super_user)]
  change set_attribute(:state, :open)
end

pipelines do
  pipeline :change_state do
    validate expr(condition), error_expr: expr(error(...))
    change atomic_update(:score, expr(score + 1))
    change switch(
      [{expr(condition1), change}]
      [{expr(condition1), change}]
      [{expr(condition1), change}]
    )
    change change2
    change change3
  end
end
```

This has many potential downstream use cases.

PR #2646, opened 2026-03-24
feat: add `extends` option for actions

### LINKED ISSUE #966: (Maybe) add `extend` action type
The idea is that sometimes you have small variations on the same action that you want with slightly different modifiers.

In my case I wanted to expose the generated `sign_in_with_password` read action generated by `AshAuthentication` but change the token type (from `:sign_in` to `:user`) via GraphQL.  What I had to do was copy the entire action and add a `set_context/1` preparation.  What I wanted to do was this:

```
extend :sign_in_with_password do
  as :sign_in_with_password_for_graphql
  prepare set_context(%{token_type: :user, strategy_name: :password})
end
```

Behind the scenes I expect that the transformer will load all the entities from the other action, and then prepend them to all the entities in the extend action.

This could be a bad idea. I am sure Zach will tell me why.

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

- Add `extends` option to all action types (create, read, update, destroy, generic action)
- When an action extends another, it inherits all configuration from the base action
- List fields (arguments, changes, preparations, metadata, etc.) are concatenated: base first, then extending action
- Scalar fields are inherited from base unless explicitly overridden
- `primary?` is never inherited to avoid conflicts

Closes #966

### Example

```elixir
actions do
  read :sign_in_with_password do
    prepare SomePreparation
    argument :email, :string
  end

  read :sign_in_for_graphql do
    extends :sign_in_with_password
    prepare set_context(%{token_type: :user})
  end
end
```


zachdaniel:
Hmm...I don't think I'd like to take this approach. I will close the original issue. I think we should do this instead: https://github.com/ash-project/ash/issues/1625

PR #2647, opened 2026-03-24
fix: recompute changed? after setup_managed_belongs_to_relationships

barnabasJ:
## Summary

- When `manage_relationship` is called from `before_transaction` or `before_action` hooks on update actions, the `changed?` flag was computed **before** `setup_managed_belongs_to_relationships` runs, so it remained `false` even after the FK was set via `force_change_attribute`
- This caused the `cond` branch to skip the DB update entirely — the related record was created but the source record's FK was never persisted
- Fix: recompute `changed?` after `setup_managed_belongs_to_relationships` returns, so FK changes from hook-added managed relationships are detected

**Companion PR:** ash-project/ash_postgres — regression tests (will link once created)

## Test plan

- [x] Added test in `test/manage_relationship_test.exs` for update + `before_transaction` hook with ETS data layer
- [x] All 33 manage_relationship tests pass
- [x] Full ash_postgres suite passes (713 tests) with this fix

### DIFF
--- a/lib/ash/actions/update/update.ex
+++ b/lib/ash/actions/update/update.ex
@@ -471,8 +471,22 @@ defmodule Ash.Actions.Update do
                 {:error, error}
 
               {changeset, manage_instructions} ->
+                # Recompute changed? after setup_managed_belongs_to_relationships
+                # may have set FK attributes via force_change_attribute (e.g. when
+                # manage_relationship is called from before_transaction hooks on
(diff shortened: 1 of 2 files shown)


barnabasJ:
Companion regression tests: ash-project/ash_postgres#727

zachdaniel (approved):

zachdaniel:

PR #2649, opened 2026-03-25
fix: handle nil return from generic actions with load option

barnabasJ:
## Summary
- Fix `BadMapError` crash when a generic action returns `nil` and `load:` option is specified
- Enforce the existing `allow_nil?` DSL option (defaults to `false`) which was previously never checked at runtime
- Extract load logic into `maybe_load/4` and validation into `validate_allow_nil/3` for clarity

## Context
When a generic action returns `nil` (e.g., a lookup action that finds no record) and the caller passes `load: [:some_relationship]`, `Ash.Actions.Helpers.load/4` would call `Map.get(nil, :id)` via `Ash.Resource.selected?/3`, causing a `BadMapError`.

The `allow_nil?` option existed in the generic action DSL (defaulting to `false`) but was never enforced. Now:
- `allow_nil? true` → `nil` is returned successfully, loads are skipped
- `allow_nil? false` (default) → returns a clear `InvalidReturnType` error instead of crashing

## Test plan
- [x] Test generic action returning nil with `load:` and `allow_nil? true` succeeds
- [x] Test generic action returning nil without `load:` and `allow_nil? true` succeeds
- [x] Test generic action returning nil with `allow_nil? false` (default) returns proper error
- [x] All existing generic action tests pass (62 tests, 0 failures)

### DIFF
--- a/lib/ash/actions/action.ex
+++ b/lib/ash/actions/action.ex
@@ -79,18 +79,7 @@ defmodule Ash.Actions.Action do
             end
           end)
 
-        result =
-          with {:ok, result, _} <-
-                 Ash.Actions.Helpers.load(result, input, domain,
-                   actor: opts[:actor],
-                   reuse_values?: true,
-                   action:
-                     Ash.Resource.Info.primary_action(input.resource, :read) || input.action,
-                   authorize?: opts[:authorize?],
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2659, opened 2026-04-03
improvement: add `%Ash.Expr{}` wrapper struct

### LINKED ISSUE #1069: Add `%Ash.Expr{}`, which will be returned from `expr/1`
Its not possible to easily tell “did they pass an expression”. You can use `Ash.Expr.expr?` but that literally traverses the entire value.

I *think* we can actually change this transparently to end users, since they shouldn’t be manipulating expressions, but we’ll need to adjust a bunch of code to just pass the contained expression by. This will also be useful in disambiguating interfaces.

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary

Closes #1069

- Introduces `%Ash.Expr{}` struct returned from `expr/1`, `where/2`, and `or_where/2` macros, enabling O(1) detection of expression values via `match?(%Ash.Expr{}, value)` instead of traversing the entire value with `expr?/1`
- Adds `Ash.Expr.expression()` type for function parameters that accept both wrapped and raw expressions, keeping `Ash.Expr.t()` for the struct itself

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -2607,6 +2607,10 @@ defmodule Ash.Changeset do
     end
   end
 
+  defp check_for_exists_with_relationships(%Ash.Expr{expr: inner}) do
+    check_for_exists_with_relationships(inner)
+  end
+
   defp check_for_exists_with_relationships(%Ash.Query.Exists{path: path}) when path != [] do
     {:error, "atomic_set cannot use exists with relationships"}
   end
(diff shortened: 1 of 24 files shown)


zachdaniel:
This is a very useful change but I don't think that we can reasonably modify all expression logic to return `%Ash.Expr{}` without making a major version bump. There are tons of scenarios that use expressions.

nallwhy:
The goal of this issue is O(1) `expr?/1` detection. For that to work, `expr/1` needs to return a distinguishable struct `%Ash.Expr{}`, which means any code pattern-matching on `expr/1` results (e.g. `%Ash.Query.Call{} = expr(...)`) would break.

Should this be scoped as a 4.0 change?

zachdaniel:
Yeah, I think we will need to do that unfortunately. What we can do is introduce this as an hidden option that we can use to introduce support across all our extensions, but other external parties will be relying on this structure so we can't break the shape of an expr without a major release.

We should probably start and maintain a 4.0 branch 😓

PR #2664, opened 2026-04-07
fix: dump attribute values to native in atomic upgrade path

sevenseacat:
The `assume_casted?` path in `atomic_params/4` was putting attribute values directly into `changeset.atomics` without calling `dump_to_native/3`.

For types where the cast representation differs from the native representation, this meant improperly formatted values reached the data layer.

This was leading to errors in our tests after upgrading to Ash 3.23.1, such as:

```
 {:error,
  %Ash.Error.Unknown{
    bread_crumbs: ["Returned from bulk query update: MyApp.FormBuilder.ClientFormResponse.complete"],
    changeset: "#Changeset<>",
    errors: [
      %Ash.Error.Unknown.UnknownError{
        error: "** (DBConnection.EncodeError) Postgrex expected a binary of 16 bytes, got \"019d66bc-d770-71ea-9eb7-3086dd6b77eb\". Please make sure the value you are passing matches the definition in your table or in your query or convert the value accordingly.",
        field: nil,
        value: nil,
        splode: Ash.Error,
        bread_crumbs: ["Returned from bulk query update: MyApp.FormBuilder.ClientFormResponse.complete"],
        vars: [],
        path: [],
        class: :unknown
      }
    ]
  }}
```

## Notes

While UUID is the most visible case (hard Postgrex error), any type with a non-trivial `dump_to_native` could be silently affected on this code path. It may be worth auditing other types — though most built-in types have identity or near-identity dump functions.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -1752,7 +1752,13 @@ defmodule Ash.Changeset do
             {:cont, %{changeset | arguments: Map.put(changeset.arguments, key, value)}}
 
           attribute = Ash.Resource.Info.attribute(changeset.resource, key) ->
-            {:cont, atomic_update(changeset, attribute.name, {:atomic, value})}
+            case Ash.Type.dump_to_native(attribute.type, value, attribute.constraints) do
+              {:ok, dumped} ->
+                {:cont, atomic_update(changeset, attribute.name, {:atomic, dumped})}
+
+              :error ->
+                {:cont, add_invalid_errors(value, :attribute, changeset, attribute, "is invalid")}
+            end
 
           match?("_" <> _, key) ->
             {:cont, changeset}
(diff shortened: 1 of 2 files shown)


sevenseacat:
Though I'm seeing a very similar error for Money types not being cast correctly - investigating

```
%Ash.Error.Unknown.UnknownError{
  error: "** (DBConnection.EncodeError) cannot encode anonymous tuple {\"USD\", 
Decimal.new(\"600.00\")}. Please define a custom Postgrex extension that matches on 
its underlying type:\n\n    use Postgrex.BinaryExtension, type: \"typeinthedb\"\n",

sevenseacat:
My version of ash_sql was out of date, and updating from 0.5.1 to 0.5.5 fixed it!

PR #2666, opened 2026-04-08
feat: add Ash.Query.where/2 as non-macro alternative to filter/2

### LINKED ISSUE #1092: Non-macro alternative to `Ash.Query.filter`
All `Ash.Query` modifications - `select`, `sort`, `limit`, `offset`, `build` and so on - are usual methods. Except one, the most used - `filter` - it is a macro. So it needs a require statement to be used. Which is a little bit annoying when you want to add a simple filter and only once in a file... I just go for `build(filter:)` combo in that case.

Would be convenient to have non-macro variant of `filter` exposed (which is `do_filter`). Maybe under name `where`?

In small discussion on Discord the following proposal was made: add `where` in Ash 3.1 (because there are too many things already in Ash 3 and it is too late to change `filter` to non-macro method even though it is desirable) and make it a primary way to add filters to queries in docs, then in Ash 4 deprecate `filter`.

nallwhy:
## Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

## Summary
- Add `Ash.Query.where/2` as a non-macro alternative to `Ash.Query.filter/2`
- Delegates to the existing `do_filter/3` function

Closes #1092

### DIFF
--- a/lib/ash/query/query.ex
+++ b/lib/ash/query/query.ex
@@ -695,6 +695,37 @@ defmodule Ash.Query do
   end
 
   @doc """
+  Attach a filter statement to the query.
+
+  This is a non-macro alternative to `filter/2` that doesn't require
+  `require Ash.Query`. It accepts keyword/map style filters but not
+  expression syntax (use `filter/2` for expressions like `filter(title == "foo")`).
+
+  Passing `nil` is a no-op, which is useful for conditional filtering in pipelines.
+
+  The filter is applied as an "and" to any filters currently on the query.
+
+  ## Examples
+
+      # Keyword syntax
+      MyApp.Post
+      |> Ash.Query.where(published: true)
+
+      # Map syntax
+      MyApp.Post
+      |> Ash.Query.where(%{points: %{greater_than: 100}})
+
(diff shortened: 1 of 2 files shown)


lib/ash/query/query.ex
```diff
@@ -694,6 +694,37 @@ defmodule Ash.Query do
...
+  @doc """
+  Attach a filter statement to the query.
+
+  This is a non-macro alternative to `filter/2` that doesn't require
+  `require Ash.Query`. It accepts keyword/map style filters but not
+  expression syntax (use `filter/2` for expressions like `filter(title == "foo")`).
+
+  Passing `nil` is a no-op, which is useful for conditional filtering in pipelines.
+
+  The filter is applied as an "and" to any filters currently on the query.
```
zachdaniel, line 708: So the concept here isn't that we *only* support the data structure format, thats just how it works by default. However you can do `import Ash.Expr` at which point you can do `Ash.Query.where(expr(...))`. That gives us one single function to use that does both expressions and non-expressions 😄

PR #2676, opened 2026-04-14
fixes for embedded resources within unions

jsw800:
These changes fix a few issues that I ran into with unions of embedded resources.

What I'm trying to do is:

- a union
- with constraint `storage: :map_with_tag`
- The individual types are embedded resources
- Some of these embedded resources' attributes include a `source` field that remaps them from what's in the DB (the DB versions are camel case because of bad past decisions)
- There are field policies on some of the embedded resource fields since they should not be visible to everyone

I've run into a few issues doing this. I worked with opencode to get to some solutions here. I'm struggling to put together good test reproductions since this is kind of a niche use case and I've mostly been testing it against my own codebase.

The issues themselves are:

- loads of embedded resources inside unions were being bypassed by the `def load(unions, [], _, _), do: {:ok, unions}` function head
  - the fix for this was to remove that function head and simply default the `load` variable to `[]` in the main function head
- the `__union_tag__` property was not being properly included in a few code paths that handle updates, resulting in errors about `No such input type for action <MyAction>` (`type` is my tag key)
  - the fix was to include `__union_tag__` in the constraints in those code paths
- field policies were not being properly applied to the fields in the embedded resources
  - the fix was the changes in read.ex to the `cleanup_field_auth` function. To be transparent, I understand these changes less well than I do the other two—opencode is mostly responsible for these. I would appreciate a bit of a closer look at this section from someone with a much closer understanding of this module's code.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -1357,9 +1357,23 @@ defmodule Ash.Actions.Read do
         end)
       end)
 
+    # Ensure loadable attributes are included in load_through for cleanup,
+    # even if they weren't explicitly in the query's load_through.
+    # This matches what load_through_attributes does dynamically.
+    attribute_load_through =
+      resource
+      |> Ash.Resource.Info.attributes()
+      |> Enum.filter(fn %{name: name, type: type, constraints: constraints} ->
+        Ash.Type.can_load?(type, constraints) && (is_nil(query.select) || name in query.select)
+      end)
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:
Seems reasonable to me 😄 

PR #2682, opened 2026-04-21
fix: Handle calculations on data layers that disallow filters

Munksgaard:
Previously, adding calculations using `is_nil` on resources in data layers that
didn't support filtering, such as the one in the included test, the following
error was returned (at compile time if using `:auto` as the type, at runtime if
using `:boolean` as the type):

    key :arguments not found in: is_nil(bookedInvoiceId)

Fix: remove the predicate exclusion from the hydrate_refs branch in
resolve_call. do_hydrate_refs already handles predicates correctly.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -3886,7 +3886,7 @@ defmodule Ash.Filter do
              {:func, function_module} when not is_nil(function_module) <-
                {:func, get_function(name, context.resource, context.public?)},
              {:ok, function} <- Function.new(function_module, args) do
-          if Ash.Expr.expr?(function) && !match?(%{__predicate__?: _}, function) do
+          if Ash.Expr.expr?(function) and not is_struct(function, function_module) do
             hydrate_refs(function, context)
           else
             if can_filter_expr?(context, function) do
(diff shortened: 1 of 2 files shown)


lib/ash/filter/filter.ex
```diff
@@ -3886,7 +3886,7 @@ defmodule Ash.Filter do
              {:func, function_module} when not is_nil(function_module) <-
                {:func, get_function(name, context.resource, context.public?)},
              {:ok, function} <- Function.new(function_module, args) do
-          if Ash.Expr.expr?(function) && !match?(%{__predicate__?: _}, function) do
+          if Ash.Expr.expr?(function) do
```
zachdaniel: Need to think more deeply about this one. The concept here is "if function.new returns an expression other than the function itself, then we need to send that new value through ref hydration" IIRC. So what we might want to do is instead do something like:

```elixir
if Ash.Expr.expr?(function) and not is_struct(function, function_module) do
```

Could you try that out?

Munksgaard: That also works for me when testing locally. I've amended the PR to follow your suggestion :heart:

zachdaniel:

PR #2686, opened 2026-04-28
feat: add manual keyset pagination delegated to data layer

### LINKED ISSUE #2684: Proposal: Add keyset pagination logic at the data layer
### Code of Conduct

### AI Policy

I'm currently building a data layer for atproto repos. The querying logic would work by wrapping a XRPC (HTTP) endpoint and the returned data as an Ash resource. The problem i ran into is that the response contains its own cursor, while the current implementation of keysets in Ash is completely managed by the framework. So, currently there's no native way to handle this cursor. Attempts to include a keyset struct manually result in an error when Ash tries to decode it.

An option to delegate keyset handling to the data layer.

I think this decision should be abstracted away from the data layer consumer, so maybe a callback check could decide which layer will handle it? for example, if the `manual_keyset?(resource)` callback returns true, then `run_paginated_query/2` will be used instead of `run_query/2`, which would expect a tuple with a keyset struct generated by the data layer itself, then include that when returning the result of the action.

I also thought of using `can?/2`, but i thought that something that changes the query behavior should be more explicit. And i considered changing the return type of `run_query/2` to include the keyset in the tuple.

1. **Handling pagination using context** — This works, but it's somewhat unintuitive compared to other data layer implementations. It would be another pagination system implemented alongside the native Ash one, which would require the consumer to pass data to the context directly rather than using the pagination API Ash provides, so it's not ideal.

### Additional context

- [atproto API spec](https://atproto.blue/en/latest/atproto/atproto_client.models.com.atproto.repo.list_records.html)

Rekkice:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

## Summary
- Add new `can?` feature to data layer (`:keyset`), which is checked to decide which layer will handle keyset pagination
- Add conditions to avoid keyset pagination in the Ash core layer in case the data layer is handling it
- Change `more?` logic to rely on keyset passed by data layer if delegated
- Pass page options directly to data layer through context if delegated

Closes #2684

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -2986,7 +2986,13 @@ defmodule Ash.Actions.Read do
         data
       end
 
-    more? = not Enum.empty?(rest)
+    more? =
+      if use_data_layer_keyset?(original_query, action.pagination) do
+        last_record = List.last(data)
+        not is_nil(last_record) && not is_nil(last_record.__metadata__[:keyset])
+      else
+        not Enum.empty?(rest)
+      end
 
     if page_opts[:offset] do
       Ash.Page.Offset.new(data, count, original_query, more?, opts)
@@ -3014,10 +3020,15 @@ defmodule Ash.Actions.Read do
   end
 
   defp add_keysets(original_query, data, sort) do
-    if Enum.any?(
-         Ash.Resource.Info.actions(original_query.resource),
-         &(&1.type == :read && &1.pagination && &1.pagination.keyset?)
-       ) do
+    action =
+      Enum.find(
+        Ash.Resource.Info.actions(original_query.resource),
+        &(&1.type == :read && &1.pagination && &1.pagination.keyset?)
+      )
+
+    pagination = action && action.pagination
+
+    if pagination && !use_data_layer_keyset?(original_query, pagination) do
       Ash.Page.Keyset.data_with_keyset(data, original_query.resource, sort)
     else
       data
@@ -3045,6 +3056,15 @@ defmodule Ash.Actions.Read do
          _missing_pkeys?
        ) do
     data
+  end
+
+  defp use_data_layer_keyset?(query, pagination) do
+    cond do
+      !Ash.DataLayer.data_layer_can?(query.resource, :keyset) -> false
+      pagination.via_data_layer? != :data_layer_default -> pagination.via_data_layer?
+      Ash.Resource.Info.data_layer(query.resource).data_layer_keyset_by_default?() == true -> true
+      true -> false
+    end
   end
 
   defp attach_fields(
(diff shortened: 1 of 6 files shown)


Rekkice:
Opening this as a draft to get a check on the approach. I'd gladly add unit tests if it looks fine

lib/ash/actions/read/read.ex
```diff
@@ -2983,7 +2983,13 @@ defmodule Ash.Actions.Read do
         data
       end
 
-    more? = not Enum.empty?(rest)
+    more? =
+      if Ash.DataLayer.data_layer_can?(original_query.resource, :keyset) do
+        last_record = List.last(data)
+        not is_nil(last_record) && not is_nil(last_record.__metadata__[:keyset])
```
zachdaniel, line 2992: So the concept here is that the data layer would only add a final keyset if there is (or might be?) a next page?

zachdaniel:
Generally this makes sense to me, would be good to understand an example data layer e2e use case so we can document how its used etc.

zachdaniel:
We may also want to go beyond just checking if the data layer supports keyset, and make this explicitly opt-in via pagination options, i.e `pagination ..., via_data_layer?: true` or something like that.

lib/ash/actions/read/read.ex
Rekkice: Yes, generally APIs include a cursor if there's more records to fetch. I noticed a quirk in the atproto API where it returns a cursor even if there's no more pages left, so maybe it should check if it matches the limit too?

Rekkice:
> Generally this makes sense to me, would be good to understand an example data layer e2e use case so we can document how its used etc.

My data layer implementation looks like this right now:

```elixir
def can?(_, :keyset), do: true

def run_query(%Query{} = query, resource, _parent) do
  target_repo_did = query.context[:atproto_repo_did] || get_did_from_actor(query)

  if is_nil(target_repo_did) do
    {:error,
     Error.Query.InvalidQuery.exception(
       message: "The context must include the `:atproto_repo_did` to read from atproto."
     )}
  else
    pds_url = get_pds_endpoint!(target_repo_did)
    record_type = AshAtproto.DataLayer.Info.atproto_repo_record_type!(resource)

    page_opts = query.context[:data_layer][:keyset_opts] || []
    cursor = page_opts[:after]

    params = [
      repo: target_repo_did,
      collection: record_type,
      limit: min(query.limit || @default_limit, @max_limit)
    ]

    params = if cursor, do: Keyword.put(params, :cursor, cursor), else: params

    params =
      if query.context[:atproto_reverse], do: Keyword.put(params, :reverse, true), else: params

    case Atex.XRPC.unauthed_get(pds_url, "com.atproto.repo.listRecords", params: params) do
      {:ok, %{status: 200, body: body}} ->
        valid_attributes = Ash.Resource.Info.attributes(resource)

        records =
          body["records"]
          |> Enum.map(fn item ->
            map_to_record(item, resource, valid_attributes, body["cursor"])
          end)

        {:ok, records}

      {:error, reason} ->
        {:error, Ash.Error.to_ash_error(reason)}
    end
  end
end

defp map_to_record(map, resource, valid_attributes, cursor \\ nil) do
  raw_json = map["value"] || %{}

  {:ok, %{authority: target_repo_did, rkey: rkey}} = get_uri_data(map["uri"])

  base_attrs = %{
    repo_did: target_repo_did,
    rkey: rkey,
    cid: map["cid"]
  }

  safe_attrs =
    Enum.reduce(valid_attributes, base_attrs, fn attribute, acc ->
      json_key = to_string(attribute.source || attribute.name)

      if Map.has_key?(raw_json, json_key) do
        case Ash.Type.cast_stored(
               attribute.type,
               raw_json[json_key],
               attribute.constraints
             ) do
          {:ok, value} -> Map.put(acc, attribute.name, value)
          :error -> acc
        end
      else
        acc
      end
    end)

  Kernel.struct(resource, safe_attrs)
  |> Ash.Resource.put_metadata(:keyset, cursor)
end
```

> We may also want to go beyond just checking if the data layer supports keyset, and make this explicitly opt-in via pagination options, i.e `pagination ..., via_data_layer?: true` or something like that.

Having an option to override the manual keyset pagination would be nice, i'll try to add that too.

But i think making it opt-in would make it a leaky abstraction. The developer would have to know how the data layer handles this specific thing, and if they choose the wrong option keyset pagination would break completely in my case. So i think giving the data layer the choice to have it enabled by default would be a good idea for this.

zachdaniel:
Hmm...I see what you mean about making it a leaky abstraction, but I do also think it kind of leaks no matter how you think about it because depending on what you're doing or the performance characteristics of what you're doing you may or may not want to do pagination app-side etc.

I do see what you're saying though. What I'm imagining is a scenario though where we make the built in ash_postgres data layer support this with a database-level cursor or something and then that affects like every single Ash application in production 😆. We could *maybe* let the data layer decide if its implementation of keysets should be the default behavior or not? Then the way you override it depends on the data layer?

Rekkice:
Yeah, i agree with that idea. That's close to what i was thinking about. The `can?` condition could be replaced by this:

```elixir
use_data_layer_keyset? = 
  cond do
    !Ash.DataLayer.data_layer_can?(query.resource, :keyset) -> false
    Keyword.has_key?(opts, :via_data_layer?) -> opts[:via_data_layer?] 
    # maybe there could be an extra feature that decides whether it's a default or not, aside from the new `:keyset` one
    Ash.DataLayer.data_layer_can?(query.resource, :data_layer_keyset_by_default) -> true
    true -> false
  end
```

zachdaniel:
For this one:

` Ash.DataLayer.data_layer_can?(query.resource, :data_layer_keyset_by_default) -> true`
We should just make it a callback on the data layer, i.e something like:

```elixir
Ash.Resource.Info.data_layer(query.resource).data_layer_keyset_by_default?() == true
```

Rekkice:
I included the option for the pagination block and the new conditions in the new commit. I ran the existing tests and tested the new cases with my data layer and it seems to be working fine, so i'll start writing the tests now

Rekkice:
I just updated the commit to include tests

lib/ash/actions/read/read.ex
zachdaniel: It sounds like its on the data layer to tell us if there are more items or not via adding a keyset to the last item, so its probably fine as is here.

zachdaniel:
Looks like some more failing tests. You may want to pull main in to fix `audit` and I think the rest are in this PR

zachdaniel:
You can run `mix check` locally to check all of these BTW

Rekkice:
I forgot to sync it with upstream first, it should be working now!

PR #2692, opened 2026-05-11
fix: preserve all errors in Required dedupe accumulator

sevenseacat:
`remove_required_if_other_errors_exist/1` is meant to keep one Required error per `{field, path}` group when duplicates exist. Instead, its `Enum.reduce` `else` branch replaced the entire accumulator each iteration.

With N duplicated groups, only the last group's representative survived; non-Required errors seeded into the accumulator were dropped too.

## How this surfaced

When building an AshPhoenix form that has an embedded resource (eg. `ServiceLocation` has an `Address` attribute), I was only seeing a single validation error for the last invalid attribute of the embedded address. eg. the address was missing a required `line1`, `state` AND `zip_code` but I was only seeing the validation error about the missing zip code.

And if there were other validations (eg. validating the zip code format), those errors were never surfaced while there was a Required validation reported (the second test).

With Claude's assistance I narrowed the issue down to this fix.

## If its from AshPhoenix, why not fix it there?

The main reason this code path gets triggered is because AshPhoenix validates the embedded attribute twice - once as part of the parent form, and once as part of building a new nested form for the attribute. But it can also be triggered in different ways - we saw the same issue last week with an action that used `manage_relationship` with `on_no_match: {:create, :custom_create_action}`.

Plus it just seems like a fairly obvious bug in the function - `Enum.reduce` replacing instead of appending to the accumulator.

# Contributor checklist

- [x] I accept the [AI Policy](https://github.com/ashproject/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/error/error.ex
+++ b/lib/ash/error/error.ex
@@ -157,7 +157,7 @@ defmodule Ash.Error do
            end) do
           errors
         else
-          [Enum.max_by(required_errors, &Enum.count(&1.bread_crumbs))]
+          [Enum.max_by(required_errors, &Enum.count(&1.bread_crumbs)) | errors]
         end
       end)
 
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:
🤦 

PR #2694, opened 2026-05-11
improvement: generate `@type t` typespec for `TypedStruct` modules

Munksgaard:
Previously, a `TypedStruct` didn't actually specify any _types_. This PR adds a `@spec` definition for typed structs, while still respecting userdefined specs.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/typed_struct.ex
+++ b/lib/ash/typed_struct.ex
@@ -193,13 +193,16 @@ defmodule Ash.TypedStruct do
             instance_of: module
           ]
 
+        typespec_ast = Ash.TypedStruct.build_typespec_ast(fields)
+
         # sobelow_skip ["RCE.CodeModule"]
         Code.eval_quoted(
           Ash.TypedStruct.__define_typed_struct__(
             enforce_keys,
             fields_with_defaults,
             map_constraints,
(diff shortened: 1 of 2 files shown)


Munksgaard:
While I can certainly fix the formatter errors, I don't think they're caused by this PR.

zachdaniel:
Thus far we have avoided adding any generated typespec information into Ash. The thought process here is that we should focus on and wait for the new type system, at which point we will add macro callbacks for types to describe their input, output, and stored formats.

This for example, is only a bit of a half-feature considering that custom types etc. wouldn't be supported. I'd rather not add this for now.

Munksgaard:
> Thus far we have avoided adding any generated typespec information into Ash. The thought process here is that we should focus on and wait for the new type system, at which point we will add macro callbacks for types to describe their input, output, and stored formats.
> 
> This for example, is only a bit of a half-feature considering that custom types etc. wouldn't be supported. I'd rather not add this for now.

Perfectly reasonable, thanks for taking the time to explain :heart:

PR #2695, opened 2026-05-11
test: reproduce compile deadlock for self-referencing NewType in action argument

olivermt:
Updated with proposed fix as well

# Contributor checklist

- [x] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/map.ex
+++ b/lib/ash/type/map.ex
@@ -34,6 +34,14 @@ defmodule Ash.Type.Map do
             constraints: [
               type: :keyword_list,
               default: []
+            ],
+            init?: [
+              type: :boolean,
+              default: true,
+              doc: """
+              If false, the field's type constraints are not initialised at compile time. \
+              Allows for recursive map fields.
+              """
             ]
           ]
         ]
--- a/lib/ash/type/new_type.ex
+++ b/lib/ash/type/new_type.ex
@@ -109,6 +109,14 @@ defmodule Ash.Type.NewType do
     end
   end
 
+  defp expand_module_refs(term, env) do
+    Macro.prewalk(term, fn
+      {:__aliases__, _, _} = ast -> Macro.expand(ast, env)
+      {:__MODULE__, _, _} -> env.module
+      other -> other
+    end)
+  end
+
   defmacro __using__(opts) do
     case Keyword.keys(opts) -- [:subtype_of, :constraints, :lazy_init?] do
       [] ->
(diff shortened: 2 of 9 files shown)


lib/ash/type/map.ex
```diff
@@ -97,7 +106,7 @@ defmodule Ash.Type.Map do
         type = Ash.Type.get_type(config[:type])
         constraints = config[:constraints] || []
 
-        if Keyword.get(config, :init?, true) do
+        if Keyword.get(config, :init?, true) and not Keyword.get(config, :lazy_init?, false) do
```
zachdaniel: 🤔 it looks like we had an undocumented flag for doing this already, which is `init?`. Perhaps we should just reuse that and document it?

olivermt: good call! force pushed a flip to using it now

lib/ash/type/new_type.ex
```diff
@@ -109,6 +133,78 @@ defmodule Ash.Type.NewType do
     end
   end
 
+  @map_subtypes [:map, Ash.Type.Map]
+
+  defp validate_field_types!(constraints, subtype_of, env) when subtype_of in @map_subtypes do
+    walk_fields(constraints, env)
```
zachdaniel: I still don't think this is quite right. First off, there are more types than just those that have sub-types (map, keyword, tuple, struct, union). I don't like the idea that `NewType` has to know about all of those types, and there could always be custom types that don't get validated. An idea:

At compile time *within the type itself*, call its own `init` callback with empty constraints (like a before compile hook). This would force the compile cycle to become apparent if any other type tries to reference this type. Then if you had a type like `Ash.Type.Union` that through some convoluted path referenced this one, we'd find out at compile time and at that point can provide a helpful error.

olivermt: That sort of works, but I dont have a better way to show error to the user other than that the WhateverModuleName is not reachable, but combined with some help text that this is probably from a recursive definition without the right settings that is probably all the DX we need.

olivermt: However me and claude are driving a bit blind here, ast macroing is not my strongest suit, but I managed to figure out what needed to be pulled out of the `qoute` blocks, but it meant a fairly substantial rewrite of the newtype file. 

I did say on discord that this might be a bit out of my league to fix, but at least I has force pushed up what I at least think is most of a solution in the direction you said, however I am not sure about the implementation changes. 

Is there a trick to do the compile into argument error rescue without doing the full thing and then doing the check at the end like this code is doing?

olivermt: Also realized the api was missing on a lot of the other types, added new commit for it now

lib/ash/type/new_type.ex
```diff
@@ -109,232 +133,193 @@ defmodule Ash.Type.NewType do
     end
   end
 
-  defmacro __using__(opts) do
```
zachdaniel, line 112: So this change can actually go directly into `Ash.Type` as opposed to `Ash.Type.NewType`, and should produce warnings on failure. The rationale is that all types must be able to be initialized w/ constraints `[]` (and then thrown away) once during compilation to defend against compile cycles.

olivermt: something like this?

olivermt:
simplied walking logic version here https://github.com/ash-project/ash/pull/2724

olivermt:
closing due to zach zaching it up better on main

Issue #2580, opened 2026-02-22
Manual Relationship Should select all


allenwyma:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Performing this action inside of a change for an update throws an error:
```elixir
change fn changeset, _context ->
      Ash.Changeset.before_action(changeset, fn changeset ->
        case Ash.get(
               App.Services.Freshbooks.Service,
               Ash.Changeset.get_attribute(changeset, :freshbooks_id),
               load: [:rate],
               authorize?: false
             ) do
          {:error, _term} ->
            Ash.Changeset.add_error(changeset, "cannot get the service from Freshbooks")
...
```
error:

```
 * ** (UndefinedFunctionError) function Ash.Resource.ManualRelationship.Function.select/1 is undefined or private
       (ash 3.17.0) lib/ash/actions/read/read.ex:88: Ash.Actions.Read.run/3
       (ash 3.17.0) lib/ash.ex:2128: Ash.do_get/6
       (ash 3.17.0) lib/ash.ex:2103: Ash.get/3
```

Detailed info: https://discord.com/channels/711271361523351632/1475051562823979082

### Operating system

macos

CaydenLords:
Since the pull request was accepted, we should probably close this issue.

allenwyma:
yeah, this one is solved, sorry

PR #2700, opened 2026-05-15
fix: Manual relationships select missing

ken-kost:
#2580 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/manual_relationship/function.ex
+++ b/lib/ash/resource/manual_relationship/function.ex
@@ -5,6 +5,8 @@
 defmodule Ash.Resource.ManualRelationship.Function do
   @moduledoc false
 
+  def select(_opts), do: []
+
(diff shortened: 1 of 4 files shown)


lib/ash/resource/manual_relationship/function.ex
```diff
@@ -5,6 +5,8 @@
 defmodule Ash.Resource.ManualRelationship.Function do
   @moduledoc false
 
+  def select(_opts), do: []
```
zachdaniel: This probably isn't the right default. We may need to update `select/1` to be able to return `:*` to imply everything, as that would be a safer default.

zachdaniel (approved):

zachdaniel:

PR #2702, opened 2026-05-16
Fix type casting

Torkan:
Round-trips embedded data through types whose embedded format differs from their native format.

Before: an embedded resource attribute typed as :binary (or any type where dump_to_embedded ≠ dump_to_native) wouldn't survive a save/load cycle — it was dumped with one encoding and read back with another.

Adds a cast_from_embedded/2 callback (with array variant) as the reverse of dump_to_embedded/2, implements it on :binary and the composite types (map, struct, tuple, keyword, union), and routes embedded-resource reads through it. Defaults to cast_stored/2, so custom types keep working unchanged.

# Contributor checklist


zachdaniel:

PR #2704, opened 2026-05-17
Feat/add manifest filtering and sorting

Torkan:
# Summary

Extends `Ash.Info.Manifest` with the filter/sort vocabulary needed by code-gen consumers (e.g. ash_graphql, ash_typescript) so they no longer have to re-derive it from raw Ash internals.

**Top-level catalog** (`%FilterCapabilities{}`, `%SortCapabilities{}`):
- Every builtin operator, function, custom expression, and data-layer-contributed function, each with normalized argument signatures and predicate flag
- Sort directions

**Per-field resolution** on every `%Field{}`:
- `filter_operators`, `filter_functions`, `filter_custom_expressions` — lists of `%Applicable*{}` records pairing each applicable name with its resolved rhs type (`:same`, `:any`, `{:concrete, t}`, `{:array, t}`)
- NewType subtype fallthrough; ordered-comparison operators (`<`, `>`, `<=`, `>=`) trimmed for arrays and booleans

**Per-relationship**: `filterable?`/`sortable?` booleans replaced with `filterable_via` (`:traversal` | `:exists` | `false`) and `sortable_via` (`:traversal` | `false`) tagged atoms reflecting cardinality — one-cardinality relationships can be dot-traversed, many-cardinality need an existential quantifier. Fields keep their booleans (no cardinality distinction applies).

**Aggregate fix**: `%Field{kind: :aggregate}.allow_nil?` now derives from aggregate kind (`:count`/`:exists`/`:list` → false; `:first`/`:max`/`:min`/`:sum`/`:avg` → true) instead of the unrelated `:include_nil?` input flag.

# Contributor checklist

### DIFF
--- a/lib/ash/info/manifest/argument_signature.ex
+++ b/lib/ash/info/manifest/argument_signature.ex
@@ -1,0 +1,92 @@
+# SPDX-FileCopyrightText: 2025 ash contributors <https://github.com/ash-project/ash/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Ash.Info.Manifest.ArgumentSignature do
+  @moduledoc """
+  Normalized argument signature for an operator, function, or custom expression.
+
+  One signature represents one accepted argument shape. The `args` list is the
+  ordered list of arg specs; each arg spec has a `kind`:
+
+    * `:concrete` — a specific type (`builtin` atom or `type_ref` module).
+    * `:same` — same type as the field being filtered (or the first arg).
+    * `:any` — any type accepted.
+    * `:array` — an array whose inner element spec is in `of`.
+    * `:ref` — fallback for shapes we don't structurally model yet.
+
+  Use `from_ash_signature/1` to normalize the raw shape returned by an Ash
+  operator's `types/0` or a function's `args/0` callback.
+  """
+
+  @type arg_spec :: %{
+          required(:kind) => :concrete | :same | :any | :array | :ref,
+          required(:builtin) => atom() | nil,
+          required(:type_ref) => module() | nil,
+          required(:constraints) => keyword(),
+          optional(:of) => arg_spec()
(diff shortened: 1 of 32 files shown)


lib/ash/info/manifest/argument_signature.ex
```diff
@@ -0,0 +1,92 @@
...
+
+    case Keyword.fetch(opts, :of) do
+      {:ok, of} -> Map.put(base, :of, of)
+      :error -> base
+    end
+  end
+
+  # Heuristic: lowercase, non-Elixir atom is a builtin (`:string`, `:integer`, ...).
+  # Modules start with uppercase (`Ash.Type.Integer`).
+  defp builtin_type?(atom) do
```
zachdaniel, line 86: Not a good heuristic, as people can add their own custom types. We should always get types from `Ash.Type.get_type` to turn any value into its proper type module. If you need to know if it's a builtin there is a function IIRC called `Ash.Type.builtin?/1`

lib/ash/info/manifest/json_serializer.ex
```diff
@@ -78,6 +167,49 @@ defmodule Ash.Info.Manifest.JsonSerializer do
...
+
+  defp serialize_rhs({:concrete, ref}) when is_atom(ref),
+    do: %{"concrete" => concrete_ref(ref)}
+
+  defp serialize_rhs({:array, inner}), do: %{"array" => serialize_rhs(inner)}
+  defp serialize_rhs(_), do: "any"
+
+  defp concrete_ref(ref) do
+    case Atom.to_string(ref) do
+      "Elixir." <> _ -> module_to_string(ref)
```
zachdaniel, line 210: You can `inspect(ref)` if it's an atom? Or if its a module?

zachdaniel:

Issue #2698, opened 2026-05-13
Proposal: Extend `exists/2` to traverse embedded array attributes


nallwhy:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

Ash expressions today let you ask "does this collection contain an element satisfying P?" over relationships via `exists/2`, but there is no expression-level equivalent for an embedded array attribute (`{:array, EmbeddedResource}`).

| Shape | Expression |
|---|---|
| `belongs_to` / `has_one` | `rel.field == x` |
| `has_many` / `many_to_many` | `rel.field == x` (auto-any) or `exists(rel, ...)` |
| Single embedded attribute | `attr[:field]` |
| **Embedded array attribute (`{:array, EmbeddedResource}`)** | **— no expression — must use `fragment`** |

To ask "any line item's name contains `word`" over a resource shaped like `Estimate.options :: {:array, Option}` / `Option.line_items :: {:array, LineItem}`, the only option today is to drop to raw SQL:

```elixir
expr(fragment(
  """
  EXISTS (
    SELECT 1
    FROM jsonb_array_elements(?) AS opt,
         jsonb_array_elements(opt->'line_items') AS li
    WHERE li->>'name' ILIKE ?
  )
  """,
  options,
  ^"%#{word}%"
))
```

This:

- bypasses Ash's type system and predicate composition,
- leaks data-layer specifics (`jsonb_array_elements`, `->>`) into application code,
- requires manual parameter handling,
- is verbose enough that teams denormalize prematurely or skip the feature.

From the user's perspective, "this collection contains an element satisfying P" is the same question regardless of whether the collection is modeled as a `has_many` or an embedded array. The expression surface should reflect that.

### Describe the solution you'd like

Generalize `exists/2` so each path segment is resolved against the resource schema and may be **either**:

- a relationship segment (existing behavior — join / `EXISTS` translation), or
- an attribute segment of type `{:array, EmbeddedResource}` where `EmbeddedResource` is `use Ash.Resource, data_layer: :embedded` (new — unnest the array, evaluate the predicate per element).

The user-facing API is unchanged; dispatch is per-segment and internal.

```elixir
# existing — relationship
exists(invoices, state == :paid)

# new — embedded array
exists(options, total_amt > 100)

# new — nested embedded arrays
[...]

PR #2705, opened 2026-05-17
poc: extend exists/2 to traverse embedded array attributes

nallwhy:
poc of #2698


zachdaniel:
Looking at the Ash PR it generally looks good! Haven't looked too deeply at the others. However, one thing I'd like to see is a test proving that the relationship path referencing works with this. There are example tests in the policy tests that exercise these over exists etc.

nallwhy:
Good call — added a test covering exactly that pattern, mirroring `parent_test.exs:164` (where `parent/1` traverses a `belongs_to` relationship into a related row's attribute) but with an embedded-array target on the outside:

```elixir
exists(options, name == parent(company.name))
```

PR #2706, opened 2026-05-17
fix(manifest): pass real resource to data_layer.functions/1

Torkan:
Previously called functions(nil), which crashed for data layers like AshPostgres that read repo config off the resource. Now passes a representative resource per data layer; misbehavior is rescued + logged.


zachdaniel (approved):

zachdaniel:

PR #2708, opened 2026-05-18
Allow can to not log policy breakdown

sezaru:
This PR basically add a `log_policy_breakdown?` option to `Ash.can?` so users can call it as authorization barriers (I use in LV to enable/disable parts of the UI based on what the user can/cannot do) without flooding the terminal with policies breakdowns

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -979,6 +979,11 @@ defmodule Ash do
       type: :boolean,
       default: false,
       doc: "Whether or not to log the authorization result."
+    ],
+    log_policy_breakdown?: [
+      type: :boolean,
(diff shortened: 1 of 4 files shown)


sezaru:
NOTE: I would prefer to have the `log_policy_breakdowns?` as  false by default for `Ash.can?` calls, but I left the default unsigned (meaning it uses the global value) to keep the current behavior

zachdaniel (approved):

zachdaniel:

Issue #2360, opened 2025-10-09
Auto-detect whether fields from validation errors are attributes or arguments


jsw800:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

When writing a custom validation, if I return

```elixir
{:error, field: :my_argument, message: "message"}
```

from the `validate` callback, it generates an `Ash.Error.Changes.InvalidAttribute` error, even if the specified field is an argument.

### Describe the solution you'd like

It seems to me like it would make sense to automatically detect whether it's an attribute or argument and return the associated exception type.

### Describe alternatives you've considered

This isn't a huge deal because I can call `Ash.Error.Changes.InvalidArgument.exception` directly and return that from the validate callback, but it's a bit more verbose IMO

ken-kost:
Does this mean `Ash.Error.Changes.InvalidArgument` should exist? 🤔 🐛

ThomaseLucas:
I'll be working on this issue!

ThomaseLucas:
@zachdaniel after doing some research, I discovered the cause of this issue:

When a validation returns an error tuple ({:error, field: name, message: "..."}), the changeset always builds "Ash.Error.Changes.InvalidAttribute" regardless of name being an action argument. The logic currently is, if :field is present, then it uses InvalidAttribute. 

The change would happen with do_validation/5 in  changeset.ex, in the {:errors, errors} when is_list(errors) -> branch. A keyword list like [field: name, message: "..."] is still a list, so it hits this branch and gets passed into add_error, which, as I stated earlier, always maps :field to InvalidAttribute regardless if name is an action argument. 

**Proposal:** In that branch, if errors is a keyword list, use changeset.action / resource metadata to determine whether name is an action argument or an attribute, then add_error the appropriate struct (InvalidAttribute vs InvalidArgument) instead of passing the raw keyword through. Non-keyword lists would stay the same as they are now.

zachdaniel:
That sounds right to me 👍

PR #2714, opened 2026-05-21
fix: auto-detect argument vs attribute in validation 

ThomaseLucas:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Resolves issue #2360 

## Summary
When custom validations return `{:error, field: name, message: msg}`, Ash always built `Ash.Error.Changes.InvalidAttribute` using `add_error` → `to_change_error/1`, even if the value of name was an argument.
In `do_validation/5`, single keyword-list errors now use `changeset.action` to choose `InvalidArgument` vs `InvalidAttribute` before calling `add_error`.
## Tests
- I created two tests to ensure this change worked well: 
1. Argument `:token` → expects `%InvalidArgument{}`
2. Attribute `:name` → expects `%InvalidAttribute{}`
Both use a tiny validation that returns `{:error, field:, message:}`.

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -7791,19 +7791,24 @@ defmodule Ash.Changeset do
   defp keyword_to_validation_change_error(changeset, keyword) do
     error =
       if keyword[:field] do
-        exception =
-          if has_argument?(changeset.action, keyword[:field]) do
-            InvalidArgument
-          else
-            InvalidAttribute
-          end
+        field = keyword[:field]
 
-        exception.exception(
-          field: keyword[:field],
+        opts = [
           message: keyword[:message],
           value: keyword[:value],
           vars: keyword
-        )
+        ]
+
+        cond do
+          has_argument?(changeset.action, field) ->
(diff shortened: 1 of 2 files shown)


lib/ash/changeset/changeset.ex
```diff
@@ -7783,6 +7788,38 @@ defmodule Ash.Changeset do
     end
   end
 
+  defp keyword_to_validation_change_error(changeset, keyword) do
+    error =
+      if keyword[:field] do
+        exception =
+          if has_argument?(changeset.action, keyword[:field]) do
```
zachdaniel: I would add three cases:

`has_argument?` -> InvalidArgument

`Ash.Resource.Info.attribute(resource, field)` -> InvalidAttribute

`fallback` -> InvalidChanges

ThomaseLucas: I'll get this in by the end of today!

zachdaniel (approved):

zachdaniel:

PR #2716, opened 2026-05-22
fix: propagate include_source? from union to inner embedded type constraints

Munksgaard:
fix: propagate include_source? from union to inner embedded type constraints
When a union type has `include_source?: true` but the global config
`include_embedded_source_by_default?` is `false`, the inner embedded
type's `include_source` callback falls back to the compile-time default
(`false`) because the variant-level constraints don't include
`include_source?`. This breaks `relate_actor` on nested embedded
resources since the source changeset (carrying the actor) is never set.

Propagate the union's `include_source?` setting to the variant's
config_constraints via `Keyword.put_new_lazy`, so the inner type
inherits the union's opt-in.

~~I don't think this should be merged as is, because it includes changing the `ash` config, but I couldn't find any other way to create a reproducing test. We should either make the test reproducible without the config change or remove the test entirely, I think. Any suggestions?~~

I've removed the config setting, so the PR should be good to go now.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/type/union.ex
+++ b/lib/ash/type/union.ex
@@ -770,11 +770,21 @@ defmodule Ash.Type.Union do
             config[:constraints] || []
           end
 
+        constraints_with_include_source =
+          if Ash.Type.embedded_type?(config[:type]) and
+               function_exported?(config[:type], :include_source, 2) do
+            Keyword.put_new_lazy(config_constraints, :include_source?, fn ->
+              Keyword.get(constraints, :include_source?, @include_source_by_default)
+            end)
+          else
+            config_constraints
+          end
+
         constraints_with_source =
           Ash.Type.include_source(
             config[:type],
             constraints[:__source__],
-            config_constraints
+            constraints_with_include_source
           )
 
         case Ash.Type.cast_input(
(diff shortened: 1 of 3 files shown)


lib/ash/type/union.ex
```diff
@@ -766,7 +766,11 @@ defmodule Ash.Type.Union do
 
         config_constraints =
           if Ash.Type.embedded_type?(config[:type]) do
-            Keyword.put(config[:constraints] || [], :__union_tag__, config[:tag])
+            (config[:constraints] || [])
```
zachdaniel: We only want to do this if `function_exported?(type_module, :include_source, 2)`. Otherwise this makes sense to me 👍

Munksgaard: Done. I've also removed the config setting, so this PR should be good to go.

test/resource/changes/relate_actor_embedded_test.exs
```diff
@@ -0,0 +1,238 @@
...
+# With `config :ash, :include_embedded_source_by_default?, false`,
+# setting `include_source?: true` on the union types should be
+# sufficient to opt-in to source propagation. However, the union's
+# `cast_input` passes `config_constraints` (variant-level constraints)
+# to the inner embedded type's `include_source`, and those constraints
+# don't include `include_source?`. The embedded type then falls back
+# to `@include_source_by_default` (compiled as `false`), dropping the
+# source changeset and losing the actor.
+
+defmodule Ash.Test.Resource.Changes.RelateActorEmbeddedTest do
```
zachdaniel: Can we put this in an existing test case? Don't want to proliferate regression tests.

Munksgaard: Done!

PR #2717, opened 2026-05-22
Improvement: Include all constraints in `Ash.Type.Decimal.generator/1`

davemcorwin:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

## Summary
Currently, `Ash.Type.Decimal` values generated via `Ash.Generator`:
1. do not respect the `:scale` and `:precision` constraints and
2. may result in StreamData errors when the `:greater_than` and `:less_than` constraints are used

This PR refactors `Ash.Type.Decimal.generator/1` to respect all constraints and resolve StreamData errors.

## Details
- Updates `Ash.Type.Decimal.generator/1` to take `:scale` and `:precision` options into account
- Resolves StreamData errors when using `:greater_than` and `:less_than` constraints
- Adds tests for `Ash.Type.Decimal.generator/1`

## Notes
Repeated runs of the test suite demonstrate that the implementation appears stable.

First time contributor, please let me know if you have any questions or suggestions for improvement.

### DIFF
--- a/lib/ash/type/decimal.ex
+++ b/lib/ash/type/decimal.ex
@@ -49,26 +49,34 @@ defmodule Ash.Type.Decimal do
 
   @impl true
   def generator(constraints) do
-    params =
-      constraints
-      |> Keyword.take([:min, :max])
-      |> Enum.map(fn {key, value} ->
-        if Decimal.is_decimal(value) do
-          {key, Decimal.to_float(value)}
-        else
-          {key, value}
-        end
-      end)
+    # Use the more restrictive constraint
+    float_options = [
+      max: minimum(constraints[:less_than], constraints[:max]),
+      min: maximum(constraints[:greater_than], constraints[:min])
+    ]
 
-    params
+    float_options
     |> StreamData.float()
-    |> StreamData.map(&Decimal.from_float/1)
-    #  A second pass filter to account for inaccuracies in the above float -> decimal
+    |> StreamData.map(fn float ->
+      float
+      |> Decimal.from_float()
+      |> Map.update!(:coef, &trunc_coef(&1, constraints[:precision]))
+      |> Map.update!(:exp, &trunc_exp(&1, constraints[:scale]))
+      |> Decimal.normalize()
+    end)
+    #  A second pass filter to account for:
+    # - inaccuracies in the above float -> decimal
+    # - truncated coefficients and/or exponents that result in a number that violates other constraints
     |> StreamData.filter(fn value ->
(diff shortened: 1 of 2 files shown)


lib/ash/type/decimal.ex
```diff
@@ -49,26 +49,34 @@ defmodule Ash.Type.Decimal do
...
+    end)
+    #  A second pass filter to account for:
+    # - inaccuracies in the above float -> decimal
+    # - truncated coefficients and/or exponents that result in a number that violates other constraints
     |> StreamData.filter(fn value ->
-      !(constraints[:max] && Decimal.gt?(value, constraints[:max])) &&
-        (!constraints[:less_than] || Decimal.lt?(value, constraints[:less_than])) &&
-        !(constraints[:min] && Decimal.lt?(value, constraints[:min])) &&
-        (!constraints[:greater_than] || Decimal.gt?(value, constraints[:greater_than]))
+      Enum.all?([
```
zachdaniel: Hmm...is there no deterministic way to guarantee this? Feels unideal to have to filter when we are producing the data.

zachdaniel: Or is this one of those "floats be floats" type things 😆

davemcorwin: I am definitely no expert on this and I made some attempts at being more prescriptive about the number, ex. generating an integer and then making sure it satisfied the constraints. But things were getting messy AND the numbers being generated were distributed very differently than what we typically get. so I opted for a less efficient, but more understandable approach.

In this implementation, truncating the coefficient and exponent to satisfy the precision and scale constraints can result in the value violating the min/max constraints that were used to generate the original float. My assumption is that this won't happen often when users provide constraints that don't conflict.

All of that said, I think this can definitely be improved. The first thing to resolve is how to handle conflicting constraints, ex. `precision: 2, min: 4321` or `scale: 1, min: 0, max: .01`. Not that people will do this on purpose, but having a methodology for prioritizing them may make the implementation much simpler.

I can take a shot at this, just might take me a bit longer.

davemcorwin: Also, is there a reason for using the value `:arbitrary` for scale and precision when not specified, when `nil` is used for the other constraints?

davemcorwin:
@zachdaniel Take a look at the updated version when you get a chance. This includes more explicit manipulation of the value, and no filtering.

zachdaniel (approved):

zachdaniel:

PR #2719, opened 2026-05-24
feat: Add new type callback for `to_simple_equality_comparable`

sevenseacat:
As discussed on Discord (read discussion from [here](https://discord.com/channels/711271361523351632/711271361523351636/1507791237749411900) on). 

### The problem being solved 

The current behaviour for attaching loaded relationship records to their parent works well with a map-based lookup when the primary/foreign keys all support simple equality (with `==`). If they don't, it still works, just with a lot slower 'check every value against every other value to find which parent goes with which child' approach.

My example app uses CiStrings as IDs - so I was hitting the slow path.

In my app (which has 2000 parent `QuestVersion` records and about 12000 child `JournalEntry` records) - 

```elixir
:timer.tc fn -> Ash.read(Resdayn.Codex.Dialogue.QuestVersion, load: [:journal_entries]); :ok end
```

This would generally take 15-20 seconds, but sometimes upwards of 60 seconds (!!) Which is crazy too long. 

### The solution

The new callback, if defined for a type, defines a way of normalizing the primary/foreign keys to something that *can* be used with simple equality. I've implemented it for `Ash.Type.CiString` - it calls `Ash.CiString.to_comparable_string(value)` which returns a normalized string version, which *can* be used with `==`.

The result: The same query that took 15-20 seconds, now takes 200-300ms, most of which is in the actual database query to load the child data.

### Bits I'm not completely confident about

* The code touches three methods of loading relationships - normal, lazy, and lateral. The code for the lateral matching is a bit more complex than the rest and I'm not fully across it, but it appears to work (and has a test to cover it). The test does hit the updated code path (tested by putting a `dbg` statement in there and ensuring it printed). Please review that part well!

* The code doesn't really handle if the supplied callback definition doesn't return the right type, or if the newly-simple-equal-comparable value *isn't* actually simple equality comparable.
  eg. if I define it like this in `Ash.Type.CiString`:
```elixir
def to_simple_equality_comparable(value), do: :froo
```
  then the error is kind of obscure:
```
** (Ash.Error.Unknown)
Bread Crumbs:
  > Exception raised in: Resdayn.Codex.Dialogue.QuestVersion.read

Unknown Error

* ** (MatchError) no match of right hand side value:

    :froo

  (ash 3.26.0) lib/ash/actions/read/read.ex:472: Ash.Actions.Read.do_run/3
    (ash 3.26.0) lib/ash/actions/read/relationships.ex:2069: anonymous fn/2 in Ash.Actions.Read.Relationships.key_fn_for_type/1
    (elixir 1.20.0-rc.3) lib/enum.ex:1428: anonymous fn/4 in Enum.group_by/3
```
Defining it like this in `Ash.Type.CiString`:
```elixir
def to_simple_equality_comparable(value), do: value
```
Will compare the actual CiStrings with `==`, returning false if they don't match in case, so records will be silently dropped from the relationship. 

This might just be a case of making sure we test well for the built-in types, and documenting it clearly for users? (which I think I've done here)

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -1661,7 +1661,9 @@ defmodule Ash.Actions.Read.Relationships do
       # holds *either* a pkey map (`%{id: ...}`) or a single attribute value,
       # depending on how the lateral join was constructed. Dispatch on shape:
       # plain map → pkey path; anything else (including structs like
-      # `%Ash.CiString{}`) → attribute path.
+      # `%Ash.CiString{}`) → attribute path. This mirrors the existing
+      # slow-path dispatch below (`is_map(...)` check around the
+      # `primary_key_matches?` vs `Ash.Type.equal?` branch).
       lateral_source_key = fn
         m when is_map(m) and not is_struct(m) -> pkey_to_key.(m)
         v -> attr_to_key.(v)
@@ -2065,10 +2067,7 @@ defmodule Ash.Actions.Read.Relationships do
         & &1
 
       Ash.Type.simple_equality_comparable?(type) ->
-        fn value ->
-          {:ok, key} = Ash.Type.to_simple_equality_comparable(type, value)
-          key
-        end
+        & Ash.Type.to_simple_equality_comparable(type, &1)
 
       true ->
         nil
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -1122,18 +1122,19 @@ defmodule Ash.Actions.Read.Relationships do
          related_records,
          {:lazy, _related_query}
        ) do
-    if Ash.Resource.Info.primary_key_simple_equality?(relationship.source) do
-      pkey = Ash.Resource.Info.primary_key(resource)
+    pkey = Ash.Resource.Info.primary_key(resource)
+    pkey_to_key = pkey_normalizer(resource, pkey)
 
+    if pkey_to_key do
       records_by_pkey =
         Enum.reduce(related_records, %{}, fn related, acc ->
           Enum.reduce(related.__metadata__.__lazy_join_sources__, acc, fn source, acc ->
-            Map.update(acc, source, [related], &[related | &1])
+            Map.update(acc, pkey_to_key.(source), [related], &[related | &1])
           end)
         end)
 
       Enum.map(records, fn record ->
-        record_pkey = Map.take(record, pkey)
+        record_pkey = pkey_to_key.(Map.take(record, pkey))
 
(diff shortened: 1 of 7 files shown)


lib/ash/actions/read/relationships.ex
```diff
@@ -2022,4 +2045,58 @@ defmodule Ash.Actions.Read.Relationships do
...
+  #   * Neither: returns `nil`. The caller's contract is that `nil` means
+  #     "no safe Map key exists, fall back to the scan-based slow path."
+  defp key_fn_for_type(type) do
+    cond do
+      Ash.Type.simple_equality?(type) ->
+        & &1
+
+      Ash.Type.simple_equality_comparable?(type) ->
+        fn value ->
+          {:ok, key} = Ash.Type.to_simple_equality_comparable(type, value)
```
zachdaniel: since we don't really have anything that handles errors, I'd just make that callback return the value. The assumption can be that its a non-failable operation.

lib/ash/resource/transformers/cache_primary_key.ex
```diff
@@ -23,6 +23,10 @@ defmodule Ash.Resource.Transformers.CachePrimaryKey do
           :primary_key_simple_equality?,
           Enum.all?(attributes, &Ash.Type.simple_equality?(&1.type))
         )
+        |> Transformer.persist(
+          :primary_key_simple_equality_comparable?,
+          Enum.all?(attributes, &Ash.Type.simple_equality_comparable?(&1.type))
```
zachdaniel, line 28: For completeness, we'd likely want to know "are all of the primary key attributes *either* simple equality *or* do they define the comparable converter". In case you had a primary key like `uuid, ci_string`.

sevenseacat: good news, that's what [`Ash.Type.simple_equality_comparable?` returns](https://github.com/ash-project/ash/pull/2719/changes#diff-dd4491b1dc2629842f30d2edee72161d609bc73dff8e10a3b3e61f60b4ffc756R1792-R1795) :D

lib/ash/actions/read/relationships.ex
sevenseacat: I've gone through and refactored that - tidied things up a bit, made the whole callback optional. 👍

lib/ash/resource/transformers/cache_primary_key.ex
zachdaniel: 😆 nice, I didn't put it together.

zachdaniel:

PR #2720, opened 2026-05-25
improvement: expose required? on manifest Argument inputs

Torkan:
Adds a `required?` field on Argument in addition to `allow_nil?`, which was conflating the required presence of the input and if nil values were accepted.

### DIFF
--- a/lib/ash/info/manifest/generator/action_builder.ex
+++ b/lib/ash/info/manifest/generator/action_builder.ex
@@ -74,11 +74,20 @@ defmodule Ash.Info.Manifest.Generator.ActionBuilder do
             true -> attribute.allow_nil?
           end
 
+        required? =
+          cond do
+            name in allow_nil_input -> false
+            name in require_attributes -> true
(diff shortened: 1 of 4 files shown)


zachdaniel:
@Torkan Ash doesn't distinguish between these two things. How can the manifest?

zachdaniel:
Ah, okay I see.

lib/ash/info/manifest/generator/action_builder.ex
```diff
@@ -74,11 +74,20 @@ defmodule Ash.Info.Manifest.Generator.ActionBuilder do
             true -> attribute.allow_nil?
           end
 
+        required? =
+          cond do
+            name in allow_nil_input -> false
+            name in require_attributes -> true
```
zachdaniel, line 80: `require_attributes` is a bad name. It's really equating to `allow_nil?: false`

zachdaniel: Which actually makes this right 😄

zachdaniel (approved):

PR #2722, opened 2026-05-26
fix: Error when passing in {:ok, []} to Ash.load

cheerfulstoic:
Previous I would get:

```
** (ArgumentError) Could not determine a resource from the provided input:

  []
```

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash.ex
+++ b/lib/ash.ex
@@ -2490,9 +2490,7 @@ defmodule Ash do
   def load({:error, error}, _, _), do: {:error, error}
 
   def load({:ok, values}, query, opts) do
-    Ash.Helpers.expect_options!(opts)
-    resource = Ash.Helpers.resource_from_data!(values, query, opts)
-    load(values, query, Keyword.put(opts, :resource, resource))
+    load(values, query, opts)
   end
 
   def load(%Ash.Query{}, _query, _opts) do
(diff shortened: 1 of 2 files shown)


lib/ash.ex
```diff
@@ -2490,9 +2490,7 @@ defmodule Ash do
   def load({:error, error}, _, _), do: {:error, error}
 
   def load({:ok, values}, query, opts) do
-    Ash.Helpers.expect_options!(opts)
-    resource = Ash.Helpers.resource_from_data!(values, query, opts)
-    load(values, query, Keyword.put(opts, :resource, resource))
```
cheerfulstoic, line 2495: The tests still seem to pass 🤷 😅  Happy for another solution if needed, but I guess we'd want test to cover whatever might be missing by removing these first two lines.

zachdaniel: We should instead add a clause for `{:ok, []}` that just returns `{:ok, []}` 😄

cheerfulstoic: I'm happy to do that if that's what you want, though if it works with the one `{:ok, values}` clause for both isn't that better? Again, maybe I'm missing something about what those first two clauses do, but if the tests don't fail then it makes me wonder if they are needed.

Then again, I don't maintain a big project like ash, so I understand being conservative about these things!

zachdaniel: oh, yeah actually you're right.

zachdaniel:

PR #2728, opened 2026-05-29
fix: Arguments are not passed into expression template on relationship filters

Rutgerdj:
The documentation for the `has_many.filter` dsl [mention](https://hexdocs.pm/ash/dsl-ash-resource.html#arguments-9) that it's possible to use `^arg()` template in filter queries. 
But the source action arguments are not added to the template so the `^arg(...)` always resovles to `nil`. 

Added a reproduction test in `test/actions/has_many_test.exs:553`.

Fixed it by passing the source_query.arguments to the `hydrate_refs` function. 
Not entirely sure if it's the best way to fix it. So let me know if another approach should be taken.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel (approved):

zachdaniel:

PR #2731, opened 2026-05-31
fix: match `ci_string` join keys case-insensitively when loading `many_to_many` relationships

sevenseacat:
When a `many_to_many` is loaded as separate queries, the join map was stitched onto destination records keyed on raw attribute values, so types whose `==` differs from their semantic equality (e.g. `ci_string`) were silently dropped. 

This now uses the same `simple_equality_comparable` functionality as `has_many` relationship attachment.

The ETS data layer's lateral-join path had the same latent bug (the final join-row match used raw `==`, flagged with a TODO); use `Ash.Type.equal?/3` there.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel (approved):

zachdaniel:

PR #2741, opened 2026-06-06
Add byte_size validation

charlieaten:
- Introduced a new `byte_size` validation for attributes, allowing validation of string byte sizes with options for minimum, maximum, and exact sizes.
- Added a corresponding `byte_size/1` function to evaluate the byte size of strings in queries.
- Updated documentation to include the new validation and function.
- Added tests to ensure correct functionality of the new validation and function.

An example scenario where I found this useful is when one needs to validate that a password input is at most 72 bytes before passing it to bcrypt for hashing.

# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
The expression is cool, but in general won't work in practice for the data layers that can do real transactionality/atomicity unless we implement `byte_size` in those data layers, i.e `ash_postgres` and `ash_sqlite`.

My suggestion would be to leave out the atomic implementation and expression from this PR, and do that in a follow up if we can find a god way to support that expression in postgres/sqlite data layers?

charlieaten:
> The expression is cool, but in general won't work in practice for the data layers that can do real transactionality/atomicity unless we implement `byte_size` in those data layers, i.e `ash_postgres` and `ash_sqlite`.
> 
> My suggestion would be to leave out the atomic implementation and expression from this PR, and do that in a follow up if we can find a god way to support that expression in postgres/sqlite data layers?

Thank you for the feedback. I did think of this, `octet_length(x)` seems to be a good solution for postgres and sqlite data layers

charlieaten:
> The expression is cool, but in general won't work in practice for the data layers that can do real transactionality/atomicity unless we implement `byte_size` in those data layers, i.e `ash_postgres` and `ash_sqlite`.
> 
> My suggestion would be to leave out the atomic implementation and expression from this PR, and do that in a follow up if we can find a god way to support that expression in postgres/sqlite data layers?

I've made an update to defer support for `atomic/3`. Will be exploring `octet_length` and good equivalent for other data layers for a follow up to bring the support back in

lib/ash/resource/validation/byte_size.ex
```diff
@@ -0,0 +1,145 @@
...
+        case result do
+          {:ok, str_value} ->
+            do_validate(subject, str_value, Enum.into(opts, %{}))
+
+          {:error, error} ->
+            {:error, error}
+        end
+    end
+  end
+
```
zachdaniel, line 84: One small note: we actually can add the atomic variation of it, but it has to only run if the current change is not in `changeset.atomics`. If it is, then we return `{:not_atomic, "an error message"}`. If its not, then we can delegate to the `validate/3` function.

charlieaten: Lemme do that

zachdaniel:
@charlieaten I just pushed a fix to main, can you pull it into this branch? Want to run checks on it 🙇

charlieaten:
> @charlieaten I just pushed a fix to main, can you pull it into this branch? Want to run checks on it 🙇 

I've done a rebase to upstream and pushed the update

charlieaten:
@zachdaniel Does that resolve it?

zachdaniel (approved):

zachdaniel:

charlieaten:
> 🚀 Thank you for your contribution! 🚀
> 
> 

No, thank you for this wonderful project. Once one goes the ash way it's inconceivable building any other way

PR #2742, opened 2026-06-07
fix: preserve changeset context in `generate_many/2`

nallwhy:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

`generate_many/2` routed through `Ash.bulk_create!` but only forwarded a few `context[:private]` keys, dropping the user-supplied changeset `:context` (including `context.shared`). This made it diverge from `generate/1`, which preserves context — create-time changes reading `changeset.context` would see it empty under `generate_many/2`.

Forward the non-private context to `Ash.bulk_create!`, which merges it into each changeset. Added a regression test covering the divergence.

### DIFF
--- a/lib/ash/generator/generator.ex
+++ b/lib/ash/generator/generator.ex
@@ -339,6 +339,9 @@ defmodule Ash.Generator do
         input,
         Map.to_list(changeset_opts)
       )
+      |> Ash.Changeset.set_context(%{
+        private: %{generator_context: Map.get(changeset_opts, :context, %{})}
+      })
       |> then(fn changeset ->
         if opts[:after_action] do
           Ash.Changeset.after_action(changeset, fn _changeset, record ->
@@ -823,7 +826,7 @@ defmodule Ash.Generator do
           authorize?: first.context[:private][:authorize?],
           tenant: first.tenant,
           tracer: first.context[:private][:tracer],
-          context: first.context
+          context: first.context[:private][:generator_context] || %{}
         ]
 
         opts =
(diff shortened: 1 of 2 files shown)


lib/ash/generator/generator.ex
```diff
@@ -822,7 +822,8 @@ defmodule Ash.Generator do
           actor: first.context[:private][:actor],
           authorize?: first.context[:private][:authorize?],
           tenant: first.tenant,
-          tracer: first.context[:private][:tracer]
+          tracer: first.context[:private][:tracer],
+          context: Map.delete(first.context, :private)
```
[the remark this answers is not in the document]
zachdaniel: Yeah, I agree with this review item 👍

lib/ash/generator/generator.ex
```diff
@@ -822,7 +822,8 @@ defmodule Ash.Generator do
           actor: first.context[:private][:actor],
           authorize?: first.context[:private][:authorize?],
           tenant: first.tenant,
-          tracer: first.context[:private][:tracer]
+          tracer: first.context[:private][:tracer],
+          context: first.context
```
zachdaniel: Realizing another issue with this. It's possible for a change running on the changeset to set context that is tailored specifically to those changes. So I don't think that we should do it this way. Instead what we should do is store the `context` option that was provided in the generator in its own special context key, and then carry *that* over. Kind of like this:

```elixir
after_action = first.context[:private][:generator_after_action]
```

nallwhy: @zachdaniel Good call — done!

zachdaniel (approved):

zachdaniel:

Issue #1595, opened 2024-11-12
ash.gen.resource warning can not add to resource list


flip111:
I'm a total noobie to ash and phoenix. I'm getting a bit stuck on a point where i think the user experience could be a bit better.

I ran `mix igniter.new fitness --install ash,ash_sqlite,ash_phoenix --with phx.new --extend sqlite`

After that i want to run

```
mix ash.gen.resource Fitness.Exercise \
  --attribute title:string:required \
  --attribute category:string:required \
  --attribute filename:string:required \
  --attribute upload_date:date \
  --attribute daily_quota:integer \
  --uuid-primary-key id \
  --relationship has_many:repetitions:FitnessAsh.Repetition \
  --default-actions read,create,update,destroy \
  --extend ash_sqlite \
  --timestamps
```

And i see
```
Igniter - Warnings:

* Domain Elixir.Fitness was not an `Ash.Domain`, so could not add `Fitness.Exercise` to its resource list.
```

It's not clear to me whether:
1. I need to add it to the resource list myself
2. I am not supposed to generate ash models in the existing domain. (I thought later this domain is phoenix-only)
3. I need to create a new domain for ash first before continuing here
4. It was not a good choice to generate a new project with `--with phx.new` for a new user and this is more for advanced users who know how to integrate ash with phoenix better

albinkc:
Reasoning number 2 seems the most apt.

https://elixirforum.com/t/ash-gen-resource-fails-partially-if-resource-is-top-level/65535

flip111:
@albinkc thanks that seems to be the way to go

Is there any opportunity to improve UX here by putting the solution into the Ignitor warning or in another place so that it pops up in the terminal? Like this bit `so could not add `Fitness.Exercise` to its resource list` doesn't really seem like something you would want to do, does it?

zachdaniel:
What I think we should actually do here is say "X existed but is not an Ash.Domain, should we add use Ash.Domain to the module?"

gixtrem:
Hey guys, I am from the Ash-Fellow team.
I'd like to work on this. Will implement the suggested fix to prompt the user to add `use Ash.Domain` when the module exists but isn't an Ash.Domain.

PR #2744, opened 2026-06-08
Fix/1595 gen resource domain prompt

gixtrem:
# Contributor checklist

Fixes [[ash-project/ash#1595](https://github.com/ash-project/ash/issues/1595)](https://github.com/ash-project/ash/issues/1595)

## Problem

When `mix ash.gen.resource` is called with a `--domain` option pointing to a module that exists but does not `use Ash.Domain`, the generator emits a confusing warning:

```
Domain Elixir.MyApp.Billing was not an `Ash.Domain`, so could not add `MyApp.Billing.Invoice` to its resource list.
```

No helpful action is taken, leaving the user unsure how to proceed.

## Solution

Updated `add_resource_reference/3` and `add_new_code_interface/5` in `Ash.Domain.Igniter` to check whether the domain module exists as a file in the project. If it does, the generator automatically upgrades it to a proper `Ash.Domain` instead of warning and doing nothing.

## Changes

- **`lib/ash/domain/igniter.ex`**: Added `Igniter.Project.Module.module_exists/2` check in both `add_resource_reference/3` and `add_new_code_interface/5`. If the module exists, injects `use Ash.Domain`, creates a `resources` block with the new resource, and registers the domain in `config.exs` under `:ash_domains`.

## Usage

Before:
```
mix ash.gen.resource MyApp.Billing.Invoice --domain MyApp.Billing
# Warning: Domain Elixir.MyApp.Billing was not an `Ash.Domain`, so could not add MyApp.Billing.Invoice to its resource list.
```

After:
```
mix ash.gen.resource MyApp.Billing.Invoice --domain MyApp.Billing
# Notice: MyApp.Billing exists but is not an `Ash.Domain`. Adding `use Ash.Domain` to it.
# Proposed diff shown — user prompted to confirm
```

## Testing

Unit tests added in `test/domain/igniter_test.exs` covering:

- No confusing warning emitted when module exists but is not a domain
- Resource reference correctly patched into the upgraded module
- Original warning preserved when the domain module does not exist at all

- [ X] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ X] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [X ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/domain/igniter.ex
+++ b/lib/ash/domain/igniter.ex
@@ -82,10 +82,49 @@ if Code.ensure_loaded?(Igniter) do
           end
         end)
       else
-        igniter
-        |> Igniter.add_warning(
-          "Domain #{inspect(domain)} was not an `Ash.Domain`, so could not add `#{inspect(resource)}` to its resource list."
-        )
+
+
+        #  Checks if domain module exist as a file in the project
+        {exists?, igniter} = Igniter.Project.Module.module_exists(igniter, domain)
+
+        # if the module exist as an ash.domain or doesnt exit at all
+        if exists? do
+          app_name = Igniter.Project.Application.app_name(igniter)
+
+          # Shows notice to user on action to take
+          igniter
+          |> Igniter.add_notice(
+            "#{inspect(domain)} exists but is not an `Ash.Domain`. Adding `use Ash.Domain` to it."
+          )
+          # open existing module file for editing to : use Ash.Domain and its resources.
+          |> Igniter.Project.Module.find_and_update_module!(domain, fn zipper ->
+            {:ok,
+             Igniter.Code.Common.add_code(zipper, """
+             use Ash.Domain,
+               otp_app: :#{app_name}
+
+             resources do
+               resource #{inspect(resource)}
+             end
+             """)}
+          end)
+          # registers Domain in congig.exs so Ash knows about it.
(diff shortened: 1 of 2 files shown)


lib/ash/domain/igniter.ex
```diff
@@ -82,10 +82,49 @@ if Code.ensure_loaded?(Igniter) do
...
         end)
       else
-        igniter
-        |> Igniter.add_warning(
-          "Domain #{inspect(domain)} was not an `Ash.Domain`, so could not add `#{inspect(resource)}` to its resource list."
-        )
+
+
+        #  Checks if domain module exist as a file in the project
+        {exists?, igniter} = Igniter.Project.Module.module_exists(igniter, domain)
```
zachdaniel: This code looks the same as in the other branch, could we extract it out into a shared function?

gixtrem: Thanks, @zachdaniel. I've gone ahead and implemented your idea. Kindly let me know how this goes.

zachdaniel:
Looks like failing tests & formatting to resolve.

zachdaniel:

PR #2746, opened 2026-06-10
Allow partial identity join sorting

jechol:
## Summary

Allows `many_to_many` relationship sorting through `parent(join_relationship.field)` when the join resource has a partial identity that proves the join keys are non-nil.

Previously, Ash only treated join keys as unique when they matched the primary key or a non-partial identity. This rejected valid polymorphic join resources like:

```elixir
relationships do
  belongs_to :post, Post, allow_nil?: true
  belongs_to :comment, Comment, allow_nil?: true
  belongs_to :user, User, allow_nil?: true
  belongs_to :staff_group, StaffGroup, allow_nil?: true
end

identities do
  identity :post_user, [:post_id, :user_id],
    where: expr(not is_nil(post_id) and not is_nil(user_id))
end
```

In that example, the partial identity still guarantees a single join row for each post/user pair when both join keys are present. That makes sorting by a join attribute safe:

```elixir
many_to_many :sorted_notification_users, User do
  through NotificationRecipient
  join_relationship :user_notification_recipients
  source_attribute_on_join_resource :post_id
  destination_attribute_on_join_resource :user_id
  sort [expr_sort(parent(user_notification_recipients.order), :integer)]
end
```

## Notes

A regression test is being added in `ash_postgres` with this polymorphic join resource shape.
  
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/relationships.ex
+++ b/lib/ash/actions/read/relationships.ex
@@ -2056,14 +2056,81 @@ defmodule Ash.Actions.Read.Relationships do
 
     is_unique_on_join_keys? =
       Enum.any?(Ash.Resource.Info.identities(relationship.through), fn identity ->
-        is_nil(identity.where) && identity.nils_distinct? &&
-          Enum.all?(identity.keys, &(&1 in join_keys))
+        Enum.all?(identity.keys, &(&1 in join_keys)) &&
+          identity_where_matches_join_keys?(identity, join_keys)
       end)
 
     not (primary_key_is_join_keys? || is_unique_on_join_keys?)
   end
 
   defp is_many_to_many_not_unique_on_join?(_, _, _), do: false
+
+  defp identity_where_matches_join_keys?(%{where: nil} = identity, _join_keys) do
+    identity.nils_distinct?
+  end
+
+  defp identity_where_matches_join_keys?(%{where: where, keys: keys}, join_keys) do
+    case join_key_non_nil_refs(where, join_keys) do
+      {:ok, refs} ->
+        Enum.all?(keys, &(&1 in refs))
+
+      :error ->
+        false
+    end
+  end
+
+  defp join_key_non_nil_refs(
+         %Ash.Query.BooleanExpression{op: :and, left: left, right: right},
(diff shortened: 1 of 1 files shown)


lib/ash/actions/read/relationships.ex
```diff
@@ -2056,15 +2056,82 @@ defmodule Ash.Actions.Read.Relationships do
...
+        Enum.all?(identity.keys, &(&1 in join_keys)) &&
+          identity_where_matches_join_keys?(identity, join_keys)
       end)
 
     not (primary_key_is_join_keys? || is_unique_on_join_keys?)
   end
 
   defp is_many_to_many_not_unique_on_join?(_, _, _), do: false
 
+  defp identity_where_matches_join_keys?(%{where: nil} = identity, _join_keys) do
```
zachdaniel: Hmm....I feel like we don't need this and we can just remove the `is_nil(identity.where)` clause and be good to go? If either side is `nil` then it's not related. So we wouldn't really need this at all? Maybe I'm missing something.

jechol: The case I was worried about is a partial identity like `where: expr(active == true)`.

That identity only guarantees uniqueness among active join rows. If the join relationship used by the `many_to_many` does not also filter to `active == true`, an inactive row for the same source/destination pair can still participate in the relationship. Then the same destination record could be reached through more than one join row, and `parent(join_relationship.order)` would not have a single value.

So I don’t think we can safely accept every partial identity just by removing `is_nil(identity.where)`.

I agree with your point that nil source/destination keys cannot match the `many_to_many` join anyway. I pushed a follow-up that leans on that and removes the `nils_distinct?` requirement here. The remaining check only rejects partial identities whose `where` contains conditions other than non-nil checks on the join keys. Supporting cases like `active == true` when the join relationship has the same filter seems possible, but would require reasoning about the join relationship filter vs. the identity predicate, so I left that broader case out of this PR.

zachdaniel:

PR #2751, opened 2026-06-12
relate_actor with field: raises BadMapError for belongs_to relationships

diogomrts:
## issue

relate_actor(:relationship, field: :some_field) raises BadMapError when used with a belongs_to relationship and the field: value resolves to a scalar (e.g. a UUID string) rather than a related record struct.

## reproduction

```elixir
# Actor is a map/struct where :user_id is a raw UUID string
actor = %{user_id: "some-uuid-matching-a-user-id"}
Post
|> Ash.Changeset.for_create(:create, %{title: "Hello"}, actor: actor)
|> Ash.create!()
# => ** (BadMapError) expected a map, got: "some-uuid-matching-a-user-id"
```

Where the Post resource has:

```elixir
belongs_to :creator, User do
  source_attribute :created_by
end

create :create do
  change relate_actor(:creator, field: :user_id)
end
```

### DIFF
--- a/lib/ash/resource/change/relate_actor.ex
+++ b/lib/ash/resource/change/relate_actor.ex
@@ -73,7 +73,7 @@ defmodule Ash.Resource.Change.RelateActor do
         Changeset.force_change_attribute(
           changeset,
           relationship.source_attribute,
-          Map.get(actor, relationship.destination_attribute)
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2753, opened 2026-06-12
fix: ensure modules are loaded before optional callback checks

### LINKED ISSUE #2752: Notifier load/2 dependencies are skipped when the notifier module is not yet loaded
### Code of Conduct

### AI Policy

### Versions

- ash 3.26.0, reproduced on main (7b15bd2)
- Elixir 1.19.2 / OTP 28

### Operating system

macOS (Darwin 24.6.0)

### Current Behavior

`Ash.Notifier` gates notifier dependency loading on `function_exported?(notifier, :load, 2)` (three call sites in `lib/ash/notifier/notifier.ex`). `function_exported?/3` does not load the module, so in interactive-mode VMs (dev, test) a notifier that has not been called yet reports no `load/2` and its dependencies are silently skipped.

Most visible symptom: `Ash.Notifier.PubSub` publications using `transform: :some_calc` broadcast a `nil` payload for the **first** notification after VM start, then work afterwards — dispatching the notification loads the module as a side effect, so the bug masks itself.

### Reproduction

With any resource publishing via `Ash.Notifier.PubSub` with `transform: :some_calc`:

```elixir
:code.purge(Ash.Notifier.PubSub)
:code.delete(Ash.Notifier.PubSub)
# simulates a freshly started dev/test VM

# subscribe to the topic, run the published action twice:
# 1st broadcast arrives with payload nil, 2nd with the full payload
```

### Expected Behavior

Notifier `load/2` dependencies load regardless of whether the notifier module happens to be loaded, e.g. by checking `Code.ensure_loaded?/1` before `function_exported?/3`.

---

_Drafted with AI assistance (Claude); behavior reproduced and verified against main._

### LINKED ISSUE #2755: Ash.Type.Keyword splice_nil_values/2 override is skipped when the module is not yet loaded
### Code of Conduct

### AI Policy

### Versions

- ash main (7b15bd2)
- Elixir 1.19.2 / OTP 28

### Current Behavior

Same class as #2752, different site: `splice_with_type/3` checks `function_exported?(type, :splice_nil_values, 2)` without ensuring the type module is loaded (`lib/ash/type/type.ex`). In a lazily-loading VM, `Ash.Type.Keyword`'s `splice_nil_values/2` override is missed and keyword values are flattened by the default implementation, producing malformed load results.

### Reproduction

On a clean build, run the composite load tests twice:

```
rm -rf _build/test
mix test test/type/composite_load_test.exs   # 10 tests, 0 failures
mix test test/type/composite_load_test.exs   # 10 tests, 1 failure
```

The first run passes because compiling in-VM loads every module; the second run loads modules lazily and `loaded[:nested]` comes back as the bare tuple `{:nested, doc}` instead of a keyword list. The same effect surfaces in CI when a build cache is restored (one matrix leg of #2753's CI failed exactly this way).

### Expected Behavior

`splice_nil_values/2` overrides are honored regardless of module load state.

---

_Drafted with AI assistance (Claude); behavior reproduced and verified against main._

vasspilka:
This PR fixes a bug that would prevent executing the `load/2` callback when module is lazily loaded (usually dev/tests environments). It is done by doing `Code.ensure_loaded?/1` before checking that the load function is exported

-----AI BELOW------

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Fixes #2752
Fixes #2755

Two sites where an optional-callback check uses `function_exported?/3` without ensuring the module is loaded, so in interactive-mode VMs (dev, test — anywhere modules load lazily) the callback is silently skipped:

1. **Notifier `load/2`** (#2752): notifier dependencies are never loaded; most visibly, the first notification after VM start loses calc-transform payloads (`transform: :some_calc`), broadcasting nil. Fixed with a `Code.ensure_loaded?/1` guard shared by the three call sites. The regression test unloads a purpose-built notifier (placed in `test/support` so the VM can reload it from disk after `:code.delete/1` — modules defined inside test files cannot be reloaded) and asserts its `load/2` dependencies still arrive.

2. **`splice_nil_values/2`** (#2755): `Ash.Type.Keyword`'s override is missed and keyword values get flattened by the default implementation. Reproducible on main without this PR by running `mix test test/type/composite_load_test.exs` twice from a clean build — the first run passes because compiling in-VM loads every module, the second fails. This is also what the SimpleSat `mix test` leg of this PR's first CI run hit (restored build cache → lazy loading); the regression test purges `Ash.Type.Keyword` and asserts keyword shape is preserved.

Both regression tests fail without their lib change. Full suite passes locally (3572 tests, 0 failures). The `mix dialyzer` failures in CI are pre-existing — they fail identically on main's latest runs and the logs contain no mention of the files touched here.

_Prepared with AI assistance (Claude); reproduced, both fixes verified red/green against main, vetted by me._

### DIFF
--- a/lib/ash/notifier/notifier.ex
+++ b/lib/ash/notifier/notifier.ex
@@ -93,7 +93,7 @@ defmodule Ash.Notifier do
     notifier_statements =
       Enum.reduce(notifiers, %{}, fn notifier, acc ->
         statement =
-          if function_exported?(notifier, :load, 2) do
+          if implements_load?(notifier) do
             load(notifier, resource, action)
           else
             []
@@ -145,7 +145,7 @@ defmodule Ash.Notifier do
     notifier_statements =
       Enum.reduce(notifiers, %{}, fn notifier, acc ->
         statement =
-          if function_exported?(notifier, :load, 2) do
+          if implements_load?(notifier) do
             load(notifier, resource, action)
           else
             []
@@ -223,6 +223,13 @@ defmodule Ash.Notifier do
     |> List.flatten()
   end
 
+  # `function_exported?/3` does not load the module, so in interactive-mode
+  # VMs (dev, test) a notifier that has not been called yet would silently
+  # report no `load/2` and its dependencies would never be loaded.
+  defp implements_load?(notifier) do
+    Code.ensure_loaded?(notifier) and function_exported?(notifier, :load, 2)
+  end
+
   # Checks if the record already has NotifierDependencies calculations loaded
   # (meaning the action pipeline pre-loaded them).
   defp has_preloaded_notifier_data?(data) when is_struct(data) do
@@ -249,7 +256,7 @@ defmodule Ash.Notifier do
     notifier_statements =
       Enum.reduce(notifiers, %{}, fn notifier, acc ->
         statement =
-          if function_exported?(notifier, :load, 2) do
+          if implements_load?(notifier) do
             load(notifier, notification.resource, notification.action)
           else
             []
--- a/lib/ash/test.ex
+++ b/lib/ash/test.ex
@@ -88,6 +88,7 @@ defmodule Ash.Test do
     match
   end
 
+  @spec no_errors(Ash.Error.class_module() | nil) :: no_return
   defp no_errors(error_class) do
     message =
       if error_class do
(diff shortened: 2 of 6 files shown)


zachdaniel:

PR #2754, opened 2026-06-12
fix: look up field policies by calc_name when authorizing calculation sorts

jsw800:
I ran onto this while working on adding some calculations that were gated by field policies. The code changes here are basically all claude-generated but look good to me, and the fix seems sensible from my limited knowledge.

Implementation notes (Claude generated the following):

When a resource defines field policies, sorting on a calculation via `Ash.Query.sort_input/2` passed the `%Ash.Query.Calculation{}` struct itself to the field policy lookup, which is keyed by field-name atoms. The lookup always missed, so the calculation was treated as having no authorizing policy and the sort key was silently rewritten to a constant `nil`, returning rows unsorted with no error.

Passed the already-computed `calc_name` atom to the lookup instead. The struct still flows through as the actual sort expression, so authorized calculation sorts now work and unauthorized ones null out as designed.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/policy/authorizer/authorizer.ex
+++ b/lib/ash/policy/authorizer/authorizer.ex
@@ -969,7 +969,7 @@ defmodule Ash.Policy.Authorizer do
             """
 
         {_other, field} ->
-          {[], field, field, context.query.action, context.query.domain}
+          {[], field_name, field, context.query.action, context.query.domain}
       end
 
     resource =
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2757, opened 2026-06-16
fix: respect actor from scope: in bulk actions under require_actor?

emadshaaban92:
Bulk actions (bulk_create/update/destroy, update_many) resolved `scope:`
  into `actor:` too late — after the `require_actor?` check — so a domain with
  `require_actor? true` wrongly raised "no actor" even when scope carried one.

  Fix: resolve scope at the top of each bulk `run` (and update_many), like the
  single-record path already does. 
  
  Bonus: this is also before the tracing span, so span metadata now shows the
  actor from scope instead of nil.

  Tests added in scope_test.exs (all 4 APIs × strategies + telemetry).

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #2765, opened 2026-06-26
fix: pass NotLoaded/ForbiddenField through Ash.Type.rewrite/4 for array types

### LINKED ISSUE #2764: `Ash.Type.rewrite/4` crashes on a `%Ash.NotLoaded{}` value for an embedded array attribute with field policies
### Code of Conduct

### AI Policy

### Versions

- Elixir 1.19.4 / OTP 27, Elixir 1.20.2 / OTP 27
- bug reproduced on: ash 3.26.0 and 3.29.3
- last known good: ash 3.22.2

### Operating system

MacOS Tahoe 26.4 (25E246)

### Current Behavior

I've got a resource with field policies and an `{:array, embedded_resource}` attribute. My code crashes
on ash 3.26.0 like this when you read records where that attribute is left unloaded:

```
** (ArgumentError) you attempted to apply a function on {:array, Brand}.
   Modules (the first argument of apply) must always be an atom
   (ash 3.29.3) lib/ash/type/type.ex: Ash.Type.rewrite/4
   (ash 3.29.3) lib/ash/actions/read/read.ex: Ash.Actions.Read.cleanup_field_auth/3
```

In the Ash code, `cleanup_field_auth` calls
`Ash.Type.rewrite/4` with the value, but the value is `%Ash.NotLoaded{}`, so it ends up
calling `apply/3` on the type tuple `{:array, Brand}`.

### Reproduction

Distilled from my app...

```elixir
defmodule Brand do
  use Ash.Resource, data_layer: :embedded
  attributes do
  end
end

defmodule Profile do
  use Ash.Resource,
    domain: MyApp.Domain,
    authorizers: [Ash.Policy.Authorizer]

  attributes do
    uuid_primary_key :id
    attribute :brands, {:array, Brand}, public?: true
  end

  actions do
    defaults [:create, :read]
  end

  policies do
    policy always() do
      authorize_if always()
    end
  end

  field_policies do
    field_policy :* do
      authorize_if always()
    end
  end
end

# Create a Profile, then read it back WITHOUT selecting :brands (stays NotLoaded):
Profile |> Ash.Query.select([:id]) |> Ash.read!(authorize?: true)
```

I'll also try to make a PR with a failing test case.

### Expected Behavior

Reading a resource with an unloaded `{:array, _}` attribute shouldn't crash. `rewrite/4`
should pass `%Ash.NotLoaded{}` through untouched (like it does `nil`/`[]`), or
`cleanup_field_auth` should skip unloaded fields.

oshoma:
## Problem

`Ash.Type.rewrite/4` handles `{:array, _}` only when the value is a list:

```elixir
def rewrite(_type, nil, _, _), do: nil
def rewrite(_type, [], _, _), do: []
def rewrite({:array, type}, value, _, _) when is_list(value), do: # recurse
def rewrite(type, item, _, _) when not is_list(item), do: get_type(type).rewrite(item, ...)
```

So a `%Ash.NotLoaded{}` / `%Ash.ForbiddenField{}` value for an `{:array, _}` attribute falls
through to the generic clause, which calls `apply({:array, _}, :rewrite, ...)` and raises
`ArgumentError: ... must always be an atom`.

The practical effect is that any app with field policies and an {:array, embedded_resource} attribute will hit an ArgumentError at runtime whenever you do an authorized read and leave the embedded attribute unloaded (not selected) or forbidden. The read 500s instead of returning data. See the example in #2764.

Reproduces on 3.23 through 3.29.3 and `main`.

Fixes #2764.

## Fix
Change the code so that reading a resource with an unloaded {:array, _} attribute returns data instead of crashing. Let rewrite/4 pass %Ash.NotLoaded{} through untouched. This is consistent with the existing `nil` / `[]` clauses.

## Tests

`test/type/type_test.exs` asserts `rewrite/4` passes both sentinels through for `{:array, _}`
types (fails on `main` with the `ArgumentError`, passes with the fix).

`test/policy/field_policy_test.exs` adds a test with a concrete example of 
an embedded array type with a field policy.

Verified on Erlang 29.0 / Elixir 1.20.0.

### DIFF
--- a/lib/ash/type/type.ex
+++ b/lib/ash/type/type.ex
@@ -1737,6 +1737,8 @@ defmodule Ash.Type do
   """
   def rewrite(_type, nil, _rewrites, _constraints), do: nil
   def rewrite(_type, [], _rewrites, _constraints), do: []
+  def rewrite(_type, %Ash.NotLoaded{} = value, _rewrites, _constraints), do: value
+  def rewrite(_type, %Ash.ForbiddenField{} = value, _rewrites, _constraints), do: value
 
   def rewrite({:array, type}, value, rewrites, constraints) when is_list(value) do
     item_constraints = constraints[:items] || []
--- a/test/policy/field_policy_test.exs
+++ b/test/policy/field_policy_test.exs
@@ -93,6 +93,25 @@ defmodule Ash.Test.Policy.FieldPolicyTest do
     end
   end
 
+  describe "loadable array of an embedded type under field policies (#2764)" do
+    test "an {:array, embedded} field forbidden by field policy returns ForbiddenField, not a crash",
+         %{user: user, admin: admin, post: post} do
+      # `admin_notes` is `{:array, AdminNote}`: a loadable embedded type visible only to admins
+      assert %Ash.ForbiddenField{field: :admin_notes, type: :attribute} =
+               Post
+               |> Ash.Query.filter(id == ^post.id)
+               |> Ash.read_one!(authorize?: true, actor: user)
+               |> Map.get(:admin_notes)
+
+      # an admin can still read the field
+      assert [] =
+               Post
+               |> Ash.Query.filter(id == ^post.id)
+               |> Ash.read_one!(authorize?: true, actor: admin)
+               |> Map.get(:admin_notes)
+    end
+  end
+
   describe "rendering fields" do
     test "when creating as a user that cannot see the field, its value is not displayed", %{
       representative: rep,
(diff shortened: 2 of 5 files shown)


zachdaniel:

oshoma:
Thanks, my dude. @zachdaniel you are doing a great thing with Ash, truly
appreciate your gift.

Osh

oshoma:
Thanks, my dude. @zachdaniel you are doing a great thing with Ash, truly appreciate your gift.

PR #2768, opened 2026-06-30
fix: ensure manual action module is loaded before bulk dispatch

diogomrts:
## Problem

The bulk runners decide whether to invoke a manual action's bulk callback (`bulk_create/3`, `bulk_update/3`, `bulk_destroy/3`) with `function_exported?/3`:

```elixir
manual_action_can_bulk? =
  case action.manual do
    {mod, _opts} -> function_exported?(mod, :bulk_update, 3)
    _ -> false
  end
```

`function_exported?/3` returns `false` for a module that **hasn't been loaded yet** — it does not attempt to load it. So in any process where the manual module isn't already loaded, the bulk callback is silently skipped and execution falls back to the per-record path.

For a **soft destroy** this is fatal. A soft destroy is run as a bulk *update*, so the per-record fallback calls `process_non_bulk_result(..., :bulk_update, ...)`, which reads `changeset.context[:bulk_update]`. But a soft-destroy changeset carries a `:bulk_destroy` context, so `:bulk_update` is `nil` and it crashes:

```
** (BadMapError) expected a map, got: nil
    (ash) lib/ash/actions/bulk_manual_action_helpers.ex:138: ...process_non_bulk_result/6
    (ash) lib/ash/actions/update/bulk.ex: ...run_batch/12
```

Because it depends on whether the module happens to be loaded, the failure is intermittent: it surfaces in isolated/async processes (e.g. the first relevant test in a fresh process) and disappears once anything else has loaded the module.

Ash already guards every other optional-callback check this way (e.g. throughout `Ash.DataLayer`: `Code.ensure_loaded?(mod) && function_exported?(...)`). The manual-action bulk dispatch sites are the only exceptions.

## Fix

Guard all six manual-action bulk dispatch sites with `Code.ensure_loaded?/1`, matching the existing convention:

```elixir
Code.ensure_loaded?(mod) and function_exported?(mod, :bulk_update, 3)
```

- `Ash.Actions.Create.Bulk` (2 sites — `:bulk_create`)
- `Ash.Actions.Update.Bulk` (2 sites — `:bulk_update`)
- `Ash.Actions.Destroy.Bulk` (2 sites — `:bulk_destroy`)

## Why the test uses a manual action

The bug lives entirely in the `case action.manual do {mod, _} -> ...` branch. It only triggers for an action whose `action.manual` is a `{module, opts}` tuple where that module defines a `bulk_*` callback that isn't currently loaded. A plain (non-manual) action has `action.manual == nil`, never reaches this code, and cannot reproduce the issue — so the test resource must use a manual action.

This is not an exotic configuration. Extensions wrap **every** tracked action into a manual action transparently. For example, [`ash_events`](https://github.com/ash-project/ash_events) rewrites each event-tracked action — including a plain `destroy ... soft? true` — into `action.manual = {AshEvents.DestroyActionWrapper, opts}` so it can intercept the write and log an event. The resource author writes an ordinary soft destroy and never sees a manual action, but the *compiled* action is manual. A single shared wrapper module backs every tracked resource, so as soon as one bulk soft-destroy runs before that module is loaded, it hits this crash — regardless of which resource it is.

## Test

Added to the existing `bulk_destroy_manual_test.exs` (with a `SoftDestroyManual` module in the shared `test/support/bulk_test_changes.ex`). It unloads the manual module, runs a soft destroy, and asserts success. The support module is the minimal stand-in for a wrapper like `AshEvents.DestroyActionWrapper`: it `use`s both `ManualDestroy` and `ManualUpdate` (a soft destroy runs as an update, so `update/3` + `bulk_update/3` are invoked), and lives in `test/support` so the runtime can reload it from disk after the test unloads it.

Verified the test fails without the fix (`BadMapError`) and passes with it. Full `test/actions/bulk` suite: 411 passing.

### DIFF
--- a/lib/ash/actions/create/bulk.ex
+++ b/lib/ash/actions/create/bulk.ex
@@ -149,7 +149,7 @@ defmodule Ash.Actions.Create.Bulk do
         manual_action_can_bulk? =
           case action.manual do
             {mod, _opts} ->
-              function_exported?(mod, :bulk_create, 3)
+              Code.ensure_loaded?(mod) and function_exported?(mod, :bulk_create, 3)
 
             _ ->
               false
@@ -1244,7 +1244,7 @@ defmodule Ash.Actions.Create.Bulk do
                       %{}
                   end
 
-                if function_exported?(mod, :bulk_create, 3) do
+                if Code.ensure_loaded?(mod) and function_exported?(mod, :bulk_create, 3) do
                   Ash.Resource.ManualCreate.bulk_create(
                     mod,
                     batch,
--- a/lib/ash/actions/destroy/bulk.ex
+++ b/lib/ash/actions/destroy/bulk.ex
@@ -1147,7 +1147,7 @@ defmodule Ash.Actions.Destroy.Bulk do
     manual_action_can_bulk? =
       case action.manual do
         {mod, _opts} ->
-          function_exported?(mod, :bulk_destroy, 3)
+          Code.ensure_loaded?(mod) and function_exported?(mod, :bulk_destroy, 3)
 
         _ ->
           false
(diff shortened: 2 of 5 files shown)


zachdaniel:

PR #2772, opened 2026-07-02
Treat NotLoaded actor fields the same as nil actor in policy checks

### LINKED ISSUE #2057: Ash should raise an error if an actor template expand to not loaded
If a policy references a field on actor that is not loaded, it might incorrectly succeed:
```elixir
    policy action_type(:create) do
      forbid_if expr(^actor(:journals_at_limit?) == true)
      authorize_if always()
    end
```

In the above example, the `:journals_at_limit?` is a calculated field that evaluates to `true`, yet the action succeeds.

Expected behavior: an error should be raised if the field is not loaded.

gixtrem:
Closes #2057

# Contributor checklist

- [x ] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Chores
- [ ] Documentation changes
- [x ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/expr/expr.ex
+++ b/lib/ash/expr/expr.ex
@@ -294,7 +294,7 @@ defmodule Ash.Expr do
   end
 
   @doc false
-  def get_path(%Ash.NotLoaded{}, _), do: nil
+  def get_path(%Ash.NotLoaded{} = not_loaded, _), do: not_loaded
 
   def get_path(map, [key]) when is_map(map) do
     Map.get(map, key)
@@ -306,7 +306,15 @@ defmodule Ash.Expr do
 
   def get_path(_, _), do: nil
 
-  defp raise_if_not_loaded!(%Ash.NotLoaded{}, _actor, _path), do: nil
+  defp raise_if_not_loaded!(%Ash.NotLoaded{}, actor, path) do
+    raise ArgumentError, """
+    Actor field is not loaded: #{inspect(path)}
+
+    Actor: #{inspect(actor)}
+
+    Ensure the field is loaded on the actor before using it in a filter template.
+    """
+  end
 
   defp raise_if_not_loaded!(value, _actor, _path), do: value
 
--- a/lib/ash/expr/expr.ex
+++ b/lib/ash/expr/expr.ex
@@ -294,7 +294,7 @@ defmodule Ash.Expr do
   end
 
   @doc false
-  def get_path(%Ash.NotLoaded{} = not_loaded, _), do: not_loaded
+  def get_path(%Ash.NotLoaded{}, _), do: nil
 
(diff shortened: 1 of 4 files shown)


lib/ash/expr/expr.ex
```diff
@@ -294,7 +294,7 @@ defmodule Ash.Expr do
   end
 
   @doc false
-  def get_path(%Ash.NotLoaded{} = not_loaded, _), do: not_loaded
+  def get_path(%Ash.NotLoaded{}, _), do: nil
```
zachdaniel: This isn't quite right I think. The concept is that we should raise the same way we are raising if a *field* of the actor is not loaded, when the *actor itself* is not loaded. So this should produce *more* errors, not less.

gixtrem:
Thanks for the clarification, @zachdaniel. So, to confirm: the fix should ensure that when the actor is (nil) and a policy references (^actor(:field), it raises an error matching the existing behavior when the field is (%Ash.NotLoaded{}). Essentially, both "no actor" and "unloaded actor field" should produce errors rather than silently evaluating to false. I'll rework the PR accordingly

zachdaniel:
Yes that would be my preference.

gixtrem:
@zachdaniel implementing the raise for nil actor broke 26 existing policy tests, including ones like simple_test.exs's "when no actor is provided, filtering is applied as expected" and the field_policy_test setup, which rely on nil actor being a handled state (deny/filter) rather than an error. It seems the framework intentionally supports anonymous access via nil actor.
Should the raise be scoped more narrowly, for example, only raise on %Ash.NotLoaded{} fields (an actor that exists but wasn't fully loaded), while still allowing nil actor to filter/deny as before? Or did you intend for these no-actor flows to change too? I just want to confirm the intended scope before I rework the existing tests

zachdaniel:
Hmm....damn. You're right I think I've led you astray sorry. I was misreading your original contribution as well as the original issue 😢.

It looks like the code already does what the original author wants it to do at least in some scenarios. Perhaps we need to start by reproducing their desire in a test, showing that not loaded fields on an actor raise an error. That was the original ask from the author. So `actor == nil` is fine, but `actor != nil and the field is not loaded` should be an error.

gixtrem:
Hey @zachdaniel, reworked this as we discussed. Went back and reverted the source changes, and you were right; the framework already does the right thing. actor == nil just resolves the references to nil and keeps evaluating, and a present actor with a NotLoaded field raises as it should. Confirmed both hold for direct field refs and nested paths.

zachdaniel:

Issue #1498, opened 2024-10-04
Handle Notifications when running an action inside an action


maennchen:
**Describe the bug**

When defining an action change, which internally calls another action using `Ash.run_action`, the following warning is logged and the notification is not sent.

```
     11:35:26.111 [warning] Missed 1 notifications in action [REDACTED].
     
     This happens when the resources are in a transaction, and you did not pass
     `return_notifications?: true`. If you are in a changeset hook, you can
     return the notifications. If not, you can send the notifications using
     `Ash.Notifier.notify/1` once your resources are out of a transaction.
     
         (elixir 1.17.3) lib/process.ex:864: Process.info/2
         (ash 3.4.21) lib/ash/actions/helpers.ex:349: Ash.Actions.Helpers.warn_missed!/3
         (ash 3.4.21) lib/ash/actions/action.ex:143: Ash.Actions.Action.run/3
         (ash 3.4.21) lib/ash.ex:1384: Ash.run_action/2
         (ash 3.4.21) lib/ash/actions/action.ex:129: Ash.Actions.Action.run/3
         (ash 3.4.21) lib/ash.ex:1384: Ash.run_action/2
         ([REDACTED] 0.1.0) lib/[REDACTED].ex:27: anonymous fn/2 in [REDACTED].change/3
         (ash 3.4.21) lib/ash/changeset/changeset.ex:3803: anonymous fn/2 in Ash.Changeset.run_after_actions/3
         (elixir 1.17.3) lib/enum.ex:4858: Enumerable.List.reduce/3
         (elixir 1.17.3) lib/enum.ex:2585: Enum.reduce_while/3
         (ash 3.4.21) lib/ash/changeset/changeset.ex:3280: anonymous fn/3 in Ash.Changeset.with_hooks/3
         (ecto_sql 3.12.0) lib/ecto/adapters/sql.ex:1382: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
         (db_connection 2.7.0) lib/db_connection.ex:1756: DBConnection.run_transaction/4
         (ash 3.4.21) lib/ash/changeset/changeset.ex:3278: anonymous fn/3 in Ash.Changeset.with_hooks/3
         (ash 3.4.21) lib/ash/changeset/changeset.ex:3422: anonymous fn/2 in Ash.Changeset.transaction_hooks/2
         (ash 3.4.21) lib/ash/changeset/changeset.ex:3259: Ash.Changeset.with_hooks/3
[...]

PR #2773, opened 2026-07-02
fix: handle notifications when running nested actions

ThomaseLucas:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

in relation to issue #1498

### DIFF
--- a/lib/ash/actions/action.ex
+++ b/lib/ash/actions/action.ex
@@ -77,7 +77,9 @@ defmodule Ash.Actions.Action do
             else
               run_without_transaction(domain, input, module, run_opts, context, opts)
             end
+            |> strip_return_notifications_for_hooks(input, opts)
           end)
+          |> restore_return_notifications_after_hooks(input, opts)
 
         result = maybe_load(result, input, domain, opts)
 
@@ -133,6 +135,20 @@ defmodule Ash.Actions.Action do
   defp maybe_load(:ok, _input, _domain, _opts), do: :ok
   defp maybe_load({:ok, nil}, _input, _domain, _opts), do: {:ok, nil}
 
+  defp maybe_load({:ok, result, notifications}, input, domain, opts) do
+    case maybe_load({:ok, result}, input, domain, opts) do
+      {:ok, loaded} -> {:ok, loaded, notifications}
+      other -> other
+    end
+  end
+
+  defp maybe_load({:ok, notifications}, %{action: %{returns: nil}} = _input, _domain, %{
+         return_notifications?: true
+       })
+       when is_list(notifications) do
+    {:ok, notifications}
+  end
+
   defp maybe_load({:ok, result}, input, domain, opts) do
     constraints = input.action.constraints || []
     returns = input.action.returns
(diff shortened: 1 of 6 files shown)


lib/ash/actions/action.ex
```diff
@@ -154,6 +170,41 @@ defmodule Ash.Actions.Action do
...
+       })
+       when not is_nil(input.action.returns) do
+    Process.put(:ash_return_notifications, notifications)
+    {:ok, result}
+  end
+
+  defp strip_return_notifications_for_hooks({:ok, notifications}, %{action: %{returns: nil}}, %{
+         return_notifications?: true
+       }) do
+    Process.put(:ash_return_notifications, notifications)
```
zachdaniel: We shouldn't need to do this, we ought to be able to update the hook running code to thread through notifications. I'm cool with using process dictionary for ` for the `ash_notifications` key which is what gets passed upward automatically but for these we should just be able to gather them up while running hooks without a need for the pdict.

ThomaseLucas: I removed the pdict workaround that I added initially, and kept :ash_notifications for nested actions. 

Right now I'm peeling/attaching notifications around hooks in action.ex so hooks still see {:ok, result}. Is that okay, or did you want that logic moved into run_after_transaction_hooks in ActionInput? Don't want to keep going in the wrong direction.

lib/ash/actions/action.ex
```diff
@@ -154,6 +170,41 @@ defmodule Ash.Actions.Action do
...
+  defp strip_return_notifications_for_hooks({:ok, result, notifications}, input, %{
+         return_notifications?: true
+       })
+       when not is_nil(input.action.returns) do
+    Process.put(:ash_return_notifications, notifications)
+    {:ok, result}
+  end
+
+  defp strip_return_notifications_for_hooks({:ok, notifications}, %{action: %{returns: nil}}, %{
+         return_notifications?: true
```
zachdaniel: aren't the opts a keyword list here?

lib/ash/actions/action.ex
```diff
@@ -77,7 +77,9 @@ defmodule Ash.Actions.Action do
             else
               run_without_transaction(domain, input, module, run_opts, context, opts)
             end
+            |> strip_return_notifications_for_hooks(input, opts)
```
zachdaniel: It looks like this would be dead code/not needed given your changes to `Ash.ActionInput` here?

test/notifier/notifier_test.exs
```diff
@@ -334,6 +388,42 @@ defmodule Ash.Test.NotifierTest do
     assert_receive {:notification, %Ash.Notifier.Notification{data: %Comment{name: "auto"}}}
   end
 
+  test "a nested generic action notification is sent automatically on create" do
+    Post
```
zachdaniel: These tests use the `ETS` data layer which unfortunately does not support transactions, and therefore doesn't test the transactional path, meaning we never set `ash_started_transaction?`. We want to test these against resources using the `Ash.DataLayer.Mnesia` which does support transactions.

lib/ash/actions/action.ex
```diff
@@ -199,30 +250,15 @@ defmodule Ash.Actions.Action do
...
+                finalize_notifications(
+                  notifications,
+                  input,
+                  opts,
+                  notify?
+                )
+                |> build_result(result, input, opts)
 
-              # Run after_transaction hooks
               Ash.ActionInput.run_after_transaction_hooks(final_result, input)
```
zachdaniel: We're now passing in `{:ok, result, notifications}` into here which would break existing callers. We'd have to strip the notifications and reattach them after this.

lib/ash/actions/action.ex
ThomaseLucas: I totally forgot to push the changes I made, I will do that now, then you can review what changes I made. I'm sorry for the confusion.

PR #2778, opened 2026-07-07
Added Support for Aggregated Policy Evaluator

CyanideDragon:
# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

Created a solution for issue 2432

Ash only had Ash.can?/3 which only checks one permission at a time. A batch API was suggested and has been created for this PR. LiveView no longer has to call Ash.can? repeatedly. Actor, tenant, and context options are remembered so that they do not have to called again with each permission.

We create a shared context map that removes repetition of data that the normal method would do. Then we normalize the checks by turning the inputs into a uniform internal list of {key, subject, per_check_opts}. Then we use Enum.reduce for each check, merge the shared opts with any per-check opts, resolve the domain, and then call the existing can/3 all without duplicating any policy logic. Finally the results are either a list or a map depending on how the input was structured. This will allow the UI to use the set of permissions in an easier way.


zachdaniel:
In general its best to create a real PR title that describes what the change does 🙇

CyanideDragon:
@zachdaniel I'm sorry, I have changed it now. I also updated the original comment to provide a little more context.

zachdaniel:

zachdaniel:
Also: very cool addition :sunglasses

PR #2782, opened 2026-07-13
improvement: add `error_handler` to generic actions

jimsynz:
## Summary

Generic actions run on `Ash.ActionInput` rather than a changeset, so the `error_handler` DSL option available on create/update/destroy actions had no equivalent. This adds it, and routes both errors *returned* from and exceptions *raised* by the action's `run` implementation through the handler.

## Changes

- **`lib/ash/action_input.ex`** — added a `handle_errors` field (default `nil`) to the struct + typespec, a `handle_errors/2` setter (2-arity fun or mfa), and dispatch in `handle_error/2` that interprets the same return values as the changeset (`:ignore`, `{:ignore, input}`, an `%ActionInput{}`, `{input, error}`, or a transformed error). Wired into `for_action/4` before casting/validation so those errors flow through the handler.
- **`lib/ash/actions/action.ex`**:
  - Errors returned from the `run` implementation (and other run-phase errors) are piped through the input's handler at the single chokepoint in `run_with_lifecycle`, mirroring `Ash.Changeset.add_error` at the commit stage of create/update/destroy.
  - Exceptions raised in the implementation are converted to an Ash error, routed through the handler, then reraised as an error class.
- **`lib/ash/resource/actions/action/action.ex`** — added `:error_handler` to the struct + typespec and exposed the `error_handler` option on the generic action schema.
- **`documentation/dsls/DSL-Ash.Resource.md`** — regenerated cheat sheet.
- **`test/actions/generic_action_error_handler_test.exs`** — covers fun handler, mfa handler, `:ignore`, errors returned from `run`, and exceptions raised in `run`.

## Scope / notes

- Read actions still don't support `error_handler` (they run on `Ash.Query`, which has no `handle_errors`). This change is scoped to generic actions only.

### DIFF
--- a/documentation/dsls/DSL-Ash.Resource.md
+++ b/documentation/dsls/DSL-Ash.Resource.md
@@ -977,6 +977,7 @@ end
 | [`constraints`](#actions-action-constraints){: #actions-action-constraints } | `keyword` |  | Constraints for the return type. See `Ash.Type` for more. |
 | [`allow_nil?`](#actions-action-allow_nil?){: #actions-action-allow_nil? } | `boolean` | `false` | Whether or not the action can return nil. Unlike attributes & arguments, this defaults to `false`. |
 | [`run`](#actions-action-run){: #actions-action-run } | `(any, any -> any) \| module \| module` |  | Module may be an `Ash.Resource.Actions.Implementation` or `Reactor`. |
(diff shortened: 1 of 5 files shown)


zachdaniel (approved):

zachdaniel:

PR #2788, opened 2026-07-16
improvement: redact keyset values in `InvalidKeyset` messages

### LINKED ISSUE #2785: InvalidKeyset error message includes the cursor value, which is reversible to row data (sort values + primary key)
**Describe the bug**

`Ash.Error.Page.InvalidKeyset.message/1` interpolates the offending keyset value into the error message:

```elixir
def message(%{value: value, key: nil}) do
  "Invalid value provided as a keyset: #{inspect(value)}"
end
```

A keyset cursor is `:erlang.term_to_binary/1` + `Base.encode64/1` over the record's **sort-field values plus the primary key** (`Ash.Page.Keyset.keyset/2`). So the error message — which commonly ends up in logs, error trackers, and API error responses — carries data that is trivially reversible to row values:

```elixir
cursor |> Base.decode64!() |> :erlang.binary_to_term()
#=> [20, "5c9fd71b-1c5c-4a9c-a5d2-0723a8f4db31"]   # sort value + pk
```

If the sort field is sensitive (or the pk is), a routine "invalid keyset" error (e.g. a caller applying a cursor to a query whose sort shape changed between requests) leaks that data into logs. The `zip_fields/4` mismatch path (`Ash.Page.Keyset.filters/4`) additionally raises with the fully **decoded** value list.

**To Reproduce**

```elixir
Mix.install([{:ash, "~> 3.29"}])

defmodule Repro.Doc do
  use Ash.Resource, domain: Repro.Domain, data_layer: Ash.DataLayer.Ets

  attributes do
    uuid_primary_key :id
    attribute :title, :string, public?: true
    attribute :score, :integer, public?: true
  end

  actions do
    default_accept [:title, :score]
    defaults [:create]

    read :read do
      primary? true
      pagination keyset?: true, required?: false
    end
  end
end

defmodule Repro.Domain do

  resources do
    resource Repro.Doc
  end
end

for i <- 1..4 do
  Repro.Doc
  |> Ash.Changeset.for_create(:create, %{title: "t#{i}", score: i * 10})
  |> Ash.create!()
end

page1 = Repro.Doc |> Ash.Query.sort(score: :asc) |> Ash.read!(page: [limit: 2])
cursor = List.last(page1.results).__metadata__.keyset

# The cursor is reversible to row data:
IO.inspect(cursor |> Base.decode64!() |> :erlang.binary_to_term())
#=> [20, "<pk-uuid>"]

# Apply it to a different sort shape -> the error message carries the cursor:
Repro.Doc
|> Ash.Query.sort(score: :asc, title: :asc)
|> Ash.read!(page: [limit: 2, after: cursor])
#=> ** (Ash.Error.Invalid) ...
#   * Invalid value provided as a keyset: "g2wAAAACYRRtAAAAJDVjOWZk..."
```

Observed on ash 3.29.3 / Elixir 1.19.5 / OTP 28.

**Expected behavior**

The error message names the failure class (and optionally the `key`) without embedding the cursor value — e.g. `"Invalid value provided as a keyset"` / `"keyset does not match the query's sort (expected N fields)"`. Callers who need the offending value still have it (they supplied it); logs and API error bodies shouldn't.

**Context**

Found while building keyset support for an Ash data layer whose error contract is value-free (no row data in any error message); happy to PR the message change if the direction is agreeable.

nseaSeb:
Closes #2785, following @zachdaniel's suggestion there:

> I think we can enrich the error message to redact sensitive fields like we do elsewhere, WDYT?

## The problem

A keyset cursor is `:erlang.term_to_binary/1` + `Base.encode64/1` over the record's sort values and primary key, so it is reversible:

```elixir
cursor |> Base.decode64!() |> :erlang.binary_to_term()
#=> [20, "5c9fd71b-1c5c-4a9c-a5d2-0723a8f4db31"]   # sort value + pk
```

`Ash.Error.Page.InvalidKeyset.message/1` interpolated that cursor into the message, which is what ends up in logs and error trackers.

## The change

`message/1` now passes the value through `Ash.Helpers.redact/1` when `redact_sensitive_values_in_errors?` is set, as `Ash.Resource.Validation.maybe_redact/3` does for validation errors. Both raise sites (`Ash.Page.Keyset.decode_values/2` and the `zip_fields/4` mismatch) go through `message/1`, so both are covered.

## Two decisions worth a look

**The whole cursor is redacted, not only sensitive fields.** Per-field detection is not reliable at the `zip_fields` mismatch site: the `sort` available there is the *current* query's sort, not the sort that built the cursor. A cursor built from a sensitive field and then applied to a non-sensitive sort is exactly the mismatch case, so a per-field check would decide not to redact and the value would still be printed. Since the cursor is always reversible, redacting it as a whole has no such hole. Happy to switch if you'd rather have per-field.

**The raw value stays on the error's `value` field.** This matches the scope in your comment (the issue is the logging; the caller already supplied the cursor). It does differ from the validation precedent, where the value is redacted *before* being stored, so the struct is clean too — an error tracker that serializes struct fields rather than `Exception.message/1` would still capture the cursor. If you want the precedent matched exactly, that's an `exception/1` override redacting `:value` in the opts; I left it out to keep this small, glad to do it here or as a follow-up.

Gated on `redact_sensitive_values_in_errors?` (default `false`), so this is opt-in for existing apps and on for apps generated by the installer.

## Tests

Added to `test/actions/pagination_test.exs`: a real cursor applied to a mismatching sort, asserting the cursor string is absent from the message. It fails on `main` (the message contains the base64 cursor) and passes here. Full suite green (3645 tests, 0 failures) on Elixir 1.19.5 / OTP 28.

### DIFF
--- a/lib/ash/error/page/invalid_keyset.ex
+++ b/lib/ash/error/page/invalid_keyset.ex
@@ -8,21 +8,10 @@ defmodule Ash.Error.Page.InvalidKeyset do
   use Splode.Error, fields: [:value, :key], class: :invalid
 
   def message(%{value: value, key: nil}) do
-    "Invalid value provided as a keyset: #{inspect(maybe_redact(value))}"
+    "Invalid value provided as a keyset: #{inspect(value)}"
   end
 
   def message(%{value: value, key: key}) do
-    "Invalid value provided as a keyset for #{to_string(key)}: #{inspect(maybe_redact(value))}"
-  end
-
-  # a keyset encodes the record's sort values and primary key, and is reversible
-  # with `Base.decode64/1` + `:erlang.binary_to_term/1`, so the message is redacted
-  # as a whole. The offending value remains on the `value` field of the error.
-  defp maybe_redact(value) do
-    if Application.get_env(:ash, :redact_sensitive_values_in_errors?, false) do
-      Ash.Helpers.redact(value)
-    else
-      value
-    end
+    "Invalid value provided as a keyset for #{to_string(key)}: #{inspect(value)}"
   end
 end
--- a/lib/ash/error/page/invalid_keyset.ex
+++ b/lib/ash/error/page/invalid_keyset.ex
@@ -8,10 +8,21 @@ defmodule Ash.Error.Page.InvalidKeyset do
   use Splode.Error, fields: [:value, :key], class: :invalid
 
   def message(%{value: value, key: nil}) do
-    "Invalid value provided as a keyset: #{inspect(value)}"
+    "Invalid value provided as a keyset: #{inspect(maybe_redact(value))}"
   end
 
   def message(%{value: value, key: key}) do
-    "Invalid value provided as a keyset for #{to_string(key)}: #{inspect(value)}"
+    "Invalid value provided as a keyset for #{to_string(key)}: #{inspect(maybe_redact(value))}"
+  end
+
+  # a keyset encodes the record's sort values and primary key, and is reversible
+  # with `Base.decode64/1` + `:erlang.binary_to_term/1`, so the message is redacted
+  # as a whole. The offending value remains on the `value` field of the error.
+  defp maybe_redact(value) do
+    if Application.get_env(:ash, :redact_sensitive_values_in_errors?, false) do
+      Ash.Helpers.redact(value)
+    else
+      value
+    end
   end
 end
--- a/lib/ash/page/keyset.ex
+++ b/lib/ash/page/keyset.ex
@@ -117,7 +117,34 @@ defmodule Ash.Page.Keyset do
     with {:ok, decoded} <- decode_values(values, after_or_before),
          {:ok, zipped} <- zip_fields(sort, decoded, values) do
       {:ok, filters(Enum.with_index(zipped), resource, query, after_or_before)}
+    else
+      {:error, %Ash.Error.Page.InvalidKeyset{} = error} ->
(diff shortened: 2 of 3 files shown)


nseaSeb:
I saw the CI/CD failure, it's on main too.
Same 128 :picosat_nif.solve/1 is undefined errors, e.g. run 29471771317. Doesn't look related to this PR.

lib/ash/error/page/invalid_keyset.ex
```diff
@@ -8,10 +8,21 @@ defmodule Ash.Error.Page.InvalidKeyset do
...
-    "Invalid value provided as a keyset for #{to_string(key)}: #{inspect(value)}"
+    "Invalid value provided as a keyset for #{to_string(key)}: #{inspect(maybe_redact(value))}"
+  end
+
+  # a keyset encodes the record's sort values and primary key, and is reversible
+  # with `Base.decode64/1` + `:erlang.binary_to_term/1`, so the message is redacted
+  # as a whole. The offending value remains on the `value` field of the error.
+  defp maybe_redact(value) do
+    if Application.get_env(:ash, :redact_sensitive_values_in_errors?, false) do
+      Ash.Helpers.redact(value)
```
zachdaniel: This just redacts the value in 100% of cases, even if there are no sensitive attributes in the errors. It doesn't matter to redact fields the user doesn't have access to, only `sensitive?: true` attributes.

zachdaniel:

PR #2789, opened 2026-07-16
Preserve embedded resource constraint errors

BrendanWillis:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Preserves detailed validation errors when invalid values are found in embedded resources.

Previously, an embedded value could fail during dumping and lose useful constraint information. This caused the final error to have less detail than expected.

**Changes**

- Reapply attribute constraints when dumping an embedded value fails.
- Preserve detailed validation information when a constraint error is available.
- Convert supported keyword error inputs into structured Ash errors.
- Preserve the field, embedded path, invalid value, and constraint details.
- Add regression tests for embedded-resource errors and keyword error normalization.

### DIFF
--- a/lib/ash/error/error.ex
+++ b/lib/ash/error/error.ex
@@ -53,6 +53,7 @@ defmodule Ash.Error do
   def to_ash_error(value, stacktrace \\ nil, opts \\ []) do
     value =
       value
+      |> normalize_error_input(opts)
       |> List.wrap()
       |> Enum.map(fn
         %struct{} = changeset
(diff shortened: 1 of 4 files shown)


lib/ash/error/error.ex
```diff
@@ -135,6 +137,62 @@ defmodule Ash.Error do
...
+      |> maybe_set_error_path(keyword)
+    else
+      keyword
+    end
+  end
+
+  defp normalize_error_input(other, _opts), do: other
+
+  defp supported_error_keyword?(keyword) do
+    Enum.any?([:field, :fields, :message, :path, :index, :value, :private_vars], fn key ->
```
zachdaniel, line 153: I think this is a bit of a code smell, us having to detect a special kind of keyword list. In general I think this should be detected in the code for embedded types and handled at the source of these errors.

PR #2791, opened 2026-07-18
improvement: raise when a keyset sort field is forbidden by a field policy

### LINKED ISSUE #2786: Keyset cursor over a field-policy-redacted sort field encodes the redaction marker — page 2 silently returns no rows
**Describe the bug**

Keyset cursors are computed from the record **after** field policies redact it (`Ash.Page.Keyset.data_with_keyset/3` reads `Map.get(record, field)`). When the sort field is protected by a field policy the actor fails, the cursor encodes the redaction marker instead of the sort value:

```elixir
cursor |> Base.decode64!() |> :erlang.binary_to_term()
#=> [#Ash.NotLoaded<:attribute, field: :secret_score, ...>, "<pk-uuid>"]
```

Requesting the next page with that cursor then builds a keyset filter comparing rows against the marker term — on the ETS data layer page 2 **silently returns 0 results** while more rows exist (no error, pagination just ends early). On a data layer that pushes the comparison down, it surfaces as a confusing filter error carrying the `%Ash.ForbiddenField{}` struct. Either way a non-admin actor sorting by a field-policy-protected field cannot paginate past page 1, and the failure mode on the reference data layer is silent.

(There's also a small information-exposure angle: the cursor is `term_to_binary`+Base64, so whatever gets encoded — marker or value — is caller-decodable. Computing cursors from **pre-redaction** values would fix pagination but would leak the protected value through the cursor; rejecting a keyset sort over a field the actor can't read, or omitting the redacted field from the cursor with a loud error, avoids both.)

**To Reproduce**

```elixir
Mix.install([{:ash, "~> 3.29"}, {:simple_sat, "~> 0.1"}])

defmodule Repro.Doc do
  use Ash.Resource,
    domain: Repro.Domain,
    authorizers: [Ash.Policy.Authorizer]

  attributes do
    uuid_primary_key :id
    attribute :title, :string, public?: true
    attribute :secret_score, :integer, public?: true
  end

  policies do
    policy always() do
      authorize_if always()
    end
  end

  field_policies do
    field_policy :secret_score do
      authorize_if actor_attribute_equals(:admin, true)
    end

    field_policy :* do
      authorize_if always()
    end
  end

  actions do
    default_accept [:title, :secret_score]
    defaults [:create]

    read :read do
      primary? true
      pagination keyset?: true, required?: false
    end
  end
end

defmodule Repro.Domain do

  resources do
    resource Repro.Doc
  end
end

for i <- 1..4 do
  Repro.Doc
  |> Ash.Changeset.for_create(:create, %{title: "t#{i}", secret_score: i * 10},
    authorize?: false
  )
  |> Ash.create!()
end

# Non-admin actor, sorted by the protected field:
page1 =
  Repro.Doc
  |> Ash.Query.sort(secret_score: :asc)
  |> Ash.read!(actor: %{admin: false}, page: [limit: 2])

cursor = List.last(page1.results).__metadata__.keyset

IO.inspect(cursor |> Base.decode64!() |> :erlang.binary_to_term())
#=> [#Ash.NotLoaded<:attribute, field: :secret_score, ...>, "<pk>"]

page2 =
  Repro.Doc
  |> Ash.Query.sort(secret_score: :asc)
  |> Ash.read!(actor: %{admin: false}, page: [limit: 2, after: cursor])

IO.inspect(length(page2.results))
#=> 0        (4 rows exist; 2 were returned on page 1 — pagination silently ends)
```

Observed on ash 3.29.3 / Elixir 1.19.5 / OTP 28.

**Expected behavior**

One of (maintainers' call):
1. A keyset-paginated read sorted by a field the actor cannot read is rejected loudly (clear error at page 1), or
2. the redacted field is excluded from the cursor with a loud error when the cursor can't be built —

rather than a cursor that encodes the redaction marker and silently terminates pagination.

**Context**

Found while building keyset support for an Ash data layer with fail-closed field-classification semantics; happy to help with a PR once a direction is picked.

nseaSeb:
Closes #2786.

Following the discussion there — `Ash.Query.sort` is opting in, but the actor still can't read the field, so the field policy redacts it and keyset pagination silently breaks.

## What happens today

A keyset-paginated read sorted on a field a field policy forbids the actor from reading:

```elixir
# non-admin, secret_score protected by a field policy
Doc |> Ash.Query.sort(secret_score: :asc) |> Ash.read!(actor: %{...}, page: [limit: 2])
```

The field is deselected, so the cursor is built from the redaction marker (`%Ash.NotLoaded{}`) instead of the value. Page 1 is fine (2 rows, correct order); page 2 seeks against that marker and returns 0 of the remaining rows — no error, pagination just ends early.

## The change

Rather than building an unusable cursor, this detects a statically-forbidden keyset sort field and raises `Ash.Error.Page.KeysetSortOnForbiddenField`. Two things worth calling out about how:

- **It raises before the cursor is built, not by inspecting a marker.** The check runs at query preparation, next to `deselect_known_forbidden_fields` (which already computes the statically-forbidden fields — I extracted that list into `known_forbidden_fields/4`). So there's no "detect the marker on the built cursor" code to look for; the read is rejected upstream and the error threads through the existing `with` (so `Ash.read` returns `{:error, _}` and `Ash.read!` raises).
- **`sort_input` is unaffected.** Sorts from `Ash.Query.sort_input/2` are already rewritten to be policy-aware (they encode `nil`, not a marker), so they keep paginating.

## Scope

This covers **static** field policies — the case in #2786's repro. **Record-dependent** field policies (e.g. `owner_id == actor(:id)`) behave differently and are **not** handled here: the field is loaded to sort, so the cursor is built from the real value. That's a leak rather than an empty page, and a separate concern — flagging it so this isn't read as fully closing that class.

One deliberate choice for review: the check is gated on the read actually paginating (`query.page`). A keyset-action read with no page opts sorted on a forbidden field still attaches a marker cursor as metadata without raising. Happy to drop that gate if you'd rather it be unconditional.

## Tests

`test/policy/field_policies/keyset_pagination_test.exs`: forbidden plain sort raises, admin (who can read the field) still paginates, a readable field still paginates, and the existing `sort_input` test still passes. Full suite green.

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -491,20 +491,25 @@ defmodule Ash.Actions.Read do
                      opts[:authorize?],
                      false
                    ) do
-              data
-              |> Helpers.restrict_field_access(query)
-              |> add_tenant(new_query)
-              |> attach_fields(opts[:initial_data], initial_query, query, missing_pkeys?)
-              |> cleanup_field_auth(query)
-              |> add_page(
-                query.action,
-                count,
-                query.sort,
-                query,
-                new_query,
-                opts
-              )
-              |> add_query(query, opts)
+              restricted =
+                data
+                |> Helpers.restrict_field_access(query)
+                |> add_tenant(new_query)
+                |> attach_fields(opts[:initial_data], initial_query, query, missing_pkeys?)
+                |> cleanup_field_auth(query)
+
+              with :ok <- reject_forbidden_keyset_sort(restricted, query) do
+                restricted
+                |> add_page(
+                  query.action,
+                  count,
+                  query.sort,
+                  query,
+                  new_query,
+                  opts
+                )
+                |> add_query(query, opts)
+              end
             else
               {:error, %Ash.Query{errors: errors} = query} ->
                 {:error, Ash.Error.to_error_class(errors, query: query)}
@@ -1108,19 +1113,24 @@ defmodule Ash.Actions.Read do
                         opts[:authorize?],
                         false
                       ) do
-                 data
-                 |> Helpers.restrict_field_access(query)
-                 |> add_tenant(query)
-                 |> attach_fields(nil, initial_query, query, false)
-                 |> cleanup_field_auth(query)
-                 |> add_page(
-                   query.action,
-                   count,
-                   query.sort,
-                   initial_query,
-                   query,
-                   opts
-                 )
+                 restricted =
+                   data
+                   |> Helpers.restrict_field_access(query)
+                   |> add_tenant(query)
+                   |> attach_fields(nil, initial_query, query, false)
+                   |> cleanup_field_auth(query)
+
+                 with :ok <- reject_forbidden_keyset_sort(restricted, query) do
+                   add_page(
+                     restricted,
+                     query.action,
+                     count,
+                     query.sort,
+                     initial_query,
+                     query,
+                     opts
+                   )
+                 end
                else
                  {:error, %Ash.Query{errors: errors} = query} ->
                    {:error, Ash.Error.to_error_class(errors, query: query)}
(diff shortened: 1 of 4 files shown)


lib/ash/actions/read/read.ex
```diff
@@ -3036,6 +3050,58 @@ defmodule Ash.Actions.Read do
     end
   end
 
+  # A keyset cursor seeks against the sort field's value. When a field policy
+  # statically forbids this actor from reading a sort field, that field is
+  # deselected and the cursor would be built from the redaction marker, silently
+  # breaking pagination. Raise instead. Sorts coming from `Ash.Query.sort_input/2`
+  # are rewritten to be policy-aware before this point, so they are not affected.
+  defp validate_keyset_sort_readable(
```
zachdaniel: I think this has to happen on the resulting data actually. Because a field might not be statically forbidden, but forbidden via a calculation in the query. So when generating the keyset on a result we should raise the error there.

PR #2793, opened 2026-07-22
fix: enforce sortable fasle consistently

riegger-tech:
# Problem
Top-level sorts only checked whether the data layer supported the field type, allowing fields configured with sortable?: false to be sorted. Expression sorts could bypass the restriction by referencing unsortable fields or traversing unsortable relationships. This makes sortable unusable for use right now as a wrong sort could kill our production database.

# Summary

Enforces sortable?: false consistently across Ash.Sort.parse_input/2, Ash.Query.sort_input/3, Ash.Query.sort/3, and Ash.Resource.Info.sortable?/3.
Adds ETS-backed regression coverage for attributes, aggregates, calculations, expressions, and nested sorts.

# AI Disclosure

I manually verified the bug. I used AI to analyse the problem and generate the fix. I manually reviewed and iterated on the fix and manually tested it with our production code

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/resource/info.ex
+++ b/lib/ash/resource/info.ex
@@ -982,6 +982,9 @@ defmodule Ash.Resource.Info do
       nil ->
         false
 
+      %{sortable?: false} ->
+        false
+
       %{type: {:array, _}} ->
         false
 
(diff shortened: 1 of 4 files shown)


zachdaniel:

PR #2795, opened 2026-07-23
improvement: add strict? constraint to Ash.Type.UUIDv7

### LINKED ISSUE #2794: Ash.Type.UUIDv7 accepts non-v7 uuid strings at cast but cannot load the bytes it then stores
Observed on ash 3.29.3; the code is unchanged on main (3.30.1).

`cast_input/2` only checks the version bits when the input is a raw 16-byte binary (the `<<_::48, 7::4, _::12, 0b10::2, _::62>>` clause). A uuid in string form skips that clause and hits the generic hex-format clause, which accepts any version. `dump_to_native/2` uses `Ash.UUIDv7.decode/1`, which is also version-agnostic, so the value writes to Postgres fine. `cast_stored/2` delegates back to `cast_input/2`, where the stored raw bytes now do hit the version-checked binary clauses, and anything non-v7 returns `:error`.

Repro:

```elixir
v4 = Ecto.UUID.generate()
Ash.Type.cast_input(Ash.Type.UUIDv7, v4)
#=> {:ok, "b24891cd-c05a-4169-9a60-31aa5a3259f5"}

{:ok, raw} = Ash.Type.dump_to_native(Ash.Type.UUIDv7, v4, [])
Ash.Type.cast_stored(Ash.Type.UUIDv7, raw, [])
#=> :error
```

In an action this means a v4 uuid argument passes input validation, the insert commits, and loading the RETURNING value raises `ArgumentError: cannot load <<...>> as type Ash.Type.UUIDv7.EctoType`, rolling back the transaction. The failure surfaces after every validation has passed. In our case the action charges a card through a billing vendor before the insert, so the late crash cost us the idempotency record for a completed charge (a caller-generated `:uuid_v7` argument doubles as the idempotency key; the rollback erased it).

The `match_v4_uuids?` config (#2416) is the closest existing knob, but it widens `cast_input` to accept raw v4 binaries, so it loosens loads rather than tightening casts, and it only covers v4.

The two callbacks should agree:

- If the type means "must be v7", `cast_input` should reject non-v7 strings the way it already rejects non-v7 raw binaries. Callers get a validation error before anything writes.
- If the type means "uuid, generated as v7 by default", `cast_stored` should load any uuid version, like `Ash.Type.UUID` does.

Today it validates like the second and loads like the first, and the gap only shows up after a successful write. Per discussion, a PR follows adding an opt-in `strict?` constraint that rejects non-v7 values at constraint application, keeping the default behavior unchanged.

olivermt:
Fixes #2794.

Adds an opt-in `strict?` constraint (default `false`) to `Ash.Type.UUIDv7`. When set, non-v7 uuids are rejected at constraint application with `must be a version 7 UUID`, so callers get an input validation error instead of a load failure after the row is written.

The default preserves current behavior, including the `match_v4_uuids?` compile config. When both are set, `strict?` wins: the raw v4 binary casts, then fails the constraint.

Usage:

```elixir
argument :id, :uuid_v7, constraints: [strict?: true]
```

Whether strictness should eventually become the default (the coherent semantics for a type named UUIDv7, but breaking for anyone currently feeding other versions through it) is left to #2794.

### DIFF
--- a/lib/ash/type/uuid_v7.ex
+++ b/lib/ash/type/uuid_v7.ex
@@ -3,10 +3,24 @@
 # SPDX-License-Identifier: MIT
 
 defmodule Ash.Type.UUIDv7 do
+  @constraints [
+    strict?: [
+      type: :boolean,
+      default: false,
+      doc: """
+      Requires the value to actually be a version 7 UUID. Recommended for externally-supplied values: without it, any well-formed UUID passes input casting, but non-v7 values cannot be loaded back from storage.
+      """
+    ]
+  ]
+
   @moduledoc """
   Represents a UUID.
 
   A builtin type that can be referenced via `:uuid_v7`
+
+  ### Constraints
+
+  #{Spark.Options.docs(@constraints)}
   """
 
   use Ash.Type,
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2799, opened 2026-07-23
feat: support before_transaction? option on validations

barnabasJ:
## What

Adds a `before_transaction?: true` option to validations (both action-level `validate` and global `validations`), as a counterpart to `before_action?`, for validations that must run outside the transaction — e.g. validations that call external services.

```elixir
create :create do
  validate {MyValidation, []}, before_transaction?: true
end
```

## Semantics

- The validation runs in a `before_transaction` hook, **prepended** so it runs before any other `before_transaction` hooks, and therefore outside the transaction.
- Takes precedence over `before_action?` when both are set.
- `only_when_valid?` is re-checked at hook time rather than at build time.
- If the changeset/query/input is already invalid before hooks run, `before_transaction` hooks are never reached, so the validation does not run and the original error surfaces (matching existing hook behavior).
- A failing validation makes the subject invalid, which short-circuits any later `before_transaction` hooks.
- Supported on changesets (create/update/destroy), queries (read actions), and action inputs (generic actions).
- Global validations keep running in the `before_transaction` hook even when the action sets `delay_global_validations? true` (`before_transaction?` wins over the `before_action?` rewrite).

## Bulk actions

`Ash.bulk_create`, `Ash.bulk_update`, and `Ash.bulk_destroy` (including soft destroys) run these validations in the `before_transaction` hook of each changeset, before other hooks. Previously the bulk create dispatch ran them silently at build time. The `before_transaction?` dispatch also takes precedence over the `batch_validate/3` fast path — batch-capable validations are invoked per changeset at hook time.

## Atomicity

`before_transaction?` validations can never run atomically:

- `run_atomic_validation/3` returns `{:not_atomic, ...}` with a message telling users to set `require_atomic? false` or use `strategy: [:stream]`.
- Actions with `require_atomic? true` raise `Ash.Error.Framework` at runtime.
- Bulk actions with `strategy: [:atomic]` return a `NoMatchingBulkStrategy` error carrying that reason; when `:stream` is allowed they fall back to streaming.
- The `VerifyActionsAtomic` verifier now warns at compile time when a `require_atomic?` action has a `before_transaction?` validation, even if the validation module is otherwise atomic-capable.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -58,6 +58,7 @@ spark_locals_without_parens = [
   base_filter: 1,
   batch_size: 1,
   before_action?: 1,
+  before_transaction?: 1,
   belongs_to: 2,
   belongs_to: 3,
   broadcast_type: 1,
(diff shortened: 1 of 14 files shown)


barnabasJ:
closing for now, claude was to eager, I still need to review myself first, will reopen, once it's ready.

PR #2800, opened 2026-07-23
feat: add all/1 builtin validation as the complement to any/1

barnabasJ:
## What

Adds `Ash.Resource.Validation.All` and the corresponding `all/1` builtin, which passes only when every provided validation passes.

```elixir
validate all([
  present(:foo),
  absent(:bar)
])
```

## Why

Together with `any/1` (OR) and `negate/1` (NOT), this completes the validation combinators. It enables compositions like `any([all([...]), ...])` that were previously inexpressible, since AND only existed at the DSL level via multiple `validate` statements or `where` clauses.

## Implementation

Implemented as the dual of `Ash.Resource.Validation.Any`:

- `validate/3` fails with a combined `"must pass all of: ..."` error when any constituent fails.
- `atomic/3` combines the constituent error conditions with `or` (where `Any` combines with `and`), drops `:ok` constituents, and short-circuits on static `{:error, ...}` results.
- Returns `{:not_atomic, ...}` if any constituent cannot be run atomically.

### DIFF
--- a/lib/ash/resource/validation/all.ex
+++ b/lib/ash/resource/validation/all.ex
@@ -1,0 +1,164 @@
+# SPDX-FileCopyrightText: 2019 ash contributors <https://github.com/ash-project/ash/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Ash.Resource.Validation.All do
+  @moduledoc false
+
+  @opt_schema [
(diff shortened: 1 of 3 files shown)


zachdaniel (approved):

zachdaniel:

*sick*

PR #2802, opened 2026-07-24
improvement: avoid quadratic cost evaluating in-list filters at runtime

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

# Summary
This PR is to address a performance issue found with Ash.Filter.Runtime filtering. The improvement here is to avoid quadratic cost evaluating in-list filters at runtime

The first part of the PR is 'test only', and is really to describe existing behaviour - the Ash equality performs a semantic compare rather than a structural one.

The second part of the PR is to refactor to remove a quadratic performance issue in the existing code based on the convention.

# Evidence

Measured on Ash `main`, Elixir 1.20.0 / OTP 29. Each run filters N records against a list of the N ids those records hold — the shape a relationship load takes, since Ash frames `load: :subject` as `source_attribute in [N parent ids]`.

## The cost

| records | before | after |
| ---: | ---: | ---: |
| 500 | 79.9 ms | 0.4 ms |
| 1,000 | 380.0 ms | 0.7 ms |
| 2,000 | 1,417.7 ms | 1.5 ms |
| 4,000 | 5,551.7 ms | 3.2 ms |
| 8,000 | 21,653.6 ms | 5.9 ms |

Before: per-record cost doubles as N doubles, totals quadruple. After: flat at ~0.8 µs per record.

## Why quadratic

- `Ash.Filter.Runtime` evaluates the filter once per record.
- A relationship load frames as `source_attribute in [N parent ids]` over the
  N records it selects → N records × N members.

### DIFF
--- a/lib/ash/query/operator/in.ex
+++ b/lib/ash/query/operator/in.ex
@@ -30,6 +30,15 @@ defmodule Ash.Query.Operator.In do
   def evaluate(%{left: nil}), do: {:known, nil}
   def evaluate(%{right: nil}), do: {:known, nil}
 
+  def evaluate(%{left: left, right: %MapSet{} = right}) do
+    # `Comp.equal?/2` is semantic rather than structural equality — `1` equals
+    # `1.0`, a `Decimal` equals the integer and the string spelling the same
+    # number, an `Ash.CiString` equals a binary differing only in case — so a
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2807, opened 2026-07-24
feat: add mfa constraint to Ash.Type.Function for portable persistence

### LINKED ISSUE #2806: Enhance Ash.Type.Function portability with an mfa: true constraint
### Code of Conduct

### AI Policy

This proposal is about enhancing both the usefulness and portability of `Ash.Type.Function`, when persisted by datalayer.  

Currently `Ash.Type.Function` dumps either a `:local` or `:external` function to `:binary`, which carries no guarantee of binary compatibility across OTP versions.

The proposal allows `Ash.Type.Function` to have a non-default constraint: `mfa: true` 
- ensuring that the function is of type `:external` (represented in mfa form)
- changing the storage type to `:string`

This form has its own caveats: no guarantee the MFA exists in the reading environment, or that a same-arity function does the same thing. Even so, I've found this reference form more useful than the binary.

The constraint is opt-in and defaults off, so existing :function attributes are unchanged — same :binary storage, closures still accepted.

We contemplated `Ash.Type.Mfa`, but as MFA is a subset of `Ash.Type.Function` this creates more problems that it solves.

We contemplated the `external: true` constraint, which mirrors the `:external` value that `Function.info/1` reports for a function's type (`:external` vs `:local`), however we rejected this as it doesn't imply the change in default storage type the same way that `mfa: true` does.

### Additional context

_No response_

matt-beanland:
closes #2806

# Contributor checklist

- [*] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Chores
- [ ] Documentation changes
- [*] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

# Summary

This PR adds a `mfa: true` constraint to 'Ash.Type.Function' for portable persistence.
There are three commits in this PR:
- the `mfa` constraint refusing closures at cast (validation only, storage untouched)
- constrained functions are persisted as portable capture strings instead of VM-fragile binaries
- tests documenting late binding

This PR provides a persistence option for `Ash.Type.Function` that is safer than the default binary form. The default dumps via `:erlang.binary_to_term/2` - a deserialization surface the moduledoc flags as "NOT SAFE to use with external input".

The `mfa: true` form needs no binary deserialization: the stored value is a readable capture string, parsed with `String.to_existing_atom/1` (never minting atoms) and resolved with `Function.capture/3`.

That said it does need care: the referenced `Module.function/arity` may not be present in the reading runtime. Reconstruction needs only the module atom, so the record still loads; invocation is late-bound and raises `UndefinedFunctionError` if the code is genuinely absent. That trade-off is documented and tested.

The constraint is opt-in and defaults off, so existing `:function` attributes are unchanged - same `:binary` storage, closures still accepted.

### DIFF
--- a/lib/ash/type/function.ex
+++ b/lib/ash/type/function.ex
@@ -4,6 +4,15 @@
 
 defmodule Ash.Type.Function do
   @constraints [
+    mfa: [
+      type: :boolean,
+      default: false,
+      doc: """
+      Restricts the value to external captures (`&Module.function/arity`), refusing \
+      anonymous functions and closures. Only external captures can be stored: a closure \
+      captures its environment, which cannot be rehydrated in another VM.
+      """
+    ],
     arity: [
       type: :pos_integer,
       doc: "Enforces a specific arity on the provided function"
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2809, opened 2026-07-24
fix: order durations semantically instead of by struct field layout

### LINKED ISSUE #2808: Ash.Type.Duration comparisons return semantically wrong results
### Code of Conduct

### AI Policy

### Versions

Elixir | 1.20.0 (compiled with Erlang/OTP 27)
Erlang/OTP | 29 (erts-17.0.3)
Ash | main @ 90baa9fc0 (mix.exs 3.30.1)

### Operating system

### Current Behavior

`Ash.Type.Duration`  comparisons return semantically wrong results. Ash.Type.Duration currently has no comparator (which was a decision on #2036 ) which means pairs fall through to a default that orders the `%Duration{}` struct by its fields, rather than elapsed time. For example `PT25H > P1D` incorrectly returns false, `P1W = P7D` incorrectly returns false, `PT60M == PT1H` incorrectly returns false. Not only that but `P1Y6M == P18M` returns false (should be true — both are 18 months)

Filters and sorts on `:duration` attributes return or order the wrong records.

### Reproduction

The following should be executed in iex:

```elixir
Mix.install([{:ash, "~> 3.30"}])

alias Ash.Query.Operator.{GreaterThan, Eq}

# 25 hours is longer than 1 day
GreaterThan.evaluate(%{left: Duration.new!(hour: 25), right: Duration.new!(day: 1)})
#=> {:known, false}   # should be true

# 1 week is 7 days
Eq.evaluate(%{left: Duration.new!(week: 1), right: Duration.new!(day: 7)})
#=> {:known, false}   # should be true

# 60 minutes is 1 hour
Eq.evaluate(%{left: Duration.new!(minute: 60), right: Duration.new!(hour: 1)})
#=> {:known, false}   # should be true

```

### Expected Behavior

Durations should compare by magnitude within each of the two regimes that are internally, totally ordered:

day-time — week/day/hour/minute/second/microsecond (e.g. `PT25H > P1D`, `P1W == P7D`)
year-month — year/month, a year being 12 months (e.g. `P2Y > P1Y`, `P1Y6M == P18M`)
A comparison should only fail loudly when it crosses the regimes (`P1M vs P30D` — a month is not a fixed number of days and there is no reference instant) or when a single duration mixes units from both, since no context-free order exists there.

matt-beanland:
Closes #2808

# Contributor checklist

- [*] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [*] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

This PR orders durations semantically instead of by struct field layout.

Duration values reaching Comp fell through to the `Any/Map` default comparator, which orders `%Duration{}` structs by their fields in map key-order — a comparison with no temporal meaning. On released Ash this silently returns wrong answers: `P1W == P7D` is false, `PT25H > P1D` is false, `P1Y6M == P18M` is false.

Added a `Duration :: Duration` comparator built on the two regimes that `Duration` is internally ordered within, mirroring xs:yearMonthDuration and xs:dayTimeDuration in XML Schema. Day-time durations (week/day/hour/minute/second/microsecond) are ordered as microseconds — week = 7 days, day = 24 hours — at full precision. This is the same boundary to_timeout/1 draws, without its millisecond truncation, so sub-millisecond durations stay distinct. Year-month durations (year/month) are ordered as months, a year being 12 months.

A duration is only compared within its own regime; a wholly-zero duration belongs to either. Comparing across the regimes (`P1M` versus `P30D` has no context-free answer — a month is not a fixed number of days), or comparing a duration that mixes units from both, raises `Ash.Error.Query.InvalidExpression` rather than inventing an order, noting that `Comparable` can only return an ordering or raise.

### DIFF
--- a/lib/ash/type/duration.ex
+++ b/lib/ash/type/duration.ex
@@ -144,4 +144,70 @@ defmodule Ash.Type.Duration do
   def dump_to_native(value, _) do
     Ecto.Type.dump(:duration, value)
   end
+
+  @usec_per_second 1_000_000
+  @seconds_per_minute 60
+  @minutes_per_hour 60
+  @hours_per_day 24
+  @days_per_week 7
+  @days_per_month 30
+  @months_per_year 12
+
+  @doc """
+  Compares two durations as a total order, matching how the AshPostgres data
+  layer (PostgreSQL `interval`) compares them: a fixed conversion of `month` → 30
+  days and `day` → 24 hours (so `year` → 360 days, `week` → 7 days), down to
+  microseconds.
+
+  `Duration` is only *partially* ordered in general — a month is not a fixed
+  number of days — which is why Elixir ships `Duration` without a `compare/2`, and
+  why data layers disagree on cross-unit comparison: PostgreSQL uses 30-day
+  months, Neo4j ~30.44-day months, and Elixir's `to_timeout/1` refuses `month`/
+  `year` outright. This adopts PostgreSQL's convention so in-memory comparison
+  stays aligned with the dominant data layer rather than raising or drifting.
+  Within the day/time units, or within the year/month units, the result is exact
+  and portable across those backends; only comparison *across* that boundary
+  depends on the 30-day convention.
+
+  Computed from the integer fields directly, so microsecond precision is kept
+  (unlike `to_timeout/1`, which truncates to milliseconds).
+
+  This function is the single place the convention lives; if Elixir core later
+  gains a `Duration.compare/2`, it can delegate here.
+  """
+  @spec compare(Duration.t(), Duration.t()) :: :lt | :eq | :gt
+  def compare(%Duration{} = left, %Duration{} = right) do
+    compare_ints(total_microseconds(left), total_microseconds(right))
+  end
+
+  defp total_microseconds(%Duration{
+         year: year,
+         month: month,
+         week: week,
+         day: day,
+         hour: hour,
+         minute: minute,
+         second: second,
+         microsecond: {microsecond, _precision}
+       }) do
+    days = (year * @months_per_year + month) * @days_per_month + week * @days_per_week + day
+    hours = days * @hours_per_day + hour
+    minutes = hours * @minutes_per_hour + minute
+    seconds = minutes * @seconds_per_minute + second
+    seconds * @usec_per_second + microsecond
+  end
+
+  defp compare_ints(left, right) do
+    cond do
+      left > right -> :gt
+      left < right -> :lt
+      true -> :eq
+    end
+  end
+end
+
+import Ash.Type.Comparable
+
+defcomparable left :: Duration, right :: Duration do
+  Ash.Type.Duration.compare(left, right)
 end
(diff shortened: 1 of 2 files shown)


zachdaniel:
Do you know how durations compare inside of postgres for example? In postgreSQL I think you just get "unexpected results". The main reason I'm asking is that generally we try to retain semantics across data layers and core where possible. The way that this raises errors on certain comparable but postgres for example wouldn't would be drift that could be hard to internalize.

matt-beanland:
Zach, I've looked into this for a couple of data layers recognising that AshPostgres dominates our community. The interesting result is they contradict each other, so there's no single cross-layer semantics to retain:

# Postgres

total order, never errors, month→30d / day→24h

```sql
SELECT interval '1 month' = interval '30 days'      AS p1m_eq_p30d,    -- t
       interval '1 year'  = interval '360 days'     AS p1y_eq_p360d,   -- t
       interval '1 mon 15 days' < interval '2 mons' AS mixed_45d_60d;  -- t
```

# Neo4j
 comparison operators return null; 
 ORDER BY approximates (year = 365.2425d, month = 30.436875d = ¹⁄₁₂ year, day = 24h).
 interestingly storage is months, days, seconds, nanoseconds as integers
 
 # Elixir
 `Duration` lacks `compare/2`
`to_timeout/1` and XML Schema (yearMonthDuration/dayTimeDuration) — refuse across the boundary.
`to_timeout/1` also is lossy as rounds microseconds to milliseconds
 
The month-day ratio is where they diverge (30 vs 30.436875 vs undefined); other data layers will differ again. But within a regime they all agree — day = 24h, year = 12 months — so that part of this PR is portable everywhere; the divergence is purely cross-regime.

Given that, I'm happy to align core's cross-regime behaviour with Postgres's convention (month = 30d, day = 24h). It's also strictly better than the current misalignment and the raise you flagged on this PR as it stands.  Comparable.compare can only order-or-raise and Ash.Filter.Runtime doesn't rescue it, so a cross-regime comparison currently crashes Ash.read where every data layer returns something.

Repro of the core side (from the branch), showing within-regime agreement with PG and the current cross-regime raise this change replaces:

```elixir
    Mix.install([{:ash, github: "matt-beanland/ash", ref: "fix/duration-ordering"}])
    alias Ash.Type.Duration, as: D
    D.compare(Duration.new!(year: 2), Duration.new!(year: 1))             #=> :gt
    D.compare(Duration.new!(year: 1, month: 6), Duration.new!(month: 18)) #=> :eq
    D.compare(Duration.new!(hour: 25), Duration.new!(day: 1))             #=> :gt
    D.compare(Duration.new!(month: 1), Duration.new!(day: 30))            #=> raises → will become :eq (PG convention)
```
The units constraint (#2811) stays as the opt-in to refuse cross-regime values at cast, for anyone who wants stricter semantics than any single backend guarantees.

zachdaniel:
Nice, aligning to postgres is a reasonable default 👍

zachdaniel:

PR #2811, opened 2026-07-24
feat: add units constraint to Ash.Type.Duration

### LINKED ISSUE #2810: Enhance Ash.Type.Duration utility and safety with a units constraint
### Code of Conduct

### AI Policy

The utility of `Ash.Type.Duration` is improved if it can be constrained to just carry particular units, or combinations of units.

When #2808 is addressed by #2809 `units` can be used to ensure that comparisons involving any Duration with non-zero year-month against non-zero day-time components are avoided.

Add an opt-in constraint on `Ash.Type.Duration` to whitelist units allowed to be non-zero. Any unit outside the list must be zero, otherwise the cast fails.

Add an opt-in units constraint on `Ash.Type.Duration` that whitelists the units allowed to be non-zero. Any unit outside the list must be zero, otherwise the cast fails.

units accepts either a shorthand atom or an explicit list:

```elixir
constraints: [units: :day_time]        # == [:week, :day, :hour, :minute, :second, :microsecond]
constraints: [units: :year_month]   # == [:year, :month]
constraints: [units: [:hour, :minute]]  # == an arbitrary list
```

The two atoms are convenience shorthands for the unit groups either side of the comparability boundary. They are only shorthands: comparison is valid for any durations that stay on one side of the boundary, so a narrower whitelist like `[:hour, :minute]` is equally comparable.

Elixir's `Duration` lacks compare. We rejected proposing one as it would lack the ability to be constraint driven.

We considered treating years and months as non-comparable, but they are comparable when not mixed with units from the other side of the boundary.

### Additional context

_No response_

matt-beanland:
closes #2810

Duration attributes accepted any duration, with no way to declare which units a value may use. Add a `units` constraint — a whitelist of the Duration fields permitted to be non-zero; any unit outside the list must be zero, otherwise casting fails.

# Contributor checklist

- [* ] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Chores
- [ ] Documentation changes
- [*] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

# Summary

This PR adds units constraint to `Ash.Type.Duration`

`units` accepts an explicit list, or a shorthand for one side of the comparability boundary - `:day_time` (week/day/hour/minute/second/ microsecond) or `:year_month` (year/month):

    attribute :timeout, :duration, constraints: [units: :day_time]
    attribute :term,    :duration, constraints: [units: :year_month]
    attribute :offset,  :duration, constraints: [units: [:hour, :minute]]

This is cast-time validation - data hygiene, keeping unintended units out of storage. Confining an attribute to a single side of the boundary also keeps its values comparable (see `Ash.Type.Duration.compare/2`).

A units whitelist is a property of the decoded `Duration` struct's fields and cannot be checked inside an atomic expression, so `may_support_atomic_update?/1` returns false when the constraint is set, routing the value through the non-atomic path where it is enforced.

While this PR is related to issue #2808 and its PR #2809 it does not *require* them.

### DIFF
--- a/lib/ash/type/duration.ex
+++ b/lib/ash/type/duration.ex
@@ -3,12 +3,32 @@
 # SPDX-License-Identifier: MIT
 
 defmodule Ash.Type.Duration do
+  @year_month_units [:year, :month]
+  @day_time_units [:week, :day, :hour, :minute, :second, :microsecond]
+  @duration_units @year_month_units ++ @day_time_units
+
+  @constraints [
+    units: [
+      type: {:or, [{:in, [:year_month, :day_time]}, {:list, {:in, @duration_units}}]},
+      doc: """
+      The units permitted to be non-zero; any unit outside the set must be zero, otherwise casting fails. Either an explicit list of units, or a shorthand for one side of the comparability boundary: `:year_month` (`[:year, :month]`) or `:day_time` (`[:week, :day, :hour, :minute, :second, :microsecond]`). Confining an attribute to a single side keeps its values comparable (see `Ash.Type.Duration.compare/2`).
+      """
+    ]
+  ]
+
   @moduledoc """
   Represents a Duration
 
   A builtin type that can be referenced via `:duration`
+
+  ### Constraints
+
+  #{Spark.Options.docs(@constraints)}
   """
   use Ash.Type
+
+  @impl true
+  def constraints, do: @constraints
 
   @impl true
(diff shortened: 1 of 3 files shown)


zachdaniel:

PR #2812, opened 2026-07-31
fix: allow_forbidden_field_for_relationships_by_default? missing the ?

frankdugan3:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

The key for the config did not match the documentation/generators.

### DIFF
--- a/lib/ash/resource/relationships/shared_options.ex
+++ b/lib/ash/resource/relationships/shared_options.ex
@@ -127,7 +127,12 @@ defmodule Ash.Resource.Relationships.SharedOptions do
     allow_forbidden_field?: [
       type: :boolean,
       default:
-        Application.compile_env(:ash, :allow_forbidden_field_for_relationships_by_default, false),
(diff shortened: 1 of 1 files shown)


lib/ash/resource/relationships/shared_options.ex
```diff
@@ -127,7 +127,7 @@ defmodule Ash.Resource.Relationships.SharedOptions do
     allow_forbidden_field?: [
       type: :boolean,
       default:
-        Application.compile_env(:ash, :allow_forbidden_field_for_relationships_by_default, false),
+        Application.compile_env(:ash, :allow_forbidden_field_for_relationships_by_default?, false),
```
zachdaniel: We're going to have to check for both now to avoid a breaking change 😢

frankdugan3: I added both variants via an `||`. I don't mind any nits if you want to do it a different way.

zachdaniel (approved):

PR #2818, opened 2026-08-02
improvement: index runtime_sort rekey by primary key instead of scanning

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

# Summary

maybe_rekey/4 mapped each of the n sorted records back onto the input with Enum.find. As the sorted list is a permutation of the input, Enum.find lands on each position exactly once, so the cost is exactly triangular, n(n+1)/2, regardless of sort order.

Performance has been made linear for the common simple-equality primary key case, where matching is plain term equality over Map.take/2, by first building a key -> record index. Primary keys using semantic equality (Ash.Type.equal?/3) keep the existing scan.

The first part of the PR is 'test only'.

The second part refactors maybe_rekey/4 to the indexed form, preserving both of those behaviours.

# Evidence

Measured on Ash `main`, Elixir 1.20.0 / OTP 27. Each run rekeys a shuffled permutation of N records back onto the input — the shape maybe_rekey/4 sees after every runtime sort — with the before/after outputs asserted equal.

## The cost

| records | before | after |
| ---: | ---: | ---: |
| 500 | 6.4 ms | 0.1 ms |
| 1,000 | 25.5 ms | 0.3 ms |
| 2,000 | 102.9 ms | 0.6 ms |
| 4,000 | 426.6 ms | 1.2 ms |
| 8,000 | 1,679.5 ms | 2.6 ms |

Before: cost quadruples as N doubles. After: linear.

### DIFF
--- a/lib/ash/actions/sort.ex
+++ b/lib/ash/actions/sort.ex
@@ -148,14 +148,41 @@ defmodule Ash.Actions.Sort do
   end
 
   defp maybe_rekey(new_results, results, resource, true) do
-    if Ash.Resource.Info.primary_key(resource) == [] do
-      new_results
-    else
-      Enum.map(new_results, fn new_result ->
-        Enum.find(results, new_result, fn result ->
-          resource.primary_key_matches?(new_result, result)
+    pkey = Ash.Resource.Info.primary_key(resource)
+
+    cond do
+      pkey == [] ->
+        new_results
+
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:

PR #2819, opened 2026-08-02
improvement: memoize Comparable's type-pair dispatch in a release

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

# Summary

`Comp.new/2` rebuilds the dispatch module on every comparison. For any pair with no specific comparator (the common case) that is a raised and caught `ArgumentError`. This memoises the resolution, a pure function of the two type atoms, in the process dictionary. No semantics change.

It shows up in runtime `in`-filters. A `MapSet` miss falls back to `Enum.any?(set, &Comp.equal?(&1, left))`, so every non-matching record scans the set and each element pays that dispatch twice over. The change is to the vendored `Comp`. The O(n) miss scan is left as is.

# Evidence

`In.evaluate/1` over a miss (binary left, `MapSet` of binaries) is the per-record shape of a runtime `in`-filter. Elixir 1.20 / OTP 27.

| build | before | after | speedup |
| --- | ---: | ---: | ---: |
| dev (unconsolidated protocols) | 289.6 µs | 102.0 µs | 2.8x |
| prod (consolidated protocols) | 266.0 µs | 24.0 µs | ~11x |

`Module.safe_concat/1` for one 421-element miss: 842 to 0.

### DIFF
--- a/lib/comparable/comp.ex
+++ b/lib/comparable/comp.ex
@@ -316,17 +316,97 @@ if !Code.ensure_loaded?(Comp) do
     end
 
     defp new(left, right) do
-      lr_type =
-        try do
-          [Comparable, Type, type_of(left), To, type_of(right)]
-          |> Module.safe_concat()
-          |> fall_back_unless_implemented()
-        rescue
-          ArgumentError ->
-            any_to_any()
-        end
+      %{__struct__: dispatch(type_of(left), type_of(right)), left: left, right: right}
+    end
 
-      %{__struct__: lr_type, left: left, right: right}
+    # Elixir protocols dispatch on their first argument only. To dispatch on
+    # both operand types, `Comparable` encodes the type pair into a synthetic
+    # struct (`Comparable.Type.<L>.To.<R>`) and dispatches the protocol on that.
+    # Resolving a pair to that module is a pure function of the two type atoms,
+    # so memoize it rather than rebuilding the module name — and catching the
+    # `ArgumentError` for every pair with no specific comparator — on every
+    # single comparison.
+    #
+    # Whether it is cached at all depends on whether code can still change.
+    # Comparators only ever appear, so the one way a cached answer can go stale
+    # is a pair resolving to the `Any.To.Any` fallback before its own
+    # comparator is implemented.
+    #
+    #   * `:embedded` — a release. Every module is loaded at boot and none
+    #     arrives later, so the resolution is fixed for the life of the VM and
+    #     `:persistent_term` can hold it once for every process.
+    #
+    #   * `:interactive` — `mix`, `iex`, tests. Code is loaded lazily and
+    #     recompiled in place, so a comparator's atom can appear after a
+    #     fallback was cached. Nothing is cached, which is what this did before.
+    #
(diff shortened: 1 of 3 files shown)


lib/comparable/comp.ex
```diff
@@ -316,17 +316,36 @@ if !Code.ensure_loaded?(Comp) do
...
+    # struct (`Comparable.Type.<L>.To.<R>`) and dispatches the protocol on that.
+    # Resolving a pair to that module is a pure function of the two type atoms,
+    # so memoise it in the process dictionary rather than rebuilding the module
+    # name — and catching the `ArgumentError` for every pair with no specific
+    # comparator — on every single comparison. Per-process memoisation always
+    # reflects the currently loaded comparators and keeps no global state.
+    defp dispatch(left_type, right_type) do
+      key = {__MODULE__, :dispatch, left_type, right_type}
+
+      case Process.get(key) do
```
zachdaniel: Maybe we should use persistent term here actually to memoizenit forever?

matt-beanland: That was my first thought too. I'm trying to avoid subtle bugs if/when comparators are changed during development or runtime synthesis. At least a process is restarted in development and can be easily restarted on hot code swap, whereas persistent_term lodges these in the VM where they will survive tests, etc.  I've not had lots of experience with either approach so I'm happy to be guided by your decision.

With persistent_term I briefly considered using :erase in defcomparable, but it is a convenience macro so no guarantee that it would be used, another footgun.

zachdaniel: Hmm...thats a fair point, but I feel like we could work around it. We could use a sentinel like `build_embedded` or something like that to not do this in development environments 🤔

zachdaniel:
I think there may be a simpler way to do this. I think what we can do is use something like `@ref make_ref()` at the top, and use that as the persistent_term key. Then if the module is ever recompiled, that `@ref` will change. It will be recompiled on the target types being recompiled since the protocol has a compile time dependency on them. Not sure we need a toggle for turning this off either?

matt-beanland:
I was drafting an detailed explanation, but you've already read the code.

In my view we should provide an option to control the memoization irrespective of the environment. So can memoize or not (benchmarks, etc), and can reset if necessary at runtime rather than restart a VM.

I don't think we can rely on a Comparator implementing anything other than the protocol.

matt-beanland:
I've tried the `@ref` idea and it compiles if you just use an integer token.
But with`@ref` in the Type, `Comp` won't be recompiled with it unless in the same source file. A protocol implementation is a module of its own.

And being compile-time, it wouldn't reach the case I'm most concerned about anyway: a comparator
synthesised into a running release, where nothing is recompiled at all.

So I think the two functions on Comp are necessary.

matt-beanland:
#2840 should be reviewed/merged first, it deals with a fairly unlikely hazard regarding Comp.compare dispatch which is compounded with memoization.

zachdaniel:
will merge once conflicts are resolved 👍

zachdaniel:

PR #2824, opened 2026-08-03
fix: order ETS min/max aggregates with Comp instead of Erlang term order

### LINKED ISSUE #2823: ETS min and max aggregates wrong for every struct-valued type except Decimal
### Code of Conduct

### AI Policy

### Versions

| | version |
| --- | --- |
| Ash | 3.31.0 (hex) - also reproduced on main @ `ef5998486` (mix.exs 3.31.0) |
| AshPostgres | 2.11.0 (ash_sql 0.6.6) |
| PostgreSQL | 17.10 |

### Operating system

### Current Behavior

`Ash.DataLayer.Ets` computes `min` and `max` aggregates with Erlang term order, so for
struct-valued types they are wrong. For two values they are not merely inaccurate, they are
**swapped** - `max` of a datetime field returns the earliest row.

`Decimal` is special-cased and is correct. Every other struct type is affected — measured with
two values per type, chosen so that term order and the type's own order disagree:

| attribute type | `min` / `max` |
| --- | --- |
| `:utc_datetime_usec` | inverted |
| `:naive_datetime` | inverted |
| `:date` | inverted |
| `:time_usec` | inverted |
| `:duration` | inverted |
| `:ci_string` | inverted |
| `:decimal` | correct — special-cased |
| `:integer` (control) | correct |

The same field, sorted by the same data layer in the same read, sorts correctly. Only the
aggregate disagrees.

### Reproduction

Paste into `iex`:

```elixir

defmodule Domain do
end

defmodule Row do

  attributes do
    uuid_primary_key :id
    attribute :at, :utc_datetime_usec, public?: true
    attribute :n, :integer, public?: true
  end

  actions do
    defaults [:read, create: :*]
  end
end

require Ash.Query

# Chronology and Erlang term order point opposite ways here: a DateTime's keys
# compare as text, so :day (31 vs 1) decides before :year is ever reached.
early = ~U[2020-12-31 23:00:00.000000Z]
late = ~U[2026-01-01 00:00:00.000000Z]

Ash.create!(Row, %{at: early, n: 1})
Ash.create!(Row, %{at: late, n: 2})

agg = fn kind, field ->
  Row |> Ash.Query.new() |> Ash.aggregate!({:v, kind, field: field}) |> Map.get(:v)
end

agg.(:max, :at)
#=> ~U[2020-12-31 23:00:00.000000Z]    expected ~U[2026-01-01 00:00:00.000000Z]

agg.(:min, :at)
#=> ~U[2026-01-01 00:00:00.000000Z]    expected ~U[2020-12-31 23:00:00.000000Z]

agg.(:max, :n)
#=> 2                                  integer control, correct

# the same field, sorted by the same data layer in the same read:
Row |> Ash.Query.sort(at: :desc) |> Ash.read!() |> Enum.map(& &1.n)
#=> [2, 1]                             correct

# and the comparator that sort already uses gives the right answer:
Enum.max([early, late], Comp)
#=> ~U[2026-01-01 00:00:00.000000Z]    correct
```

The last two lines are the argument in miniature: the same data layer, given the same field,
orders it correctly.

### Expected Behavior

`max(:at)` should return `~U[2026-01-01 00:00:00.000000Z]` and `min(:at)` should return `~U[2020-12-31 23:00:00.000000Z]`.

`AshPostgres.DataLayer` returns the chronologically correct extreme for the same model and the
same `Ash.aggregate!` call - for a `:date` field, `max` → `~D[2026-01-01]` and `min` →
`~D[2020-12-31]`. 

`Ash.DataLayer.Ets` is divergent. Both routes into `aggregate_value/6` are affected — a direct `Ash.aggregate!`, and a declared `min`/`max` aggregate over a relationship.

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Fixes #2823.

`min` and `max` in `Ash.DataLayer.Ets` compared with `Enum.min/1` and `Enum.max/1`, which fall
back to Erlang term order. For struct-valued types that is wrong, and for two values the answers
are not merely inaccurate but **swapped**. The `max` of a datetime field returned the earliest record.

`Decimal` was already special-cased, which is why it worked.
Measured before and after, with two values per type chosen so term order and the type's own order
disagree:

| attribute type | before | after |
| --- | --- | --- |
| `:utc_datetime_usec` | inverted | correct |
| `:naive_datetime` | inverted | correct |
| `:date` | inverted | correct |
| `:time_usec` | inverted | correct |
| `:duration` | inverted | correct |
| `:ci_string` | inverted | correct |
| `:decimal` | correct — special-cased | correct |
| `:integer` (control) | correct | correct |

AshPostgres answers these correctly, so the two in-tree data layers disagreed. Repro in the
issue.

# Cause

`aggregate_value/6` in `lib/ash/data_layer/ets/ets.ex`:

```elixir
:max ->
  if is_struct(first_item, Decimal) do
    Enum.reduce(items, &Decimal.max(&1, &2))
  else
    Enum.max(items)          # ← Erlang term order
  end
```

Erlang orders maps by size, then keys, then values **in key order** — and the keys are atoms,
which sort lexicographically. So a struct's fields are compared in alphabetical name order:

```elixir
~U[2020-01-01 00:00:00Z] |> Map.keys() |> Enum.sort()
#=> [:__struct__, :calendar, :day, :hour, :microsecond, :minute, :month,
#=>  :second, :std_offset, :time_zone, :utc_offset, :year, :zone_abbr]
```

`:day` decides and `:year` is never reached, so `2020-12-31` sorts above `2026-01-01`. The same
shape explains `time_usec`: `:microsecond` outranks `:minute` and `:second`.

Reordering fields would not be a fix either.  For `DateTime`, `Comp` does not read struct fields at all — it
delegates to the stdlib comparator:

```elixir
defcomparable left :: DateTime, right :: DateTime do
  DateTime.compare(left, right)
end
```

and `DateTime.compare/2` applies `utc_offset`/`std_offset`, so it is instant-based. Two rows with
different wall-clock fields in different zones are the same instant and compare `:eq`, which no
ordering of the stored fields can reproduce.

This is the one place in the file that does not use `Comp`. `distinct_and_sort/6` routes to
`Ash.Actions.Sort.runtime_sort/3`, which compares with `Comp`, so *sorting* the same field in the
same read already gives the chronological answer. Comparing with `Comp` here makes `min`/`max`
agree with it:

```elixir
:max -> Enum.max(items, Comp)
:min -> Enum.min(items, Comp)
```

`Comp` is vendored at `lib/comparable/`, so this adds no dependency. It handles `Decimal`, so the
special case goes with it, and `nil`s are already rejected before this point.

Both routes into `aggregate_value/6` are covered by the tests: `run_aggregate_query/3` (a direct
`Ash.aggregate!`) and `do_add_aggregates/4` (a declared aggregate over a relationship). Both fail
without the change.

# Commits

1. `test:` pin that a datetime field sorts and `first`s chronologically. These pass as is. That is
   the contrast which gives `min`/`max` their expected answer, and nothing pinned it.
2. `fix:` compare with `Comp`, plus the `min`/`max` cases for both routes.

### DIFF
--- a/lib/ash/data_layer/ets/ets.ex
+++ b/lib/ash/data_layer/ets/ets.ex
@@ -1043,19 +1043,19 @@ defmodule Ash.DataLayer.Ets do
                   Enum.sum(items)
                 end
 
+              # `Comp` is what `Ash.Actions.Sort.runtime_sort/3` orders with, so
+              # comparing with it here keeps `min`/`max` consistent with how this
+              # data layer sorts the very same field. `Enum.max/1` would fall
+              # back to Erlang term order, which compares a struct's fields in
+              # alphabetical key order — for a `DateTime` that means `:day`
+              # decides and `:year` is never reached. No field ordering would be
+              # right anyway: `DateTime.compare/2` applies the UTC offsets, so
+              # two rows with different wall clocks can be the same instant.
               :max ->
-                if is_struct(first_item, Decimal) do
-                  Enum.reduce(items, &Decimal.max(&1, &2))
-                else
-                  Enum.max(items)
-                end
+                Enum.max(items, Comp)
 
               :min ->
-                if is_struct(first_item, Decimal) do
-                  Enum.reduce(items, &Decimal.min(&1, &2))
-                else
-                  Enum.min(items)
-                end
+                Enum.min(items, Comp)
             end
         end
     end
--- a/test/actions/aggregate_test.exs
+++ b/test/actions/aggregate_test.exs
@@ -52,6 +52,10 @@ defmodule Ash.Test.Actions.AggregateTest do
         public?(true)
       end
 
+      attribute :thing4, :utc_datetime_usec do
+        public?(true)
+      end
+
       create_timestamp :created_at
     end
 
@@ -287,6 +291,34 @@ defmodule Ash.Test.Actions.AggregateTest do
       |> Ash.create!(authorize?: false)
 
       assert %{count: 2} = Ash.aggregate!(Post, {:count, :count}, authorize?: false)
+    end
+
+    test "a datetime field sorts and firsts by chronology" do
+      # Erlang term order compares a struct's fields in alphabetical key order,
+      # so for a DateTime :day (31 vs 1) decides before :year is ever reached.
+      # These two instants therefore order one way by term order and the
+      # opposite way by chronology.
+      early = ~U[2020-12-31 23:00:00.000000Z]
+      late = ~U[2026-01-01 00:00:00.000000Z]
+
+      for at <- [early, late] do
+        Comment
+        |> Ash.Changeset.for_create(:create, %{public: true, thing: "chrono", thing4: at})
+        |> Ash.create!(authorize?: false)
+      end
+
+      mine = Ash.Query.filter(Comment, thing == "chrono")
+
+      assert [^late, ^early] =
+               mine
+               |> Ash.Query.sort(thing4: :desc)
+               |> Ash.read!(authorize?: false)
+               |> Enum.map(& &1.thing4)
+
+      assert %{first: ^late} =
+               mine
+               |> Ash.Query.sort(thing4: :desc)
+               |> Ash.aggregate!({:first, :first, field: :thing4}, authorize?: false)
     end
 
     test "honors tenant" do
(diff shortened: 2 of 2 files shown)


zachdaniel (approved):

zachdaniel:

PR #2825, opened 2026-08-03
chore(ci): run cargo test, fmt and clippy for rustler projects

shahryarjb:
The reusable workflow compiles the NIF as part of `mix test`, but it never runs
the crate's own Rust checks — `cargo test`, `cargo fmt` and `cargo clippy` 

For: https://github.com/ash-project/igniter_js/pull/68

> I ran the tests locally and in the repo CI, and everything passed without issues. That said, since Ash has many projects that require testing, I wanted to make sure the changes wouldn’t cause any pipeline failures—especially since I only modified the Rust portion.

### DIFF
--- a/.github/workflows/ash-ci.yml
+++ b/.github/workflows/ash-ci.yml
@@ -69,6 +69,12 @@ on:
         type: string
       rustler-precompiled-module:
         type: string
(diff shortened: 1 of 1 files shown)


zachdaniel:

PR #2826, opened 2026-08-03
fix: Don't throw exceptions raised based on missing context while validating expr refs

Rutgerdj:
We ran into a regression when upgrading from Ash `3.30.1` to `3.31.0`. 
We have a calculation that has an `expression/2` callback which raises when a certain context key is missing. 

(Illustrative example):
```elixir
defmodule ContextDependent do
  @moduledoc false
  use Ash.Resource.Calculation

  @impl true
  def expression(_opts, context) do
    case context.source_context[:sort_field] do
      nil -> raise "requires :sort_field in the source context"
      field -> expr(^ref(field))
    end
  end
end
```

The recently introduced `validate_sortable` function calls the `expression/2` callback on the calculation without the query context, causing the exception to be raised. 

PR that introduced the error: https://github.com/ash-project/ash/pull/2793/changes

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -1944,49 +1944,22 @@ defmodule Ash.Actions.Read do
       end)
       |> Enum.reduce_while({:ok, []}, fn
         {%Ash.Resource.Calculation{} = resource_calculation, direction}, {:ok, sort} ->
-          case Ash.Query.Calculation.from_resource_calculation(
-                 query.resource,
-                 resource_calculation,
-                 source_context: query.context
-               ) do
-            {:ok, calc} ->
-              case hydrate_calculations(query, [calc]) do
-                {:ok, [{calc, expression}]} ->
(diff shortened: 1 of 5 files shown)


Rutgerdj:
I (through Claude) fixed it by threading the context through so the exception would not be thrown. 

There still remain cases where the context is not known at this pre-check time tho. So for those cases I have it now rescue the exception and continue processing. 

I'm not sure if that's what's best so just let me know if you prefer another approach.

zachdaniel:
I see the issue you're pointing out here, but I think it actually points to the fact that we can't validate portability here at all. I think we need to remove the check that reaches inside of calculations for portability and instead we need to check it in `read.ex` after context has been set. We can leave the checks that check basic attributes as an eager validation though.

zachdaniel:

PR #2828, opened 2026-08-03
fix: sort a ranked union in ETS, not only a ranked union_all

### LINKED ISSUE #2827: ETS sorts a ranked union_all but not a ranked union — maybe_not_distinct? is set for union_all only
### Code of Conduct

### AI Policy

### Versions

| | version |
| --- | --- |
| Ash | 3.31.0 (hex) — also reproduced on main @ `e1238da28` |

### Operating system

### Current Behavior

In `Ash.DataLayer.Ets`, sorting by a combination calculation does nothing when the parts are joined with `union`. The same query with `union_all` sorts correctly.

### Reproduction

Paste into `iex`:

```elixir

defmodule Domain do
end

defmodule Item do

  attributes do
    uuid_primary_key :id
  end

  actions do
    defaults [:read, create: :*]
  end
end

require Ash.Query
import Ash.Expr

Ash.create!(Item, %{name: "alpha"})

# ONE record, reached by three parts that each rank it differently,
# sorted by that rank descending. Only the combination TYPE varies.
run = fn type ->
  Item
  |> Ash.Query.combination_of([
    Ash.Query.Combination.base(
      filter: expr(name == "alpha"),
      calculations: %{so: calc(1, type: :integer)}
    ),
    apply(Ash.Query.Combination, type, [
      [filter: expr(name == "alpha"), calculations: %{so: calc(2, type: :integer)}]
    ]),
    apply(Ash.Query.Combination, type, [
      [filter: expr(name == "alpha"), calculations: %{so: calc(3, type: :integer)}]
    ])
  ])
  |> Ash.Query.sort([{calc(^combinations(:so)), :desc}])
  |> Ash.read!()
  |> Enum.map(& &1.calculations[:so])
end

run.(:union_all)
#=> [3, 2, 1]     correct

run.(:union)
#=> [1, 2, 3]     unsorted
```

Same record, same data, same sort - only the combination type differs.

### Expected Behavior

`run.(:union)` should also return `[3, 2, 1]`.

A `union` keeps both copies of a record when the parts differ on the combination fieldset, which includes the calculations, that is what makes ranking work at all. So a `union` can return the same primary key twice, exactly as `union_all` can.

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Fixes #2827 .

Sorting by a combination calculation did nothing when the parts were joined with `union`. The
same query with `union_all` sorted correctly. One record, reached by three parts ranking it
1/2/3, sorted descending:

```
union_all -> [3, 2, 1]     correct
union     -> [1, 2, 3]     unsorted
```

That is the ranked shape the combinations guide documents. Repro in the issue.

# Cause

`run_query/3` in `lib/ash/data_layer/ets/ets.ex` tells `Ash.Actions.Sort.runtime_sort/3` whether
the fold can return the same primary key twice:

```elixir
maybe_not_distinct? = Enum.any?(combination_of, &(elem(&1, 0) == :union_all))
```

`runtime_sort/3` uses `maybe_not_distinct?` when choosing how to load the sort key — batched, or one record at a time. The batched branch cannot resolve two records sharing a primary key into two records, so the key
is mis-assigned and the sort silently does nothing.

The condition is too narrow. `union_all` is not the only way to get a repeated primary key: for a
`union` the equality basis is the combination fieldset, so parts carrying different combination
calculations keep both copies.

Now widening to any part carrying combination calculations:

```elixir
maybe_not_distinct? =
  Enum.any?(combination_of, fn {type, combination} ->
    type == :union_all or not Enum.empty?(combination.calculations)
  end)
```

`Ash.Query.Combination`'s `calculations` defaults to `%{}`, so the emptiness test has to be
`Enum.empty?/1`. Comparing against `[]` is `true` for `%{}`, which would set the flag for every
combination query and put them all on the unbatched path.

# Commits

1. `test:` pin the ranked three-part shape from the guide, asserted for `union_all` - passes  as-is.
2. `fix:` widen the condition, and assert the same result for `union`.

### DIFF
--- a/lib/ash/data_layer/ets/ets.ex
+++ b/lib/ash/data_layer/ets/ets.ex
@@ -397,7 +397,15 @@ defmodule Ash.DataLayer.Ets do
         _resource,
         parent \\ nil
       ) do
-    maybe_not_distinct? = Enum.any?(combination_of, &(elem(&1, 0) == :union_all))
+    # A fold can return the same primary key twice in two ways: `union_all`
+    # keeps every row, and a `union` between parts carrying different
+    # combination calculations keeps both copies, because the equality basis
+    # is the combination fieldset. `runtime_sort/3` needs to know, or it
+    # batches the sort-key load and cannot tell the copies apart.
+    maybe_not_distinct? =
+      Enum.any?(combination_of, fn {type, combination} ->
+        type == :union_all or not Enum.empty?(combination.calculations)
+      end)
 
     with {:ok, records} when records != [] <-
            get_records(resource, combination_of, parent, tenant),
--- a/test/query_test.exs
+++ b/test/query_test.exs
@@ -304,6 +304,7 @@ defmodule Ash.Test.QueryTest do
       end
 
       assert [3, 2, 1] = ranked.(:union_all)
+      assert [3, 2, 1] = ranked.(:union)
     end
 
     test "combination with offset" do
--- a/test/query_test.exs
+++ b/test/query_test.exs
@@ -274,6 +274,38 @@ defmodule Ash.Test.QueryTest do
       assert Enum.any?(result, &(&1.name == "john"))
     end
 
+    test "a ranked union_all sorts by its combination calculation" do
+      Ash.create!(User, %{name: "fred", email: "a@bar.com"})
+
+      # One record, reached by three parts that each rank it differently.
+      ranked = fn type ->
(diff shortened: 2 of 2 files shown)


zachdaniel:

PR #2836, opened 2026-08-04
fix: report the primary key in the combination fieldset

### LINKED ISSUE #2835: Combinations report a fieldset without the primary key, though every part selects it
### Code of Conduct

### AI Policy

`Ash.DataLayer.Ets` answers a combination whose parts `select:` a fieldset without the primary key;
`AshPostgres.DataLayer` refuses it. Two parts over three records, varying only the `select:`:

| `select:` | AshPostgres | Ets |
| --- | --- | --- |
| `[:region]` -  no primary key | refuses | 2 records |
| `[:region, :id]` | 2 records | 2 records |

Ets answers because it ignores `select:` altogether.
Ash.DataLayer.select/3` gates on the `:select` capability, which Ets does not grant . It materialises whole records and appears to succeed at something it is not doing.

Neither layer narrows the equality basis on `select:`: two records identical on `[:region]` come
back as two, not one. So this is not one layer implementing something the other lacks.

Ets should refuse a combination whose fieldset omits the primary key, matching AshPostgres.

An Ash record is identified by its primary key, so a combination projecting a fieldset without one cannot produce well-formed records. AshPostgres refuses because SQL would hand it rows it cannot turn into records; Ets only appears to cope because it ignores the `select:` that creates the problem.

A combination whose `select:` includes the primary key is unaffected.

**Make Ets honour `select:` and narrow the equality basis**, so a union of two parts selecting `[:region]` returns one record rather than two. 

**Leave the divergence.** A combination keeps meaning something different on each data layer, with
no error either way.

### Additional context

Ideally the ETS error should be a typed `Ash.Error.Query.InvalidQuery` naming the missing primary key rather than the bare `RuntimeError` AshPostgres raises, so it arrives as an invalid-class error like any other query problem.

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Closes #2835

# Summary

Reworked based on guidance from review of #2835

Now the primary key is always selected, so the fix is to make the combination fieldset say so. 

# Cause

`select/3` already adds the primary key and the always-selected attributes to every select
(`lib/ash/query/query.ex:1702-1705`), and each combination part is built through it. But the
fieldset was built from the raw `select:` on the combination, not from the query `select/3`
produced, so it under-reported what the parts select. For a part with `select: [:region]`:

| | |
| --- | --- |
| raw `combination.select` | `[:region]` |
| what `select/3` makes of it | `[:id, :region]` |
| `combination_fieldset` | `[:region]` |

# The change

Derive the fieldset through `select/3`, so it cannot disagree with the queries that function
builds. Always-selected attributes are now reported too, for the same reason.

Nothing in a data layer changes. `Ash.DataLayer.Ets` keys its dedupe off the built query's select
rather than the fieldset, which is why it kept records distinct already.

# Commits
1. `test:` pin that a union of narrowly selected parts keeps records distinct Passes as is.
2. `fix:` derive the fieldset through `select/3`, and assert the primary key is in it.

### DIFF
--- a/lib/ash/query/query.ex
+++ b/lib/ash/query/query.ex
@@ -4600,7 +4600,8 @@ defmodule Ash.Query do
                  combination_of_queries?: true,
                  combination_fieldset:
                    Enum.uniq(
-                     (combination.select || default_select) ++ Map.keys(combination.calculations)
+                     selected_fields(ash_query.resource, combination.select || default_select) ++
+                       Map.keys(combination.calculations)
                    )
                }}
             end
@@ -4613,6 +4614,16 @@ defmodule Ash.Query do
         {:ok, opts[:initial_query] || Ash.DataLayer.resource_to_query(ash_query.resource, domain),
          %{}}
     end
+  end
+
+  # The fieldset must describe what each combination query selects, and `select/3`
+  # decides that — it always adds the primary key and any always-selected
+  # attributes to whatever was asked for.
+  defp selected_fields(resource, select) do
+    resource
+    |> new()
+    |> select(select, replace?: true)
+    |> Map.fetch!(:select)
   end
 
   defp combination_queries(query) do
--- a/test/query_test.exs
+++ b/test/query_test.exs
@@ -185,6 +185,25 @@ defmodule Ash.Test.QueryTest do
                |> Ash.read!()
     end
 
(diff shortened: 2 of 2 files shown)


zachdaniel:

PR #2838, opened 2026-08-04
fix: resolve a combinations reference in the outer filter

### LINKED ISSUE #2837: A ^combinations(...) reference in an outer filter silently matches nothing
### Code of Conduct

### AI Policy

### Versions

| | version |
| --- | --- |
| Ash | 3.31.0 (hex) — also reproduced on main @ `789fb6101` |

### Operating system

### Current Behavior

A `^combinations(...)` reference resolves in an outer sort but not in an outer filter, where it
matches nothing. No error, no warning — just an empty result, indistinguishable from a legitimate
"no matches".

### Reproduction

Paste into `iex`:

```elixir

defmodule Domain do
end

defmodule Item do

  attributes do
    uuid_primary_key :id
  end

  actions do
    defaults [:read, create: :*]
  end
end

require Ash.Query
import Ash.Expr

Ash.create!(Item, %{name: "alpha"})
Ash.create!(Item, %{name: "gamma"})

run = fn outer ->
  Item
  |> Ash.Query.combination_of([
    Ash.Query.Combination.base(
      filter: expr(name == "alpha"),
      calculations: %{sort_order: calc(1, type: :integer)}
    ),
    Ash.Query.Combination.union(
      filter: expr(name == "gamma"),
      calculations: %{sort_order: calc(2, type: :integer)}
    )
  ])
  |> outer.()
  |> Ash.read!()
  |> Enum.map(&{&1.name, &1.calculations[:sort_order]})
end

run.(fn q -> q end)
#=> [{"alpha", 1}, {"gamma", 2}]

run.(fn q -> Ash.Query.filter(q, ^combinations(:sort_order) > 0) end)
#=> []                                every record has sort_order > 0

run.(fn q -> Ash.Query.filter(q, ^combinations(:sort_order) == 2) end)
#=> []                                gamma expected

run.(fn q -> Ash.Query.sort(q, [{calc(^combinations(:sort_order)), :desc}]) end)
#=> [{"gamma", 2}, {"alpha", 1}]      the same reference, sorted correctly
```

The last two are the point: the same reference works in a sort and silently matches nothing in a
filter.

It behaves the same whichever order the query is built in. Filtering before or after
`combination_of/2` both return `[]`.

### Expected Behavior

`^combinations(:sort_order) > 0` should return both records, and
`^combinations(:sort_order) == 2` should return `gamma` record
Whatever the resolution, an unresolvable combination reference in a filter would be better as an
error than an empty result.

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Fixes #2837.

`^combinations(:field)` resolved in an outer sort but matched nothing in an outer filter:

```
unfiltered                             [{"alpha", 1}, {"gamma", 2}]
filter combinations(:sort_order) > 0   []      every record qualifies
filter combinations(:sort_order) == 2  []      gamma expected
sort   combinations(:sort_order) desc  [{"gamma", 2}, {"alpha", 1}]
```

# Cause

`Ash.Filter.do_hydrate_refs/2` turns the bare field name into an `Ash.Query.CombinationAttr`, but
only when the context carries `first_combination`:

```elixir
def do_hydrate_refs(%Ref{combinations?: true, attribute: attribute} = ref,
      %{resource: resource, first_combination: first_combination})
    when is_atom(attribute) and not is_nil(first_combination) do
  # → %CombinationAttr{}
end

def do_hydrate_refs(%Ref{combinations?: true} = ref, _) do
  {:ok, ref}          # ← falls through, attribute still an atom
end
```

Both clauses of `Ash.Filter.Runtime.resolve_ref/5` that handle combination references require the
hydrated struct, so an unhydrated one resolves to `:unknown` and every record is dropped. The sort
path supplies `first_combination`; the filter path did not.

# The change

Supply it in both `do_filter/3` clauses.

That covers a filter added *after* the combinations. One added *before* them cannot be hydrated
when it is built  since there is nothing to resolve it agains, so we must rehydrate existing filter using `combination_of/2` . Both cases are fixed and both are covered by tests; fixing
only the first would have left the other silently returning nothing.

Sorting is untouched as it already supplied the context, and the `test:` commit pins it.

Note this is in `lib/ash/query/query.ex` rather than a data layer, so it applies to any data layer
that resolves combination references at runtime, such as ETS.

# Commits

1. `test:` pin that a combination calculation is usable from the outer query, via the sort path
   that already works. Passes as-is.
2. `fix:` supply `first_combination` when hydrating the filter, rehydrate on `combination_of/2`,
   and assert both orders plus a discriminating predicate.

### DIFF
--- a/lib/ash/query/query.ex
+++ b/lib/ash/query/query.ex
@@ -540,7 +540,31 @@ defmodule Ash.Query do
   def combination_of(query, combinations) do
     query = new(query)
 
-    %{query | combination_of: query.combination_of ++ List.wrap(combinations)}
+    query = %{query | combination_of: query.combination_of ++ List.wrap(combinations)}
+
+    # A `^combinations(...)` reference can only be hydrated once the
+    # combinations are known, so a filter built before this call still carries
+    # the bare field name. Hydrate it now that they are known — otherwise the
+    # reference stays unresolvable and every record is filtered out.
+    case query.filter do
+      nil ->
+        query
+
+      filter ->
+        case Ash.Filter.hydrate_refs(filter, combination_hydration_context(query)) do
+          {:ok, hydrated} -> %{query | filter: hydrated}
+          {:error, error} -> add_error(query, :filter, error)
+        end
+    end
+  end
+
+  @doc false
+  def combination_hydration_context(query) do
+    %{
+      resource: query.resource,
+      public?: false,
+      first_combination: Enum.at(query.combination_of, 0)
+    }
   end
 
   @doc """
@@ -3971,7 +3995,8 @@ defmodule Ash.Query do
                  filter,
                  %{
                    resource: query.resource,
-                   public?: false
+                   public?: false,
+                   first_combination: Enum.at(query.combination_of, 0)
                  }
                  |> with_parent_stack(opts)
                  |> with_conflicting_upsert_values(opts)
@@ -4027,7 +4052,8 @@ defmodule Ash.Query do
                  filter,
                  %{
                    resource: query.resource,
-                   public?: false
+                   public?: false,
+                   first_combination: Enum.at(query.combination_of, 0)
                  }
                  |> with_parent_stack(opts)
                  |> with_conflicting_upsert_values(opts)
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2842, opened 2026-08-08
fix: resolve bulk index from changeset in after_batch filter

emadshaaban92:
If you have a change with batch_change/3 and after_batch/3 on it, and you register it with only_when_valid?: true (or a where), then after_batch/3 gets called with an empty list on Ash.bulk_create/4.

batch_change/3 gets everything fine. It's only after_batch that comes up empty.

Nothing errors either, so if that hook was writing an audit log or kicking off events, you just don't get them and never find out.

The filter deciding which records go to after_batch looks at:

result.__metadata__[:bulk_create_index]
AshPostgres doesn't put that on records. It puts :bulk_action_ref, which is what #2411 made the right thing to set.

So that lookup is nil for every record, and nothing matches.

The changeset is already sitting there in the filter unused, and it has the index on it. So I just took it from there.

The regression test adds a data layer that stamps :bulk_action_ref and not :bulk_create_index — exactly what AshPostgres does — and asserts after_batch still gets its records, and gets only the ones a where matches.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/actions/update/bulk.ex
+++ b/lib/ash/actions/update/bulk.ex
@@ -3363,7 +3363,8 @@ defmodule Ash.Actions.Update.Bulk do
           {matches, non_matches} =
             results
             |> Enum.split_with(fn
-              {:ok, result, _changeset} ->
+              {:ok, result, changeset} ->
+                result = Ash.Actions.Helpers.Bulk.put_metadata(result, changeset)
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2847, opened 2026-08-14
fix: don't raise when an operator on two literals yields a struct

### LINKED ISSUE #2846: operators on two literals raise when the result is a struct
### Code of Conduct

### AI Policy

### Versions

Elixir 1.20.2
Erlang/OTP 29 (erts-17.0.3)
ash 3.31.3

### Operating system

### Current Behavior

Raises `FunctionClauseError` when an operator over two literals evaluates to a struct.

### Reproduction

In iex:
```elixir
Mix.install([{:ash, "== 3.31.3"}])
require Ash.Expr

Ash.Expr.eval(Ash.Expr.expr(^Decimal.new(1) + ^Decimal.new(2)))
```

### Expected Behavior

The result should be evaluated without error:
` Ash.Expr.eval(Ash.Expr.expr(^Decimal.new(1) + ^Decimal.new(2)))` evaluates to `{:ok, Decimal.new("3")}`

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Fixes #2846 

Added tests and fixed guard.

When both operands are literals the operator is evaluated on construction, so what reaches `maybe_operator_expression/2` is the result. `is_struct/1` can't tell that from an operator, so a struct result reached `find_overload/1`, which matches only `%mod{left: _, right: _}`.

Tightening the guard routes those to the existing `:unknown` fallback — the same path `^1 + ^2` already takes.

### DIFF
--- a/lib/ash/filter/filter.ex
+++ b/lib/ash/filter/filter.ex
@@ -4086,10 +4086,13 @@ defmodule Ash.Filter do
     end
   end
 
-  defp maybe_operator_expression(operator, context) when is_struct(operator) do
+  # Over two literals the operator is evaluated on construction, so the result arrives here.
+  defp maybe_operator_expression(
+         %{__operator__?: true, left: left, right: right} = operator,
+         context
+       ) do
     case Ash.Query.Operator.operator_expression(operator) do
       {:ok, expr_module} ->
-        %{left: left, right: right} = operator
         arguments = [left, right]
         resource = context[:resource]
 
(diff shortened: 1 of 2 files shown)


lib/ash/filter/filter.ex
```diff
@@ -4086,7 +4086,10 @@ defmodule Ash.Filter do
     end
   end
 
-  defp maybe_operator_expression(operator, context) when is_struct(operator) do
```
zachdaniel, line 4089: We should pattern match on a key like `__operator__?: true`. I'm pretty sure that is available (or something like it) on operators.

zachdaniel:

PR #2851, opened 2026-08-15
improvement: express a duration in the largest units it is permitted …

### LINKED ISSUE #2850: Normalize duration compatible with units constraints
### Code of Conduct

### AI Policy

Ash.Type.Duration currently behaves differently across various data layers and embedded. For example:

written | PostgreSQL interval | Neo4j duration | Ets term | embedded ISO 8601
-- | -- | -- | -- | --
week: 1 | day: 7 | day: 7 | week: 1 | week: 1
hour: 36 | second: 129600 | second: 129600 | hour: 36 | hour: 36
minute: 90 | second: 5400 | second: 5400 | minute: 90 | minute: 90
month: 18 | month: 18 | month: 18 | month: 18 | month: 18
day: 10 | day: 10 | day: 10 | day: 10 | day: 10

As storage, we want to store and query natively where the database has the capability, however popular databases don't have all the 'bins' that Elixir Duration has:
  PG Interval: months, days, and microseconds
  Neo4j Duration: months, days, seconds, and nanoseconds (more precision than can be represented by Elixir)

We propose`Duration` being consistent across all data layers, and compatible constraints, which are now about representation rather than purely validation. Duration naturally has a `year-month` boundary with `week-day-hour-minute-second-microsecond`.

The solution to this is constraint honouring normalization on both read and write:

#### Write

`apply_constraints/2`

units | supplied | normalised, and stored
-- | -- | --
(none) | hour: 36 | day: 1, hour: 12
(none) | month: 18 | year: 1, month: 6
(none) | day: 10 | week: 1, day: 3
[:hour] | hour: 36 | hour: 36
[:hour] | day: 2 | hour: 48
[:hour] | minute: 90 | rejected
[:week, :hour] | week: 1, day: 1, hour: 5 | week: 1, hour: 29
:day_time | month: 1 | rejected

#### Read

`cast_stored/2`

units | store returns | handed to the caller
-- | -- | --
(none) | day: 7 | week: 1
(none) | second: 129600 | day: 1, hour: 12
[:hour] | second: 129600 | hour: 36
[:week, :hour] | day: 8, second: 18000 | week: 1, hour: 29
[:hour] | second: 10801 | hour: 3 ⚠️ lossy
[:day] | day: 1, hour: 5, minute: 3 | day: 1 ⚠️ lossy
[:year] | month: 18 | year: 1 ⚠️ lossy

This changes what an existing `:duration` attribute reads back, whether or not it has a `units` constraint:  `month: 18` becomes `year: 1`, `month: 6`, `day: 10` becomes `week: 1`, `day: 3`. Values already expressed in the largest units that hold them are unchanged. It also changes what a create or update returns, since normalization applies on write too.

This should be done now, since AshPostgres has never fully supported duration (see ash-project/ash_postgres#553 - recommended re-opening so it can be completed). Completing it depends on this: without normalization, persisting to a Postgres `interval` hands back `second: 129600` for a written `hour: 36`, and under a `units` constraint that value is rejected on re-save. Doing the representation work first also means there is no stored corpus to migrate, which will not be true once persistence lands.

Considered and rejected a `cast_stored/2` only solution, for two reasons:

Normalizing only on read guarantees that what a create or update returns disagrees with what a later read returns — the action returns the cast input, the read returns the stored value.
Normalizing on write allows a constraint to be present or added where there is imported/existing data.

### Additional context

_No response_

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Fixes #2850

Added tests, and normalisation on both sides of storage.

A duration is now expressed in the largest units it is permitted to use. No `units` constraint permits every unit, so there is one rule rather than two paths.

`apply_constraints/2` normalises before validating, so it returns a modified value, this is done so a return matches a later round-trip read.

`cast_stored/2` truncates towards zero where the permitted units cannot express what the store holds, because a read has to answer with something the attribute can hold; a write still refuses, because it must not lose part of what was asked for.

`may_support_atomic_update?/1` is deliberately unchanged as an atomic update skips apply_constraints/2, so nothing normalises on the way in however as every route back out of the store goes through cast_stored/2 what's returned is canonical either way. Measured on both Ets and Postgres.

### DIFF
--- a/lib/ash/type/duration.ex
+++ b/lib/ash/type/duration.ex
@@ -11,7 +11,7 @@ defmodule Ash.Type.Duration do
     units: [
       type: {:or, [{:in, [:year_month, :day_time]}, {:list, {:in, @duration_units}}]},
       doc: """
-      The units the value may be expressed in. A duration is always re-expressed in the largest of these units that will hold it, on the way in and on the way out, so `[:week, :hour]` turns `1 week 1 day 5 hours` into `1 week 29 hours`. A value being written that no combination of the permitted units expresses exactly is rejected — including anything that would have to cross the year/month to week/day boundary, which no conversion can. A value being *read* is truncated to what the permitted units can say instead of being refused, so a store holding `1 day 5 hours 3 minutes` in a `[:day]` attribute reads back as `1 day`: lossy, but readable and writable again. Either an explicit list of units, or a shorthand for one side of that boundary: `:year_month` (`[:year, :month]`) or `:day_time` (`[:week, :day, :hour, :minute, :second, :microsecond]`). Confining an attribute to a single side keeps its values comparable (see `Ash.Type.Duration.compare/2`). With no constraint every unit is permitted, so the same normalization applies and nothing is ever lost.
+      The units the value may be expressed in. A duration is always re-expressed in the largest of these units that will hold it, on the way in and on the way out, so `[:week, :hour]` turns `1 week 1 day 5 hours` into `1 week 29 hours`. A value that no combination of the permitted units expresses exactly is rejected — including anything that would have to cross the year/month to week/day boundary, which no conversion can. This applies on the way out as well as in: a stored duration the permitted units cannot express is refused rather than quietly rewritten. Either an explicit list of units, or a shorthand for one side of that boundary: `:year_month` (`[:year, :month]`) or `:day_time` (`[:week, :day, :hour, :minute, :second, :microsecond]`). Confining an attribute to a single side keeps its values comparable (see `Ash.Type.Duration.compare/2`). With no constraint every unit is permitted, so the same normalization applies and nothing is ever lost.
       """
     ]
   ]
@@ -75,7 +75,7 @@ defmodule Ash.Type.Duration do
 
   def apply_constraints(%Duration{} = value, constraints) do
     allowed = permitted_units(constraints[:units])
-    normalized = normalize_units(value, allowed, :reject)
+    normalized = normalize_units(value, allowed)
 
     case disallowed_units(normalized, allowed) do
       [] ->
@@ -125,15 +125,17 @@ defmodule Ash.Type.Duration do
   @impl true
   def cast_stored(nil, _), do: {:ok, nil}
 
+  # A stored value gets the same treatment as one being written: a duration the
+  # permitted units cannot express is refused, not quietly rewritten into one they can.
   def cast_stored(value, constraints) when is_binary(value) do
     with {:ok, duration} <- cast_input(value, constraints) do
-      {:ok, normalize_units(duration, permitted_units(constraints[:units]), :truncate)}
+      apply_constraints(duration, constraints)
     end
   end
 
   def cast_stored(value, constraints) do
     with {:ok, duration} <- Ecto.Type.load(:duration, value) do
-      {:ok, normalize_units(duration, permitted_units(constraints[:units]), :truncate)}
+      apply_constraints(duration, constraints)
     end
   end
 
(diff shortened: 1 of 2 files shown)


zachdaniel:
I don't think that this is correct? A duration of "1 year" applied as a distance from different starting points would yield different results that a duration of 365 days worth of milliseconds. Consider leap years for example.

matt-beanland:
No it won't do that as year-month has a boundary between week-day-hour-minute-second-microsecond, regardless of whether you constrain any units.

matt-beanland:
It is normalising towards fewer of the larger units. So 1 yr says 1 yr. 365 days would be normalised to 52 weeks and 1 day if weeks and days allowed, regardless of whether years are allowed. So magnitude is correct, and representation normalised.

matt-beanland:
Also if it reads 365 days but allow weeks and not days your'll get 52 weeks.

matt-beanland:
The comparator is unchanged from the other day, still follows PG convention for ordering.

matt-beanland:
Zach I've added tests that validate and pin the assertions above — second commit, tests only.

The central one uses Elixir's own shift rather than`compare/2`, so it doesn't lean on our own convention: for a table of durations across three anchors including 2024-02-29, `DateTime.shift(t, input) == DateTime.shift(t, normalized)`. If normalizing changed a magnitude they'd land on different instants.

And on your example directly: 365 days normalizes to `week: 52, day: 1`, never a year — `Date.shift(~D[2024-01-01], ...)` still gives `2024-12-31`, not `2025-01-01`, so the distinction you're pointing at survives. However to shift a time by a year, use `year: 1` or `month: 12`.

The summary should have said this and didn't: `year/month` and `week/day/hour/minute/second/microsecond` normalize independently, and nothing converts between them, constrained or not.

In addition I've added tests which prove that mixed signs within a bucket collapse to a single consistent representation, which is a normalization benefit I didn't include above: `day: 1, hour: -5` and `hour: 19` are the same duration, and after normalization are both `hour: 19`.

Also switched to US spelling to match the codebase.

lib/ash/type/duration.ex
```diff
@@ -11,7 +11,7 @@ defmodule Ash.Type.Duration do
...
+      The units the value may be expressed in. A duration is always re-expressed in the largest of these units that will hold it, on the way in and on the way out, so `[:week, :hour]` turns `1 week 1 day 5 hours` into `1 week 29 hours`. A value being written that no combination of the permitted units expresses exactly is rejected — including anything that would have to cross the year/month to week/day boundary, which no conversion can. A value being *read* is truncated to what the permitted units can say instead of being refused, so a store holding `1 day 5 hours 3 minutes` in a `[:day]` attribute reads back as `1 day`: lossy, but readable and writable again. Either an explicit list of units, or a shorthand for one side of that boundary: `:year_month` (`[:year, :month]`) or `:day_time` (`[:week, :day, :hour, :minute, :second, :microsecond]`). Confining an attribute to a single side keeps its values comparable (see `Ash.Type.Duration.compare/2`). With no constraint every unit is permitted, so the same normalization applies and nothing is ever lost.
```
zachdaniel: I think we should apply these constraints as aggressively on the cast_stored path as we do elsewhere. Otherwise reading through your updated explanation this makes sense to me 👍

matt-beanland: Yes thanks you are correct, I've resolved this and now consistent with behaviour of core types.

matt-beanland:
I've updated this one to apply constraints aggressively as you suggested

zachdaniel:

PR #2852, opened 2026-08-16
fix: declare the duration overloads of the plus and minus operators

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Partially delivers https://github.com/ash-project/ash_postgres/issues/553 (Duration)
Enables an ash_postgres test PR

`Ash.Query.Operator.Basic` implements every temporal ± duration combination in `evaluate/1` — `DateTime`, `Date`, `Time`, `NaiveDateTime` and `Duration` — so these forms work on data layers that evaluate expressions with Ash runtime. 

However an error was discovered that can't be reproduced without a datalayer which doesn't use Ash runtime for expression evaluation. As `plus` and `minus` declare no `types` they take the default `[[:same, :any]]`, which types the right operand as the left. A `%Duration{}` cannot be cast to a datetime, so for example on a SQL-backed data layer such as ash_postgres the expression is refused with `Invalid filter value`.

This PR declares the overloads, as `times` already does for `duration * integer`. It resolves each temporal type to a distinct module, so all seven are listed rather than a single `:datetime`.

Regression tests assert the resolved operand types through `Ash.Expr.determine_types/2`. These assert that a duration operand stays a `Ash.Type.Duration` and the expression returns the temporal type.

No ash_sql change is needed: it renders a `:duration`-typed parameter correctly already, it was simply never told the operand was one. Verified against PostgreSQL 19 — `at + ^dur > ^dt`, `at - ^dur < ^dt`, a multi-unit duration, the reversed `^dur + at` and `date + ^dur` all execute and return the expected rows.

### DIFF
--- a/lib/ash/query/operator/basic.ex
+++ b/lib/ash/query/operator/basic.ex
@@ -9,7 +9,43 @@ defmodule Ash.Query.Operator.Basic do
     plus: [
       symbol: :+,
       no_nils: true,
-      evaluate_types: :numbers
+      evaluate_types: :numbers,
+      types: [
+        [:same, :any],
+        [:duration, :duration],
+        [:date, :duration],
+        [:duration, :date],
+        [:datetime, :duration],
+        [:duration, :datetime],
+        [:utc_datetime, :duration],
+        [:duration, :utc_datetime],
+        [:utc_datetime_usec, :duration],
+        [:duration, :utc_datetime_usec],
+        [:naive_datetime, :duration],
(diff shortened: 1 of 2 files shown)


zachdaniel:
got some merge conflicts here

matt-beanland:
@zachdaniel conflict was mine with a line of test documentation and now resolved

zachdaniel:

PR #2862, opened 2026-08-17
feat: order ranges by their bounds, as Postgres does.

matt-beanland:
# Contributor checklist

- [X] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [X] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Closes #2861

Requires #2858 and #2860 to be merged as it uses `empty?/1`, the inclusivity predicates, and :eq only means "same set" once discrete ranges canonicalize.

Commit highlights:

- It changes existing sort results: `Comp.compare/2` changes its answer, which moves anything sorting through Ash

- every tie-break is Postgres's answer, not a choice. Verified on `numrange` specifically, so discrete canonicalization couldn't rewrite the bounds and hide the rule being tested: `[)` before `()` at the same lower,`5)` before `5]` at the same upper, unbounded lower first, unbounded upper last.

- empty's position is a convention, not a derivation. The non-empty part is Allen's canonical ordering of the interval relations projected onto three outcomes, but Allen defines relations only for non-empty intervals, so empty-first comes from Postgres.

- `compare/2` answers position, not containment. A future proposal should answer contains/during in Allen's terms.

PR #2860, opened 2026-08-17
fix: canonicalize the bounds of a discrete range

matt-beanland:
# Contributor checklist

- [X] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [X] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Closes #2859

Requires ash-project/ash#2858 for compilation and correctness.

The commit `fix: canonicalize the bounds of a discrete range` includes:
* `:integer` and `:date` ranges cast to `[)`: an exclusive lower and an inclusive upper each move to the next value, and an unbounded end is exclusive. Continuous inner types are unchanged.
* the shift can result in an empty range : `(1,2)` on integers becomes `[2,2)`, which  is  canonical empty range: `Ash.Range.empty/0`, flagged `empty?: true`, which Postgres renders as `empty`
* an inverted range is left as written, for `apply_constraints/2` to reject as before.
* `apply_constraints/2` canonicalises too, so a value that skipped casting still normalises.

PR #2864, opened 2026-08-17
feat: constrain a range's bounds and emptiness

### LINKED ISSUE #2863: add constraints to range
### Code of Conduct

### AI Policy

ranges are more useful with enforced constraints on lower and upper bounds, and an empty constraint.

We think the right design to constrain bounds and empty is 3 constraints. These are the constraints and their proposed defaults (unconstrained bounds, empty rejected)
```elixir
lower: [required?: false]   # inclusive? unset — omit `lower` entirely for no constraint
upper: [required?: false]
allow_empty?: false
```

We considered 3 constraints for bounds, one `bounds` (e.g. inclusive_exclusive) when values exist, and separate constraints on whether `unbounded_upper` and/or `unbounded_lower` is allowed, but this was unwieldy and has side effects with canonicalisation, where values that were never written can be rejected. It seems better to declare `lower` and `upper` separately, listing keyword for each concern.

### Additional context

_No response_

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Closes #2863

Requires #2860 and  #2862 

Commit highlights:

-  each end is checked on its own terms; present if required, carrying the asked-for inclusivity if it's there. An absent end has only its presence constrained. That's what stops a canonicalized `(,6)` failing an `inclusive?: true` lower.

- `allow_empty?: false` rejects, and there's a test pinning the message. The default behaviour is changed, since empty was allowed previously.

- the constraints compose into a period, pinned as a test: `lower: [required?: true, inclusive?: true], upper: [inclusive?: false]` accepts `[valid_at, ∞)`, refuses an unbounded start, refuses an inclusive end, refuses empty - exactly what we need for temporal.

### DIFF
--- a/lib/ash/range.ex
+++ b/lib/ash/range.ex
@@ -67,4 +67,59 @@ defmodule Ash.Range do
   @doc "Whether the range's upper bound includes the point it names."
   @spec upper_inclusive?(bounds()) :: boolean()
   def upper_inclusive?(bounds), do: bounds in [:"(]", :"[]"]
+
+  @doc """
+  Compares two ranges as Postgres orders them: empty first, then by lower bound, then
+  by upper, an unbounded end as `-∞`/`+∞`, and the earlier boundary first where two
+  bounds name the same value (`[1` before `(1`, `5)` before `5]`).
+
+  A sort order rather than containment: `[1,10)` sorting before `[3,5)` says nothing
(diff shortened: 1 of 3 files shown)


zachdaniel:

PR #2866, opened 2026-08-17
feat: classify the Allen relation between two ranges

### LINKED ISSUE #2865: classify which relation holds between two ranges
### Code of Conduct

### AI Policy

A range can be tested for overlap, but there is no way to ask which of the thirteen interval relations holds, so every caller must reconstruct before / meets / during / finishes from bound comparisons themselves.

Ash.Range.relation/2 returning one of Allen's thirteen atoms, with relations/0 giving the canonical order.

 individual predicates were rejected as can't express "which one", and can't be composed into a composition table.

a dependency on ex_tempo was rejected, although it has the same relations and also orders them canonically.

### Additional context

Allen algebra was invented for time, but is actually order theoretic and holds for ranges with ordered inner types (`:integer` and `:date` ranges) so is very useful for non-temporal ranges.

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Closes #2865 

commit highlights:

- `relation/2` classifies, `relations/0` gives the canonical order. 

- two ranges meet when one ends where the other begins and exactly one of them includes that point. `[1,5)` meets `[5,9)`; `[1,5]` overlaps it instead; `(5,9)` merely follows it. Matches Postgres `-|-`.

- an empty range gets `nil`. Postgres says `empty && empty` is false, so there is no arrangement to name, and every other relation asserts one.

- the thirteen relation atoms are pinned individually, plus the converse-mirror property: swapping the operands walks the canonical order inward from `equals`.

### DIFF
--- a/lib/ash/range.ex
+++ b/lib/ash/range.ex
@@ -112,6 +112,90 @@ defmodule Ash.Range do
     end
   end
 
+  @allen [
+    :precedes,
+    :meets,
+    :overlaps,
+    :finished_by,
+    :contains,
+    :starts,
+    :equals,
+    :started_by,
+    :during,
+    :finishes,
+    :overlapped_by,
+    :met_by,
+    :preceded_by
+  ]
+
+  @typedoc "One of Allen's thirteen interval relations."
+  @type allen :: unquote(Enum.reduce(Enum.reverse(@allen), &{:|, [], [&1, &2]}))
+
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #2878, opened 2026-08-18
feat: add range expressions for lower and upper bounds

### LINKED ISSUE #2875: add range expressions for lower and upper bounds
### Code of Conduct

### AI Policy

A range attribute's endpoints cannot be reached in an expression. There is no way to sort by where a range starts, filter on ranges beginning after a given value, or select an endpoint as a calculation.

The workaround is a calculation in Elixir, which cannot be pushed down, so sorting or filtering by an endpoint loads every row first.

- `range_lower/1` returning a range's lower endpoint, of its inner type. 
-  `range_upper/1` returning a range's upper endpoint, of its inner type. 

Both are `nil` where the range is unbounded at that end, and where it is `empty`, as Postgres's` lower()` and `upper()` are.

bare `lower/1` and `upper/1`: the names are free, but SQL users expect upper() to be string uppercase. Postgres can spell them that way because it overloads on argument type.

`lower_bound/1` and `upper_bound/1`: a bound is the endpoint plus its inclusivity, and these return only the endpoint.

`returns [:any]`, as composite_type does, or a `{:range, :same}` clause in determine_types beside the existing `{:array, :same}`. `[:any]` is proposed, as the smaller change.

### Additional context

_No response_

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Closes #2875 

Adds `range_lower/1` and `range_upper/1`. Both are nil where the range is unbounded at that end, and where it is empty.


zachdaniel:

PR #2887, opened 2026-08-21
fix: restore temporal subtraction type inference

jechol:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

The temporal-duration signatures added in #2852 left same-type temporal subtraction undeclared.

The runtime evaluator already supports operations such as `Date.diff/2`, but type resolution reports `date - date` as returning `date`. An untyped `lazy/1` operand can also select the `date - duration` signature even when the surrounding calculation expects an integer, causing SQL data layers to reject the resulting date value.

This change:

- declares same-type temporal subtraction as returning an integer;
- prefers signatures matching the known result type when operands are otherwise ambiguous;
- preserves temporal-duration subtraction behavior.

Regression tests cover all supported temporal types and distinguish a lazy date from a lazy duration using the known calculation result.

Verified with the full Ash test suite (3,971 passing) and the original failing expression in an AshPostgres application without a SQL fragment workaround.

This PR description was drafted with AI assistance and reviewed before submission.

### DIFF
--- a/lib/ash/expr/expr.ex
+++ b/lib/ash/expr/expr.ex
@@ -1526,6 +1526,7 @@ defmodule Ash.Expr do
       end
     end)
     |> Enum.filter(& &1)
+    |> prefer_known_result(known_result)
     |> case do
       [{types, returns, _, _}] ->
         {types, returns}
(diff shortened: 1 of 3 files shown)


zachdaniel:

PR #2895, opened 2026-08-22
improvement: only set `transaction? true` when the data layer supports it

C-Sinclair:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

From @zachdaniel's suggestion on ash-project/ash_sqlite#223

Create, update and destroy actions default to `transaction? true` regardless of the data layer. For data layers that can't transact, the flag is set incorrectly.

`Ash.Changeset.with_hooks/3` and `Ash.DataLayer.transaction/5` both check `can?(:transact)` first, so never actually run anything, but its still appearing as `true` and causing warnings when trying to extend AshSqlite with transactions.

Bottom line: no runtime behaviour changes. Just introspection.
create, update and destroy on a data layer that cannot transact now report `transaction?: false`. 

Allows the verifier discussed in https://github.com/ash-project/ash_sqlite/pull/223 to be dropped/

## Alternative approach 

One alternative I didn't reach for immediately: set the struct `transaction?` fields as `nil` by default, not `true`. We'd then set `true` if the datalayer can transact rather than the other way around. 

This is tempting, but it would mean that that field would become nillable.


lib/ash/resource/transformers/set_action_transactions.ex
```diff
@@ -0,0 +1,42 @@
...
+    if Ash.DataLayer.can?(:transact, dsl_state) do
+      {:ok, dsl_state}
+    else
+      transacting_actions =
+        dsl_state
+        |> Transformer.get_entities([:actions])
+        |> Enum.filter(&(&1.type in @mutations && &1.transaction?))
+
+      {:ok,
+       Enum.reduce(transacting_actions, dsl_state, fn action, dsl_state ->
```
zachdaniel, line 32: Hm...wasn't there a different transformer that sets transaction? to true for these? If so couldn't we just modify that to not do it when the data layer supports transactions? I might have to look deeper here as it's been a while since i worked on these, but I think there might a simpler approach

C-Sinclair: As far as I could see, it was set in the structs for the actions as the default `true`. So in this case we need to unset it. But I agree it feels OTT here.

zachdaniel:

PR #2903, opened 2026-08-25
ci: make the `ash-ci` audit steps individually optional

jimsynz:
The `auditor` job in the shared `ash-ci` reusable workflow runs two audit steps unconditionally. `mix hex.audit` is built into Hex and always available, but `mix deps.audit` comes from the `mix_audit` dependency. Any consuming repo that drops `mix_audit` gets a hard CI failure — "the task deps.audit could not be found" — with no way to opt out.

This adds two boolean inputs, `deps-audit` and `hex-audit`, both defaulting to `true`, and gates the corresponding steps with `if:`. It is a pure opt-out: every existing consumer keeps its current behaviour with no change to its own workflow file.

The two checks read different advisory databases — `mix_audit` uses the dependabot advisory database, `hex.audit` uses the EEF/OSV one — so they are not interchangeable. A repo was recently found with a clean `mix_audit` run against a lockfile carrying 21 advisories, 4 of them HIGH, which is why some repos want to move to `hex.audit` alone.

### DIFF
--- a/.github/workflows/ash-ci.yml
+++ b/.github/workflows/ash-ci.yml
@@ -41,6 +41,12 @@ on:
       sobelow:
         type: boolean
         default: true
+      deps-audit:
+        type: boolean
+        default: true
+      hex-audit:
+        type: boolean
+        default: true
       postgres:
         type: boolean
         default: false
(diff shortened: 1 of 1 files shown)


zachdaniel:

PR #2912, opened 2026-09-01
fix: authorize atomic updates on resources with more than one authorizer

jimsynz:
Reported by a client on 3.31.3. Present since 3.0.

`Ash.Can.alter_source/5` returns `{:ok, true, changeset, query}` when it authorizes a changeset against a base query. The reducer that walks the authorizers only accepted `{:ok, true, subject}`, so the second authorizer raised a `FunctionClauseError`.

Atomic updates and destroys set `base_query`, so any resource with two or more authorizers failed every atomic update. Only `Ash.Policy.Authorizer` ships with Ash, so a second authorizer is always application code — which is how this survived so long.

Two further defects sat behind the crash:

- Each authorizer now alters the query the previous one produced. The old code always passed the original `base_query`, so authorizer 2 discarded authorizer 1's filter.
- The reducer stops on the first error instead of raising the same `FunctionClauseError`.

Not a security issue — the crash fails closed, and the dropped-filter path was unreachable because the crash always came first. Worth noting for anyone patching this locally: widening the pattern match without threading the query turns a fail-closed crash into a dropped authorization filter.

New test in `test/authorizer/multiple_authorizers_test.exs` fails on `main`, passes here. Full suite: 4059 passed.

### DIFF
--- a/lib/ash/can.ex
+++ b/lib/ash/can.ex
@@ -830,75 +830,25 @@ defmodule Ash.Can do
           {:ok, true, subject}
 
         authorizers ->
-          authorizers
-          |> Enum.reduce(
-            {:ok, true, subject},
-            fn authorizer, {:ok, true, subject} ->
-              authorizer_state =
-                Ash.Authorizer.initial_state(
-                  authorizer,
-                  actor,
-                  subject.resource,
-                  subject.action,
-                  domain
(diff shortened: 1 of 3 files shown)


zachdaniel:

PR #2913, opened 2026-09-03
fix: avoid compile dependencies for runtime-only constraint modules

jechol:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

A `constraints` value can hold module references with different dependency requirements:

```elixir
attribute :line, :struct do
  constraints instance_of: MyApp.Invoice,
              fields: [amount: [type: MyApp.MoneyType]]
end
```

`instance_of` is only read at runtime, by `cast_input/2`, `matches_type?/2` and friends. `type` is initialized while the DSL is compiled, by `Ash.Type.init/2` from `set_type_transformation/1`.

Since `constraints` is unclassified, both currently create compile dependencies, so a resource recompiles whenever a struct it merely validates against changes.

Which of the two a reference is, is known by the type and not by the DSL — only `Ash.Type.Struct` knows what `instance_of` means. So instead of naming constraint keys in `Ash.Resource.Dsl`, this puts `constraints` in `no_depend_modules` and registers the dependencies from what the type already reports:

- `Ash.Type.constraint_referenced_types/2` collects the types reachable through `c:Ash.Type.referenced_types/1` — the same graph `detect_type_cycle!/2` walks.
- `Ash.Resource.Transformers.TrackConstraintTypeDependencies` traces each of them with `Macro.compile_apply/4`, against the `Macro.Env` Spark persists as `:env`.

`instance_of` drops out by construction: `Ash.Type.Struct.referenced_types/1` returns `constraints[:fields]` and not `instance_of`. Third-party types get the same treatment without touching the core DSL, and fields marked `init?: false` stay excluded, matching the existing cycle walk.

On a private application this cuts the recompile set for a representative resource edit from 58 files to 41.

### DIFF
--- a/lib/ash/resource/transformers/track_constraint_type_dependencies.ex
+++ b/lib/ash/resource/transformers/track_constraint_type_dependencies.ex
@@ -1,0 +1,95 @@
+# SPDX-FileCopyrightText: 2019 ash contributors <https://github.com/ash-project/ash/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Ash.Resource.Transformers.TrackConstraintTypeDependencies do
+  @moduledoc """
+  Registers compile dependencies for the types reachable from a field's constraints.
+
+  `constraints` is in `no_depend_modules`, because the module references it can hold do not
+  share one dependency requirement. `instance_of` is only read at runtime, while a nested
+  `type` is initialized while the DSL is compiled, so its compile-time state would go stale
+  if the resource stopped recompiling when it changed.
+
+  Which of the two a module reference is, is known by the type, not by the DSL: only
+  `Ash.Type.Struct` knows what `instance_of` means. So the dependency is registered here,
+  from what `c:Ash.Type.referenced_types/1` reports, rather than declared in
+  `Ash.Resource.Dsl` against constraint keys the DSL is otherwise unaware of.
+
+  Opt in with:
+
+      config :ash, :constraint_dependencies_from_referenced_types?, true
+
+  A type that references other types without implementing `c:Ash.Type.referenced_types/1`
+  keeps its compile dependencies while this is off.
+  """
+  use Spark.Dsl.Transformer
+
+  alias Spark.Dsl.Transformer
+
+  def after_compile?, do: false
+
+  # Runs as a persister rather than a transformer: `:auto` types are replaced with real ones
+  # by `ResolveAutoTypes`, which is itself a persister, and Spark silently ignores ordering
+  # declarations that cross the two phases.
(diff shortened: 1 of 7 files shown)


lib/ash/resource/transformers/track_constraint_type_dependencies.ex
```diff
@@ -0,0 +1,84 @@
...
+    {:ok, dsl_state}
+  end
+
+  # The trace has to happen against the env of the module being compiled, which Spark
+  # persists for us. Only the dependency is wanted here: `init/1` has already run for these
+  # types, through the containing type's `init/2`, so calling it again would repeat that
+  # work and invoke a callback that is not required to tolerate a second invocation.
+  defp register({type, constraints}, env) do
+    type
+    |> Ash.Type.constraint_referenced_types(constraints)
```
zachdaniel, line 55: I think we're going to have to make this opt-in, because there could be external types that have not added `referenced_types` but were relying on these compile time dependencies existing. We could turn it on in 4.0 perhaps and require that any type that references other types declare it in order to get compile time dependencies on their constraints.

jechol: Switched it to opt-in:

```elixir
config :ash, constraint_dependencies_from_referenced_types?: true
```

It defaults to `false`. While off, `constraints` creates the compile dependencies it does today, so a type that does not implement `referenced_types/1` loses nothing.

Ash's own config turns it on, the way `bulk_actions_default_to_errors?` does — otherwise the new path never runs in the test suite. I left it out of the installer: a dependency's type may not implement the callback, so it seemed better for it to arrive when the default changes in 4.0.

One thing I would like your read on.

`set_type_transformation` calls `detect_type_cycle!` with the constraints as written, but stores the ones `init/1` returns. The persister reads what was stored, so a type whose `init/1` consumes and drops a constraint it reads in `referenced_types/1` is invisible to it. That is the difference from Spark reading the aliases in the DSL itself.

None of the built-in types remove such a constraint — they normalise it in place — so nothing is affected today. For now I documented it on `c:referenced_types/1`: it is called both before and after `init/1`, so it has to report the same types either way.

Closing it properly means carrying the pre-init walk through to the persister, which adds a field to seven entity structs, `Ash.Resource.Attribute` among them. It would also collapse the two type-graph walks into one.

I am unsure whether that is worth the surgery. In a private application, cutting 40 of 52 `instance_of` usages took a representative resource edit from 58 recompiled files to 41, so the accuracy of this mechanism matters to me — but if the documented contract is enough, I am happy to leave it there.

PR #2914, opened 2026-09-04
docs: explain when and why to use `before_action?` on validations

barnabasJ:
## Summary

`before_action?` on validations was only documented mechanically ("the validation will be run in a before_action hook") with no explanation of when or why you would want it. The original 1.x name, `expensive?`, carried the intent, but that was lost in the 2.0 rename.

- **Validations guide**: new `before_action? Option` section explaining that validations normally run when the changeset/query is *built*, and the two reasons to defer one to execution time:
  1. It is expensive (runs a query, calls a service) and changesets are built far more often than executed. `AshPhoenix.Form.validate/3` rebuilds the changeset on every change event, so a querying validation would otherwise hit the database on every keystroke.
  2. It needs to observe state at execution time, inside the data layer transaction, right before the write.

  Also lists the trade-offs: errors surface only on submit, it cannot be atomic (`require_atomic? true` fails, bulk falls back to `:stream`), it runs after `before_transaction`/`around_transaction` hooks, `only_when_valid?` is checked at hook time, and `delay_global_validations?` is the same thing applied to all global validations.
- **Actions lifecycle guide**: the "Inside Transaction" bullet listed action/global validations, contradicting the paragraph below it which says they run during changeset creation before the transaction. Corrected to hooks plus `before_action?` validations.
- **Option docstring** in `Ash.Resource.Validation` expanded, DSL cheat sheet regenerated.

Behavior verified against `Ash.Changeset.validate/5` (hook wrapping), `Ash.Changeset.with_hooks/3` (transaction boundary) and `AshPhoenix.Form.validate/3`.

## Test plan

- [x] `mix spark.cheat_sheets` regenerated cleanly
- [x] `mix test test/resource/validation_test.exs test/actions/bulk/bulk_update_test.exs test/actions/update_test.exs` (105 passed)

### DIFF
--- a/documentation/dsls/DSL-Ash.Resource.md
+++ b/documentation/dsls/DSL-Ash.Resource.md
@@ -1110,7 +1110,7 @@ validate present([:first_name, :last_name], at_least: 1)
 | [`only_when_valid?`](#actions-action-validate-only_when_valid?){: #actions-action-validate-only_when_valid? } | `boolean` | `false` | If the validation should only run on valid changesets. Useful for expensive validations or validations that depend on valid data. |
 | [`message`](#actions-action-validate-message){: #actions-action-validate-message } | `String.t` |  | If provided, overrides any message set by the validation error |
 | [`description`](#actions-action-validate-description){: #actions-action-validate-description } | `String.t` |  | An optional description for the validation |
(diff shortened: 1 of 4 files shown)


zachdaniel (approved):

zachdaniel:

PR #2916, opened 2026-09-07
fix: don't pass the extra pagination row to read hooks

barnabasJ:
Builds on #2607 by @Rutgerdj (his test commit is included here, rebased onto current `main`). Supersedes that PR if merged.

## Problem

Pagination asks the data layer for `limit + 1` rows so `more?` can be determined. That extra row used to be trimmed in `to_page`, which runs *after* `authorize_results` and `after_action` hooks. So every hook on a paginated read received one record the caller never sees. For relationship pagination that was one phantom row *per parent*, since the lateral join applies the limit per parent.

## Fix

- Drop the extra row directly after the data layer returns, before keysets, `authorize_results` and `after_action`.
- Remember `more?` on the query context (`pagination_more?`), since the rows can no longer tell us.
- For paginated relationship loads (lateral joins), drop the extra row per parent and carry a per-parent `more?` map through `Ash.Page.Unpaged.more_by_source` to where each parent's page is built in `Ash.Actions.Read.Relationships`.
- Data-layer keyset pagination is untouched: it never fetched an extra row.

## Open question for @zachdaniel: `Ash.data_layer_query/2`

This PR deliberately does **not** change the `Ash.data_layer_query/2` path. Its `run` callback still hands hooks the raw `limit + 1` rows, and `load` still trims and computes `more?` from the leftover. Reason: `run` returns only `{:ok, rows}` and `load` receives the query from the caller, so there is no channel to carry `more?` from one to the other without changing that public contract.

The one clean fix would be for `run` to build and return the page itself (the typespec already declares `list | Ash.Page.page()` as its return, and `load` already accepts a page), but that changes behaviour for every caller matching a list out of `run.()`. I've documented the current behaviour in the `Ash.data_layer_query/2` docs and left a comment at the call site. Happy to do the contract change as a follow-up if you think it's the right direction.

## Tests

- `test/page_test.exs`: hook sees exactly the page for offset, keyset forward, keyset backward, last page, exact last page, and `read_one`.
- `test/actions/load_test.exs`: per-parent pages with equal and differing child counts, keyset ordering per parent, two paginated loads on one parent, exact last page.

Full suite passes.

### DIFF
--- a/lib/ash/actions/read/read.ex
+++ b/lib/ash/actions/read/read.ex
@@ -1069,78 +1069,62 @@ defmodule Ash.Actions.Read do
              :ok <- validate_combinations(query, calculations_at_runtime, query.load),
              {:ok, data_layer_query} <-
                Ash.Query.data_layer_query(query, data_layer_calculations: data_layer_calculations) do
+          # Pages remember their read opts so `Ash.page/2` can rerun the read.
+          # A rerun must produce the next page, not another data layer query.
+          page_opts = Keyword.delete(opts, :data_layer_query?)
+
           {:ok,
            %{
              query: data_layer_query,
              ash_query: query,
              load: fn query_ran, data ->
-               with {:ok, data} <-
-                      load_through_attributes(
-                        data,
-                        %{
-                          query_ran
-                          | calculations: Map.new(calculations_in_query, &{&1.name, &1})
-                        },
-                        query.domain,
-                        opts[:actor],
-                        opts[:tracer],
-                        opts[:authorize?]
-                      ),
-                    {:ok, data} <-
-                      load_relationships(data, query, opts),
-                    {:ok, data} <-
-                      Ash.Actions.Read.Calculations.run(
-                        data,
+               case data do
+                 %struct{results: results} = page
+                 when struct in [Ash.Page.Offset, Ash.Page.Keyset] ->
+                   # `run` already built the page, so only its records need
+                   # loading. The page keeps its `more?` and `count`.
+                   with {:ok, results} <-
+                          load_data_layer_results(
+                            results,
+                            query_ran,
+                            query,
+                            initial_query,
+                            calculations_at_runtime,
+                            calculations_in_query,
+                            opts
+                          ) do
+                     {:ok, %{page | results: results}}
+                   end
+
+                 results when is_list(results) ->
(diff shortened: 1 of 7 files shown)


zachdaniel:
@barnabasJ yes it does change the result but not the contract, so I'm comfortable with the change to make `run` return a page. It was always meant to and the type spec is clear.

barnabasJ:
@zachdaniel done in 1bdfdb2: `run` now drops the extra row before the hooks and returns the `Ash.Page.Offset`/`Ash.Page.Keyset` for paginated queries (list for unpaginated ones), and `load` accepts that page, loads its records and returns the page. `load` still pages raw rows for anyone who ran the data layer query themselves.

Three things I hit along the way that you may want to eyeball, since they change observable behaviour of that escape hatch:

- `count` was being passed to the page as the unresolved closure, so `page.count` was a function. It is now resolved inside `run`.
- Pages store their read opts for `Ash.page/2` reruns, and those opts carried `data_layer_query?: true`, so paging from such a page returned another query map instead of a page. `run` now strips that option before building the page.
- `load` returned the bare list rather than `{:ok, list}`. It now returns `{:ok, _}` as its typespec has always said.

The PR description's "open question" section is now moot; I've updated the `Ash.data_layer_query/2` docs to describe the new behaviour.

lib/ash/actions/read/read.ex
```diff
@@ -1068,78 +1069,62 @@ defmodule Ash.Actions.Read do
...
-                      ),
-                    {:ok, data} <-
-                      load_relationships(data, query, opts),
-                    {:ok, data} <-
-                      Ash.Actions.Read.Calculations.run(
-                        data,
+               case data do
+                 %struct{results: results} = page
+                 when struct in [Ash.Page.Offset, Ash.Page.Keyset] ->
+                   # `run` already built the page, so only its records need
```
zachdaniel, line 1084: fair few claude-isms in the comments, can we trim these/human-ify them?

zachdaniel: Otherwise this looks good :thumbs

barnabasJ: Ups, sorry. Yeah I was planning on cleaning it up a bit, didn't catch that Claude already pushed the data_layer_query changes.

zachdaniel:

PR #2918, opened 2026-09-09
perf: check period overlap against one record's versions in the ETS data layer

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

# Summary

Improves a triangular performance issue in ETS data layer when checking period overlap, for batch sizes > 1.

A temporal resource may have many versions of the one record, each over a different period, and they mustn't overlap. As `bulk_create/3` may contain multiple versions of the same record we need to ensure there is no overlap within these, and no overlap with these and versions of the same record already in the store.

The improvement is for `bulk_create/3` to make one pass over the store, then one pass over the records.

### DIFF
--- a/lib/ash/data_layer/ets/ets.ex
+++ b/lib/ash/data_layer/ets/ets.ex
@@ -1710,35 +1710,30 @@ defmodule Ash.DataLayer.Ets do
         other -> other
       end
     else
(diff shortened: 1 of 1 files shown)


zachdaniel:

PR #2920, opened 2026-09-09
chore: refactor ash.set.domains to igniter/1

### LINKED ISSUE #2919: ash.set.domains still defines the deprecated igniter/2
`mix.tasks.ash.set.domains` defines `igniter/2`, which igniter 0.8.4 deprecated. Any project that has both `ash` and `igniter` in its dependency tree gets a warning on every compile of `ash`.

```
[warning] Mix.Tasks.Ash.Set.Domains (deps/ash/lib/mix/tasks/ash.set.domains.ex):

    Module defines `igniter/2`, but `igniter/2` is deprecated.
    Please refactor to `igniter/1`. Use `igniter.args` to access `argv`,
    `positional` and `options`.
```

**Source:** `lib/mix/tasks/ash.set.domains.ex:21` — `def igniter(igniter, _argv) do`

The deprecation is raised by igniter itself at `lib/mix/task.ex:175`.

**Versions:** `ash` 3.33.1 (current latest), `igniter` 0.8.4, Elixir 1.18.4 / OTP 27.

Still present on `main` as of today — I checked `raw.githubusercontent.com/ash-project/ash/main/lib/mix/tasks/ash.set.domains.ex`, same line.

**Scope:** this is the only task in `lib/mix/tasks/` still on the two-arity form; everything else already uses `igniter/1`.

**Impact is cosmetic** — it is a warning, not a failure, and it does not fail a consuming project's `mix compile --warnings-as-errors` since it comes from dependency compilation. It is just noise on every fresh build for anyone using both packages, which is now common since `igniter` is the standard install path for the Ash ecosystem.

The fix looks like the same mechanical change already applied elsewhere: `def igniter(igniter)` and read `argv` from `igniter.args` if it is needed (it is currently discarded as `_argv`, so it may not be).

Happy to send a PR if that is useful.

pshoukry:
Closes #2919.

`Mix.Tasks.Ash.Set.Domains` defines `igniter/2`, which igniter 0.8.4 deprecates. Any project with both `ash` and `igniter` in its dependency tree gets a warning on every compile of `ash`:

```
[warning] Mix.Tasks.Ash.Set.Domains (deps/ash/lib/mix/tasks/ash.set.domains.ex):
    Module defines `igniter/2`, but `igniter/2` is deprecated.
    Please refactor to `igniter/1`.
```

The argument was already discarded as `_argv`, so nothing in the task reads it and the one-arity form is a direct replacement — no `igniter.args` access needed.

This is the only task under `lib/mix/tasks/` still on the two-arity form.

## Verification

- `mix compile --force` on this branch: no `Ash.Set.Domains` warning (was present before the change).
- Behaviour is unchanged; the body is untouched.

Tested against igniter 0.8.4, Elixir 1.18.4 / OTP 27.

### DIFF
--- a/lib/mix/tasks/ash.set.domains.ex
+++ b/lib/mix/tasks/ash.set.domains.ex
@@ -18,7 +18,7 @@ if Code.ensure_loaded?(Igniter) do
     """
 
     @impl Igniter.Mix.Task
-    def igniter(igniter, _argv) do
+    def igniter(igniter) do
       app_name = Igniter.Project.Application.app_name(igniter)
 
       Mix.shell().info("🔍 Scanning for Ash domains in :#{app_name}...")


zachdaniel:

pshoukry:
Thank you for Ash 🙇‍♂️

PR #2922, opened 2026-09-10
fix: refuse a temporal read or write the data layer cannot serve.

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

The verifier `ValidateTemporal` requires `can?(:temporal, ..)` for a given resource to compile, however the checks at runtime are bugged or missing.

This is a PR without an issue as currently temporal data layers must claim `can?(:temporal, ..)` `true` otherwise `as_of(instant)` is silently dropped.

The fix checks the datalayer's capability ahead of all reads and writes, granting the datalayer the ability to safely refuse temporal for instance when the underlying database lacks temporal capability or configuration.

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -2291,7 +2291,7 @@ defmodule Ash.Changeset do
           message:
             "Could not determine domain for changeset. Provide the `domain` option or configure a domain in the resource directly."
 
-    changeset = %{changeset | domain: domain}
+    changeset = refuse_unserved_temporal(%{changeset | domain: domain})
 
     if changeset.valid? do
       try do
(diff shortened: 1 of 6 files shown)


matt-beanland:
amended to generalise some comments and error messages

lib/ash/changeset/changeset.ex
```diff
@@ -2951,6 +2951,22 @@ defmodule Ash.Changeset do
     end)
   end
 
+  # The runtime counterpart of `Ash.Resource.Verifiers.ValidateTemporal`.
+  defp refuse_unserved_temporal(changeset) do
```
zachdaniel, line 2955: why do we need a runtime component of this?

zachdaniel: How could you get here if the app won't compile otherwise?

matt-beanland: Distributed/embedded systems where code compiled on one system and running on another.

Also where db version / temporal capability is runtime configuration or license gated.

zachdaniel: Where does that happen in practice? It would mean that *all* things that we validate at compile time have to also be validated at runtime,  which is not a position we hold.

matt-beanland: Yes with PG needs PG19 and also btree_gist configuration, allows gate to check. Depends on people's deployment scenarios I guess.

I understand wider implications for other capability checks and that you may prefer to just let the data layer overclaim in this scenario rather than establish a precedent.

zachdaniel: Yes but the user configures those things in their repo with `min_pg_version` and `extensions`. So we know this information at compile time.

PR #2925, opened 2026-09-10
fix: setting undeclared argument was done with `nil` key

Rutgerdj:
Setting an argument on the changeset that is not defined by the action put the value in the arguments map with a `nil` key. 
This was caused by the `argument` variable being reassigned 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/ash/changeset/changeset.ex
+++ b/lib/ash/changeset/changeset.ex
@@ -6641,8 +6641,15 @@ defmodule Ash.Changeset do
             |> store_casted_argument(action_argument.name, last_val, store_casted?)
         end
       else
-        %{changeset | arguments: Map.put(changeset.arguments, argument, value)}
-        |> store_casted_argument(argument, value, store_casted?)
+        add_error(
+          changeset,
+          NoSuchInput.exception(
+            resource: changeset.resource,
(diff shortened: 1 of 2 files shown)


lib/ash/changeset/changeset.ex
```diff
@@ -6612,29 +6612,33 @@ defmodule Ash.Changeset do
 
   defp do_set_argument(changeset, argument, value, store_casted? \\ false) do
```
zachdaniel, line 6613: 🤔 I didn't even realize we stored undefined arguments. We shouldn't be putting these into the arguments at all. Could we make that change instead?

Rutgerdj: Good point, I just thought this was undocumented bahaviour :D

I've updated the logic to attach an error to the Changeset. Thats a breaking change tho. Before it failed silently, now it fails loudly. 

That causes a Reactor test (test/reactor/change_test.exs:44) to fail. 
I've had our shared friend take a look at it but couldn't confidently review it's suggestions so leaving this open.

zachdaniel:

PR #2928, opened 2026-09-12
refactor: move the as_of algebra into Ash.Temporal

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

# Summary

The `as_of` algebra is implemented in more than one place, this PR centralizes the better implementation to `Ash.Temporal` and has the ETS data layer call it. This temporal code doesn't belong on `Ash.Range` while used for temporal is and should remain non-temporal.

This PR is a pure refactor, some fixes and improvements to follow.

# Tests

11 new tests
`mix test` 4216 passing, `mix credo --strict` clean, `mix format --check-formatted` clean, `mix compile --warnings-as-errors` clean.

Two inherited failures in `test/mix/tasks/ash.set.domains_test.exs`

### DIFF
--- a/lib/ash/query/query.ex
+++ b/lib/ash/query/query.ex
@@ -3123,10 +3123,6 @@ defmodule Ash.Query do
     |> set_context(%{shared: %{as_of: as_of}})
   end
 
-  @doc false
-  # Kept so that data layers calling it keep compiling; `Ash.Temporal` is where it lives.
-  defdelegate resolve_as_of(as_of), to: Ash.Temporal
-
   # Apply an `as_of` from opts without clobbering an unset query when there is none.
   defp maybe_set_as_of(query, nil), do: query
   defp maybe_set_as_of(query, value), do: as_of(query, value)
--- a/lib/ash/query/query.ex
+++ b/lib/ash/query/query.ex
@@ -3124,10 +3124,8 @@ defmodule Ash.Query do
   end
 
   @doc false
-  # `:now` means "current time at execution"; resolve it where as_of is consumed
-  # as a concrete value. A `DateTime` fixes the instant; `nil` means unset.
-  def resolve_as_of(:now), do: DateTime.utc_now()
-  def resolve_as_of(other), do: other
+  # Kept so that data layers calling it keep compiling; `Ash.Temporal` is where it lives.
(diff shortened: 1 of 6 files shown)


lib/ash/query/query.ex
```diff
@@ -3124,10 +3124,8 @@ defmodule Ash.Query do
   end
 
   @doc false
-  # `:now` means "current time at execution"; resolve it where as_of is consumed
-  # as a concrete value. A `DateTime` fixes the instant; `nil` means unset.
-  def resolve_as_of(:now), do: DateTime.utc_now()
-  def resolve_as_of(other), do: other
+  # Kept so that data layers calling it keep compiling; `Ash.Temporal` is where it lives.
+  defdelegate resolve_as_of(as_of), to: Ash.Temporal
```
zachdaniel: We don't need to keep this function around

lib/ash/temporal.ex
```diff
@@ -0,0 +1,93 @@
+# SPDX-FileCopyrightText: 2019 ash contributors <https://github.com/ash-project/ash/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Ash.Temporal do
+  @moduledoc """
+  The `as_of` algebra: the instant a read sees, and the period a write establishes.
```
zachdaniel: These docs are fairly LLM-ish. Can we write them with a bit more focus on less semantic density and easier understanding from a layperson?

zachdaniel (changes requested):

lib/ash/temporal.ex
matt-beanland: I've re-written these

lib/ash/query/query.ex
matt-beanland: agreed - has been removed

matt-beanland:
@zachdaniel thanks for these comments, have addressed each in a commit
now 10 tests as removed defdelegate one.
mix test 4215 passing, mix credo --strict, mix format --check-formatted and mix compile --warnings-as-errors all clean.
I'll raise a PR for ash_postgres temporal to rework to use the `Ash.Temporal` surface.

zachdaniel:

PR #2929, opened 2026-09-13
fix: don't drop notifications returned as a list from manual destroy actions

Torkan:
A manual destroy returning `{:ok, record, [notification]}` passed validation but never matched the `%{notifications: ...}` clause in `manage_relationships/4`, so all notifications for the destroy were silently dropped. Normalise the list to the map shape. Adds a regression test.


zachdaniel:

PR #2930, opened 2026-09-13
fix: unwrap map-form notifications from a non-bulk manual action result

Torkan:
Manual create/update modules return `{:ok, record, %{notifications: [...]}}`, but when driven by a bulk action without a bulk callback, `process_non_bulk_result/6` passed the map straight to `store_notification/3`, which stored it as a single notification. Unwrap it like `process_bulk_results/9` already does. Adds a regression test.


zachdaniel:

PR #2931, opened 2026-09-14
fix:
  don't ask whether a resource can join itself

jeroen11dijk:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Ash.Filter.validate_not_crossing_data_layer_boundaries/3 (lib/ash/filter/filter.ex:3680) gets all Resource types but doesnt dedupe them, which then makes it skip the fast path for the check. But it then also checks if two resources can join itself which I got to fail on my own repo with a calculation that looked like:

``expr(address.street <> " " <> address.city)``


zachdaniel:

PR #2932, opened 2026-09-14
test: repro for field policies not filtering calculations as nil

jsw800:
See discussion [here](https://discord.com/channels/711271361523351632/1059965824691097670/1549167948311035958)

seems like calculations and aggregates that are forbidden by field policies are not nulled out when referred to by filters.


zachdaniel:
Fixed in b3d4503241f3deacb5ceb955e10a4fa927da0f67

PR #2934, opened 2026-09-14
fix: reject the unsupported `return_query?` option in `Ash.stream!/2`

### LINKED ISSUE #2933: `return_query?` crashes `Ash.stream!/2`
**Versions:** ash 3.32.3 and 3.33.3, Elixir 1.20.0 and 1.20.2

**Operating system:** macOS (Darwin 25.6.0)

### Current Behavior

`Ash.stream!(Resource, return_query?: true) |> Enum.to_list()` crashes.

`return_query?` is in the `Ash.stream!/2` options schema, because `@stream_opts` merges the read schema and drops only `:page` (`lib/ash.ex:195`). So the option is accepted, then passed through to the `Ash.read!/2` call in each batch. That call returns `{results, query}`, which none of the four strategies in `Ash.Actions.Read.Stream` match on:

- keyset and offset pagination → `CaseClauseError`
- limit/offset and full read → `Protocol.UndefinedError`

```
** (CaseClauseError) no case clause matching: {#Ash.Page.Offset<...>, #Ash.Query<...>}
    (ash 3.33.3) lib/ash/actions/read/stream.ex:55: anonymous fn/5 in Ash.Actions.Read.Stream.do_stream/4
    (elixir 1.20.2) lib/stream.ex:1591: Stream.do_resource/5
    (elixir 1.20.2) lib/enum.ex:4667: Enum.reverse/1
    (elixir 1.20.2) lib/enum.ex:3926: Enum.to_list/1
```

Same family as #1667, which missed the stream path.

### Reproduction

Failing test in the linked PR.

### Expected Behavior

Either the option works, or it is rejected with a clear error. A stream runs one read per batch, each with different page options, so there is no single query to return — I think rejecting it is right.

jimsynz:
Fixes #2933.

`@stream_opts` merged the read options schema and dropped only `:page`, so `return_query?` was accepted by `Ash.stream!/2` and passed through to the `Ash.read!/2` call in each batch. That call returned `{results, query}`, which none of the four stream strategies match on.

A stream runs one read per batch, each with different page options, so there is no single query to return. This drops `:return_query?` from the schema next to `:page`, so the option is rejected with a validation error that names it.

One thing worth a second opinion: `@bulk_update_opts_schema` and `@bulk_destroy_opts_schema` both merge the stream options, so this removes `return_query?` from `Ash.bulk_update/4` and `Ash.bulk_destroy/4` too. Nothing read the option on those paths — the only consumer is `Ash.Actions.Read.add_query/3` — so it was already a silent no-op there. But anyone passing it explicitly to a bulk action now gets a validation error. Happy to scope the change to `Ash.stream!/2` alone if you would rather not touch the bulk schemas.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


Torkan (approved):

zachdaniel:

PR #2943, opened 2026-09-16
feat: accept a range-valued as_of, naming the period a write applies to

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Supports an `as_of` being a range or instant, so a write can name the period it applies to rather than only the instant it begins at, and a read uses the lower bound of the range. Targets ash `temporal` with the `Ash.Temporal` surface.

# Detail

## as_of(range) in Changeset:

```elixir
# Today: the edit runs to the row's own upper bound.
Ash.update!(tenancy, %{seats: 20}, as_of: ~U[2026-03-01 00:00:00Z])

# With this: it names the portion, and the prior value resumes after it.
Ash.update!(tenancy, %{seats: 20},
  as_of: %Ash.Range{lower: ~U[2026-03-01 00:00:00Z], upper: ~U[2026-06-01 00:00:00Z]})
```

| action | what a range means |
| --- | --- |
| create | the range **is** the period |
| update | the portion to carve; the prior version keeps the time either side |
| destroy | ends validity over that portion, and it resumes after |

## as_of(range) in Query:

A range in a query resolves to its lower bound.

## Refusals

A range with no lower bound is refused as `Ash.Error.Query.AsOfNotAnInstant`, `nil` already means "no particular instant", so such a range must not quietly read as current state.

An update whose range falls outside the version it acts on is refused as `Ash.Error.Changes.PeriodOutOfBounds` rather than clipped to the intersection.

# Tests

New tests in `test/ash/data_layer/ets_temporal_test.exs`.

4251/4253. The two failures are `Mix.Tasks.Ash.Set.DomainsTest` and fail identically on `temporal` without this branch, and were verified on a clean worktree of the base.


lib/ash/query/query.ex
```diff
@@ -2844,11 +2844,23 @@ defmodule Ash.Query do
...
+  # A write's `as_of` reaches its own read leg through this context, so a range arrives here
+  # even when no caller passed one.
+  defp narrow_as_of(query, %Ash.Range{lower: nil} = as_of) do
+    add_error(
+      query,
+      Ash.Error.Query.AsOfNotAnInstant.exception(resource: query.resource, as_of: as_of)
+    )
+  end
+
+  defp narrow_as_of(query, %Ash.Range{lower: lower}), do: %{query | as_of: lower}
```
zachdaniel, line 2861: I think we should actually just forbid this. Allowing `as_of: %Range{}` for queries but then *ignoring* the range part of it has two drawbacks:

1. its confusing as its just ignoring part of the input
2. if we ever *do* decide to accept ranged reads and return all instances that fall within that range, we won't be able to without it being a breaking change.

So we should only accept a range in `as_of` for change sets.

zachdaniel (changes requested):

lib/ash/query/query.ex
matt-beanland: Yes I agree have been slowly coming to same conclusion, will address.

matt-beanland:
I've  addressed and changed the `Ash.Temporal` surface so I'll follow on with a chore PR on ash_postgres.

`Ash.Temporal.resolve_as_of/1` is split into `resolve_write_as_of/1`, which resolves a range to its lower bound, and `resolve_read_as_of/1`, which raises `AsOfNotAnInstant` if it gets a a Range.   Note that write's `as_of` reaches its own read legs through the shared context, so `Ash.Changeset.as_of/2` now shares the resolved instant and keeps the range on `private`. Without that the carve's own reload gets refused by the new rule.

I've got the ash_postgres PR ready to raise and have proven together locally.

ash 4253/4255 with the two pre-existing `Mix.Tasks.Ash.Set.DomainsTest` failures.

matt-beanland:
@zachdaniel added a fourth commit, a fix to relative time expressions with a test.
ash 4254/4256 locally (two pre-existing `DomainsTest`), ash_postgres 1039/1039 via path dep.

lib/ash/query/query.ex
```diff
@@ -829,7 +829,7 @@ defmodule Ash.Query do
       doc: "set the tenant on the query"
     ],
     as_of: [
-      type: {:or, [{:struct, DateTime}, {:literal, :now}, {:literal, nil}]},
+      type: {:or, [{:struct, DateTime}, {:struct, Ash.Range}, {:literal, :now}, {:literal, nil}]},
```
zachdaniel: I think there is no need to accept or type this as a range given that we won't accept it right?

matt-beanland: Yes looks dead code

matt-beanland:
@zachdaniel added a fifth commit, dropping `Ash.Range` also from `Ash.Query.t()`, `Ash.get/3` and `Ash.calculate/3`.

ash 4254/4256 locally (two pre-existing `DomainsTest`), ash_postgres 1039/1039 via path dep.

lib/ash/query/query.ex
```diff
@@ -2844,11 +2844,20 @@ defmodule Ash.Query do
     # (the same channel multitenancy uses), and is mirrored onto the struct
     # field here so it is threaded to the data layer.
     case Map.fetch(query.context, :as_of) do
-      {:ok, as_of} -> %{query | as_of: as_of}
+      {:ok, as_of} -> put_as_of(query, as_of)
       :error -> query
     end
   end
 
+  defp put_as_of(query, %Ash.Range{} = as_of) do
```
zachdaniel: Still seeing code referencing ranges in queries 🤔

matt-beanland: yes this one is so you get a sensible error, rather than a raise on `Ash.read` or silence on `Ash.Query.as_of/2`.

zachdaniel: but how can you do it? Nothing should accept actually *setting* a range there right?

matt-beanland: Range could reach a query because `@global_opts`: `as_of` is declared and widening it for change sets put a range into `@read_opts_schema` and `@aggregate_opts` too , which meant  `Ash.read/2` plus `count`, `exists?`, `first`, `sum`, `list`, `min`, `max`, `avg` and `aggregate` all silently accepting one.

Solution is that both now declare their own 'as_of', and merge `Keyword.delete(@global_opts,
:as_of)`, the same shape the write schemas use for `:action`. Most read surfaces now refuse at the option; creates, updates and destroys are unchanged. `list` works a little differently: it applies `distinct`/`filter` before validating its opts, so it is refused by `Ash.Temporal` a moment later.

This addresses your issue with `Ash.Query` which now has no reference to`Ash.Range`. Anything reaching a query past the spec is refused by `Ash.Temporal.resolve_read_as_of/1`.

ash 4254/4256 (two pre-existing `DomainsTest`), ash_postgres 1043/1043 via
path dep.

matt-beanland:
@zachdaniel Added a sixth commit, details above.

zachdaniel (approved):

zachdaniel:

PR #2945, opened 2026-09-16
fix: reject predicate functions on attributes of an incompatible type

### LINKED ISSUE #2939: Input filters do not check a predicate's argument types against the referenced attribute
Versions: ash 3.33.3, Elixir 1.20.1 on OTP 29. Present on the default branch at `a7a5105` (checked 2026-09-15).

Found by an AI assistant working with a human who was fuzz-testing their own application; reproduced in a minimal project with default settings.

## Request

On a `:string` attribute: `Post |> Ash.Query.filter_input(%{"title" => %{"range_overlaps" => "x"}}) |> Ash.read()`, also `range_adjacent` and `range_contains`. On an `:integer` attribute: `filter_input(Post, %{"score" => %{"contains" => "x"}})`. Through AshJsonApi: `GET /posts?filter[title][range_overlaps]=x`. Through AshGraphql: `listPosts(filter: {title: {rangeOverlaps: "x"}})`.

## Observed

The filter parses and reaches Postgres:
```
{:error, %Ash.Error.Unknown{errors: [%Ash.Error.Unknown.UnknownError{error:
  "** (Postgrex.Error) ERROR 42883 (undefined_function) operator does not exist: text && unknown ..."}]}}
```
(`-|-` for adjacent, `@>` for contains, `bigint ~~ unknown` for `contains` on the integer.) JSON:API answers 500; GraphQL answers "Something went wrong".

## Expected

An invalid filter error when the filter is parsed.

## Failure point

`Ash.Filter.get_predicate_function/3`, [lib/ash/filter/filter.ex#L5147-L5157](https://github.com/ash-project/ash/blob/v3.33.3/lib/ash/filter/filter.ex#L5147-L5157), and `Ash.Query.Function.try_cast_arguments/2`, [lib/ash/query/function/function.ex#L146-L190](https://github.com/ash-project/ash/blob/v3.33.3/lib/ash/query/function/function.ex#L146-L190); the range predicates' `args/0` in [range_overlaps.ex#L19](https://github.com/ash-project/ash/blob/v3.33.3/lib/ash/query/function/range_overlaps.ex#L19), [range_adjacent.ex#L18](https://github.com/ash-project/ash/blob/v3.33.3/lib/ash/query/function/range_adjacent.ex#L18), [range_contains.ex#L18](https://github.com/ash-project/ash/blob/v3.33.3/lib/ash/query/function/range_contains.ex#L18).

## Notes

The three range predicates declare their argument types as `[:any, :same]` (`[:any, :any]` for `range_contains`). AshJsonApi's OpenAPI filter schema and AshGraphql's filter input types build their per-field predicate lists from those declarations, so both list the range predicates on every field, including text; the GraphQL introspection test in the repository shows that and is tagged with this report.

## Reproduction

* Description, with the failure point at this release: https://github.com/grempe/ash-fuzz-repros#ashpredicate-argument-types-unchecked
* Failing test: https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash/predicate_argument_types_unchecked_test.exs
* Run: `mix setup && mix test test/ash/predicate_argument_types_unchecked_test.exs` in https://github.com/grempe/ash-fuzz-repros

grempe:
## Contributor checklist

- [x] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Closes #2939

## Problem

`Ash.Query.Function.try_cast_arguments/2` never compares a `%Ash.Query.Ref{}` with the declared argument type: every clause either accepts a vague declared type (`:any`, `:same`) without looking at the value, or short-circuits on `Ash.Expr.expr?/1` before `Ash.Query.Type.try_cast/3`, and `expr?/1` is true for every reference. `Function.new/2` is the only argument validation a predicate function gets on all three parse paths (`parse_predicates/3` for input filters, `add_expression_part/4` for keyword filters, `resolve_call/2` for `Ash.Query.filter/2` expressions), so any predicate function is accepted on any attribute:

```elixir
Post |> Ash.Query.filter_input(%{"score" => %{"contains" => "x"}}) |> Ash.read()
# {:error, %Ash.Error.Unknown{}}  Postgrex: operator does not exist: bigint ~~ unknown
Post |> Ash.Query.filter_input(%{"title" => %{"range_overlaps" => "x"}}) |> Ash.read()
# {:error, %Ash.Error.Unknown{}}  Postgrex: operator does not exist: text && unknown
```

This is not specific to `:same`/composite declarations: `contains`, `string_starts_with` and `string_ends_with` are declared `[:string, :string]` and are still accepted on an `:integer` attribute; `has` and `intersects` are accepted on a scalar `:string`. The only place that compares a reference's type to a declaration is `Ash.Expr.determine_types/4`, which is called by the SQL compiler and answers a mismatch with `{[nil, nil], nil}` rather than an error, so the data layer compiles the operands untyped and the database is the first thing that objects. The `expr?` short-circuit is as old as the function (it skipped `%Ref{}` by name from its first commit in 2021); what made it reachable was resolving predicate functions by name in `parse_predicates/3` (v2.19.0) and, for the API extensions, the range predicates' `[:any, :same]` declarations (v3.32.0).

A second, smaller bug hid the little validation that exists: when `try_cast_arguments/2` does fail for a literal inside an input filter (`%{"title" => %{"contains" => %{}}}`), the `with` in `parse_predicates/3` has no `else`, the `{:error, msg}` escapes the `Enum.reduce_while/3` callback, and `Ash.Query.filter_input/2` raises `FunctionClauseError`.

## Fix

- `try_cast_arguments/3` takes `check_refs?`. `Function.new/2` passes `mod.predicate?()`, so for predicate functions a signature no longer fits a reference whose attribute type is incompatible with a concrete declared type. Compatible means the same type, a `Ash.Type.NewType` of it, or the same storage type (an `:atom` attribute still satisfies `:string`). A reference without a determinable type is left alone. Non-predicate functions (`round`, `rem`, `string_length`, ...) are deliberately unchanged because their SQL relies on data layer coercion.
- `Function.new/2` returns an `Ash.Error.Query.InvalidFilterValue` (class `:invalid`) instead of a bare string when no signature fits; the message names the reference's type. The bare string used to become an `UnknownError`, a 500 in AshJsonApi.
- `range_overlaps`, `range_adjacent` and `range_contains` keep their vague declarations (declaring the first argument as `:range` would stop `determine_types/4` from typing the range literal, because a `:same` slot only adopts a basis from an `:any` slot) and check in `new/1` that a referenced attribute is a range.
- `parse_predicates/3` propagates a `Function.new/2` error instead of crashing.
- `test/filter/predicate_argument_types_test.exs` covers wrong-typed and compatible attributes (string, NewType, `ci_string`, `atom`, `range`, list), the `Ash.Query.filter/2` path, the crash, and the `check_refs?` option.

Behaviour change worth a changelog line: `Ash.Query.filter(Post, contains(score, "x"))` now returns an invalid query at parse time instead of failing in the data layer.

## Verification

- `mix test`: 4120/4122 on v3.33.3 and on `main`, the same two pre-existing `Mix.Tasks.Ash.Set.DomainsTest` failures as without the patch.
- `mix format --check-formatted`, `mix spark.formatter --check`, `mix credo --strict`, `mix dialyzer`, `mix sobelow`, `mix git_ops.check_message`: clean.
- With the fix applied to `deps/ash` of the reproduction project in https://github.com/grempe/ash-fuzz-repros, `test/ash/predicate_argument_types_unchecked_test.exs` goes from 1/7 to 6/7: the four `filter_input` cases and the AshJsonApi 400 pass. The remaining GraphQL case needs `AshGraphql.Error` to be implemented for `Ash.Error.Query.InvalidFilterValue` (it is not today, so it renders "Something went wrong", as it already does for `filter: {score: {in: ["abc"]}}`).

Not addressed here: AshGraphql and AshJsonApi still list the range predicates on every field, because their filter-type generators treat any `[:any, _]` signature as universal.


zachdaniel:
The real fix for this is more systemic/requires properly extending the type system to cover this. I shared a branch on the issue report. Thank you for your contribution!

PR #2946, opened 2026-09-16
fix: preserve field paths and messages in embedded array errors

Alt-iOS:
Normalize raw casting and constraint errors in the simple embedded-array fast path using the same exception options as ordinary changesets. Preserve the leaf field, row and nested paths, message template, interpolation variables, and original value without routing successful casts through the full lifecycle.

Keep existing exception semantics and fail-fast behavior. Add regression coverage for scalar and nested map errors, parent attributes and arguments, required fields, multiple errors, and prebuilt exceptions.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #2947, opened 2026-09-16
fix: return an invalid error for sort input that is not a string or list

### LINKED ISSUE #2941: sort_input crashes on a value that is neither a string nor a list
Versions: ash 3.33.3, Elixir 1.20.1 on OTP 29. Present on the default branch at `a7a5105` (checked 2026-09-15).

Found by an AI assistant working with a human who was fuzz-testing their own application; reproduced in a minimal project with default settings.

## Request

`Ash.Query.sort_input(Post, 5)`; also `%{}`, `%{"title" => "asc"}` and `[%{"field" => "title"}]`.

## Observed

```
** (FunctionClauseError) no function clause matching in Ash.Sort.parse_input/3
```
For the list of maps: `no function clause matching in Ash.Resource.Info.attribute/2`. By contrast `Ash.Query.filter_input(Post, "title")` puts an `InvalidFilterValue` on the query.

## Expected

An invalid-class error on the query. A map is not a documented sort format; the request is an invalid-sort error, not map support.

## Failure point

`Ash.Sort.parse_input/3`, [lib/ash/sort/sort.ex#L94-L126](https://github.com/ash-project/ash/blob/v3.33.3/lib/ash/sort/sort.ex#L94-L126).

## Reproduction

* Description, with the failure point at this release: https://github.com/grempe/ash-fuzz-repros#ashsort-input-shape
* Failing test: https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash/sort_input_shape_test.exs
* Run: `mix setup && mix test test/ash/sort_input_shape_test.exs` in https://github.com/grempe/ash-fuzz-repros

chiliec:
Closes #2941.

`Ash.Sort.parse_input/3` has clauses for a string, an atom, a list and `nil`, so any other value (`5`, `%{}`, `%{"title" => "asc"}`) raises `FunctionClauseError`. A list is accepted, but an element that is neither a field name nor a `{field, order}` tuple (e.g. `[%{"field" => "title"}]`) falls through `parse_sort/4` into `Ash.Resource.Info.attribute/2`, which raises the same way. `Ash.Query.filter_input/2` puts an `InvalidFilterValue` on the query for a malformed shape; `sort_input/2` now does the equivalent.

- New `Ash.Error.Query.InvalidSort` (class `:invalid`, field `:sort`), alongside the existing `InvalidSortOrder`.
- `parse_input/3` gets a final clause returning `{:error, %InvalidSort{}}`.
- The catch-all `parse_sort/4` clause is guarded to `binary | atom | struct` (the shapes `get_field/5` handles), with a final clause returning `InvalidSort` for anything else. `Ash.Query.sort_input/2` then adds that error to the query instead of raising.

Tests: `test/sort/sort_test.exs` covers `sort_input` with `5`, `%{}`, `%{"title" => "asc"}` and `[%{"field" => "title"}]`, and `parse_input` with `5` and a map. On `main` the four `sort_input` cases raise `FunctionClauseError` (`Ash.Sort.parse_input/3` / `Ash.Resource.Info.attribute/2`); with the change they pass. Full suite: 4149/4150 — the one failure (`Mix.Tasks.Ash.Gen.ResourceTest` "generates resource with custom base") fails identically on `main` without this change. `mix format --check-formatted` and `mix credo --strict` clean on the touched files.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


grempe:
Thanks for picking this up. I ran the reproduction suite from https://github.com/grempe/ash-fuzz-repros with `ash` pinned to f43fd6a. The four `sort_input` cases in https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash/sort_input_shape_test.exs (`5`, `%{}`, `%{"title" => "asc"}` and `[%{"field" => "title"}]`) now put an `InvalidSort` error on the query instead of raising, and the control still passes. As a side effect the ash_lua `sort = 5` case in that suite, which was surfacing the same `FunctionClauseError` through the script boundary, also passes. Nothing else in the suite changed.

zachdaniel:

Issue #1971, opened 2025-04-13
Defining `get_by` on a code interface doesn't do type-checking of the argument


sevenseacat:
**Describe the bug**

A code interface to get a single record by UUID primary key:

```elixir
define :get_artist_by_id, action: :read, get_by: :id
```

Calling the generated function with an invalid UUID raises an error from within Ecto:

```
iex(8)> Tunez.Music.get_artist_by_id("not a uuid")
{:error,
 %Ash.Error.Invalid{
   bread_crumbs: ["Error returned from: Tunez.Music.Artist.read"],
   query: "#Query<>",
   errors: [
     %Ash.Error.Query.InvalidFilterValue{
       message: nil,
       value: "not a uuid",
       context: #Ecto.Query<from a0 in Tunez.Music.Artist, as: 0,
        where: type(
  type(as(0).id, {:parameterized, {Ash.Type.UUID.EctoType, []}}),
  {:parameterized, {Ash.Type.UUID.EctoType, []}}
) ==
  type(
    type(^"not a uuid", {:parameterized, {Ash.Type.UUID.EctoType, []}}),
    {:parameterized, {Ash.Type.UUID.EctoType, []}}
  ),
        select: struct(a0, [
  :id,
  :name,
  :inserted_at,
  :updated_at,
  :biography,
  :created_by_id,
  :previous_names,
  :updated_by_id
])>,
       splode: Ash.Error,
       bread_crumbs: ["Error returned from: Tunez.Music.Artist.read"],
       vars: [],
       path: [],
       stacktrace: #Splode.Stacktrace<>,
       class: :invalid
     }
   ]
 }}
```

Whereas if the code interface used a real action that applied a `get_by` filter:

```elixir
# in the domain 
define :get_artist_by_id, action: :get_by_id, args: [:id]

# in the resource
actions do
  read :get_by_id do
    get_by :id
  end
end
```

Calling the function would properly validate the UUID before running the action:

```
iex(5)> Tunez.Music.get_artist_by_id("not a uuid")
{:error,
 %Ash.Error.Invalid{
   errors: [
     %Ash.Error.Query.InvalidArgument{
       field: :id,
       message: "is invalid",
       value: "not a uuid",
       splode: Ash.Error,
       bread_crumbs: [],
       vars: [],
       path: [],
       stacktrace: #Splode.Stacktrace<>,
       class: :invalid
     }
   ]
 }}
```

[...]

PR #2955, opened 2026-09-20
test: cover get_by input casting on update/destroy code interfaces

lukegalea:
## Problem

#1971 — *"Defining `get_by` on a code interface doesn't do type-checking of the argument"*:

```elixir
define :get_artist_by_id, action: :read, get_by: :id
# Tunez.Music.get_artist_by_id("not a uuid")
```

surfaced an `Ash.Error.Query.InvalidFilterValue` (with the whole Ecto query embedded in
the error), while the equivalent read action with `get_by` returns a clean
`Ash.Error.Query.InvalidArgument{field: :id, message: "is invalid"}` — because the
action-level `get_by` transformer adds the field as a typed argument, and the code
interface path skipped that casting.

## Approach

Per @zachdaniel's direction in the issue thread: look up each `get_by` field
(`Ash.Resource.Info.field/2`), extract `field.type` / `field.constraints`, and validate
the input with `Ash.Type.cast_input/3`.

**Current `main` already implements exactly this** via `Ash.CodeInterface.apply_get_by_filter`,
`apply_get_by_filter_to_resource` and `cast_get_by_field`, wired into the read, update and
destroy code-interface paths — casting like an action argument, rejecting relationship
keys, and producing `InvalidArgument` errors.

I independently implemented the same fix against a stale fork base; when rebasing onto
current `main`, every hunk of my library change conflicted with (and was strictly subsumed
by) main's implementation, so the cherry-pick was resolved entirely in favor of main's
code.

## What this PR contains

**Test coverage only.** The existing `code interface get_by validation` test covers the
read path; these lock in the rest of the issue's scenarios, all passing unmodified against
main's implementation:

- update interfaces with `get_by` — invalid value errors with `InvalidArgument`, valid value succeeds
- destroy interfaces with `get_by` — same
- domain-level interfaces (the issue's reproduction is a domain-level interface)
- string fields still cast and match (no false positives)
- error shape assertions: `InvalidArgument` with `field` / `value` / `message`, never `InvalidFilterValue`

I believe #1971 can be closed as already fixed on `main` once confirmed; these tests
document the expected behavior going forward.

## Verification

- `mix test test/code_interface_test.exs` — 78 tests, 0 failures (72 existing + 6 new)
- `mix test` (full suite) — 4184 tests, 0 failures
- `mix format --check-formatted` — clean
- `mix credo --strict` — no issues
- `mix dialyzer` — 0 errors

Draft while CI validates.


lukegalea:
@zachdaniel happy to have you take a look when you get a chance — the PR is tests-only (see description: main already implements the #1971 fix, this locks in the update/destroy/domain-level scenarios). Flagging since I can't request reviewers directly from a fork.

zachdaniel:

PR #2956, opened 2026-09-20
Add repeated policy authorization benchmark

Valian:
## Summary

- add a benchmark reproducing the repeated `Ash.can?/3` workload
- run 500 authorization checks across 100 records and five update actions
- compare repeated `Ash.can?/3` calls with `Ash.can_do_all/3`
- keep the workload in memory to isolate policy preparation and evaluation

## Results

Run on an Apple M1 Pro with Elixir 1.20.1 and Erlang/OTP 29.0.2. Each benchmark iteration performs the same 500 authorization checks.

| Invocation | Iterations/s | Average | Median | p99 | Memory |
| --- | ---: | ---: | ---: | ---: | ---: |
| Repeated `Ash.can?/3` | 5.06 | 197.65 ms | 196.83 ms | 208.25 ms | 222.91 MB |
| `Ash.can_do_all/3` | 5.11 | 195.68 ms | 194.48 ms | 209.42 ms | 224.88 MB |

A second run moved `can_do_all/3` from approximately 0.4% slower to approximately 1% faster than the equivalent repeated calls, so there is no measurable throughput improvement. It consistently allocates about 2 MB more for this workload.

This matches the current implementation: `can_do_all/3` prepares shared options once, but then reduces over the entries and calls the ordinary authorization path independently for each one. It does not batch or reuse policy preparation. Its result accumulation also uses `acc ++ [entry]`, which accounts for some additional list allocation.

## Profiling overview

A local `:tprof` run over the same workload produces this approximate hot stack:

```text
Ash.can?/3
└─ Ash.Policy.Policy.solve/1
   └─ build_requirements_expression/2
      └─ expression/2
         ├─ simplify_policy_expression/2
         │  └─ Crux.Expression.simplify/1
         │     └─ Crux.Expression.RewriteRule
         │        ├─ create_walker_function
         │        ├─ default_walk_stateful
         │        ├─ compose_walker
         │        └─ individual Boolean rewrite laws
         └─ expand_invariants/2
            ├─ compare check variables pairwise
            └─ feed the enlarged expression through simplification
```

Approximately **91–92% of traced function time** is spent in `Crux.Expression` and its rewrite rules. The four main walker dispatch/composition functions alone account for approximately **54–56%**. The remainder is spread across expression postwalks and Boolean rewrite laws such as idempotency, associativity, absorption, consensus, distributivity, identity, complement, tautology, and unit resolution.

For 500 authorizations, the profile records:

- about 2.56 million calls through each of the two hottest rewrite walkers
- 400,000 expression postwalk calls
- 4,500 calls to `Crux.Expression.simplify/1`, or about nine per authorization
- no SAT-solver calls

### Repeated work

The expensive preparation is almost entirely repeated work in this benchmark. On the current implementation, policy-expression preparation receives the resource but not the concrete action, actor, or record. The same resource policies therefore produce the same pre-constant expression for all 500 checks. The records differ only by an `id` that these policies do not reference, and all five actions share the same action type and policy branch.

Within `Crux.Expression.simplify/1`, the same fixed set of eleven rewrite laws is composed and walked over every node. The profile shows how often those walkers run, but not how often a law actually changes a node; many rule invocations are expected to return the input unchanged because each law only matches particular shapes. Rule properties, chunks, and walker closures are also rebuilt for every simplification call even though the rule list is fixed.

The likely source of the cost is the order of policy preparation:

1. Ash constructs an expression for the complete policy set.
2. Action and action-type conditions remain symbolic even though the concrete action is already known.
3. Invariants are generated pairwise between symbolic checks.
4. Crux repeatedly traverses and simplifies the resulting expression.
5. Concrete checks and bypass outcomes are resolved only afterward.

Consequently, bypassing final policy evaluation does not bypass the expensive preparation and rewriting of the policy expression, and `can_do_all/3` currently repeats that preparation for every entry. Specializing action-dependent checks before invariant expansion should remove unrelated policy branches before they reach the general-purpose rewrite pipeline. Longer-term reuse could also target a prepared expression keyed by resource/action or the fixed Crux rewrite pipeline itself.

`:tprof` instruments function calls, so its wall time is substantially higher than normal execution and favors high-call-count functions. The percentages are therefore directional rather than an uninstrumented sampling flame graph, but the call counts and benchmark behavior point to the same hot path.

## Validation

- `mix format benchmarks/policy_can.exs --check-formatted`
- `mix run benchmarks/policy_can.exs` (twice)


zachdaniel:

zachdaniel:
```
Before (on my machine):

  ┌──────────────┬──────┬─────────┬────────┬────────┬────────┐
  │     Job      │ ips  │ average │ median │ 99th % │ memory │
  ├──────────────┼──────┼─────────┼────────┼────────┼────────┤
  │ manager      │ 4.83 │ 207.07  │ 206.67 │ 216.19 │ 209.54 │
  │ bypass       │      │      ms │     ms │     ms │     MB │
  ├──────────────┼──────┼─────────┼────────┼────────┼────────┤
  │ self-service │ 4.62 │ 216.35  │ 216.08 │ 222.28 │ 222.89 │
  │  policy      │      │      ms │     ms │     ms │     MB │
  ├──────────────┼──────┼─────────┼────────┼────────┼────────┤
  │ can_do_all   │ 4.61 │ 217.11  │ 216.57 │ 223.40 │ 224.84 │
  │              │      │      ms │     ms │     ms │     MB │
  └──────────────┴──────┴─────────┴────────┴────────┴────────┘

After:

  ┌──────────────┬───────┬─────────┬────────┬───────┬────────┐
  │     Job      │  ips  │ average │ median │ 99th  │ memory │
  │              │       │         │        │   %   │        │
  ├──────────────┼───────┼─────────┼────────┼───────┼────────┤
  │ manager      │ 69.49 │  14.39  │ 14.13  │ 16.42 │ 14.99  │
  │ bypass       │       │      ms │     ms │    ms │     MB │
  ├──────────────┼───────┼─────────┼────────┼───────┼────────┤
  │ self-service │ 57.03 │  17.53  │ 17.50  │ 19.28 │ 18.73  │
  │  policy      │       │      ms │     ms │    ms │     MB │
  ├──────────────┼───────┼─────────┼────────┼───────┼────────┤
  │ can_do_all   │ 55.82 │  17.91  │ 17.88  │ 19.80 │ 20.73  │
  │              │       │      ms │     ms │    ms │     MB │
  └──────────────┴───────┴─────────┴────────┴───────┴────────┘
```

